PR #6, opened 2025-01-04
[UniPoly] AddCommGroup by trimming after additions

mitschabaude:
This provides a sorry-free proof of `AddCommGroup` for `UniPolyC`, the canonical subtype of `UniPoly` which I defined as
```lean
{ p : UniPoly R // p.trim = p }
```

Addition is defined by first adding element-wise, and then trimming the array.

For trimming, I added a new custom primitive `Array.findIdxRev?` that traverses the array backwards, to search for the last non-zero element. Previously, it was defined using well-founded recursion which supposedly is slow at runtime.

Many new theorems are added that characterize the behavior of `Array.findIdxRev?`, `UniPoly.last_nonzero` and `UniPoly.trim`.

Note, this might not be the final way we prefer to represent polynomials in, either for computation or for proofs, but the added theorem infrastructure is likely reusable

### DIFF
--- a/ZKLib/Data/Math/Operations.lean
+++ b/ZKLib/Data/Math/Operations.lean
@@ -166,19 +166,32 @@ theorem rightpad_eq_if_rightpad_eq_of_ge (l l' : List α) (m n n' : Nat) (h : n 
     simp at h
     by_cases h' : m ≤ l.length <;> omega
 
--- TODO: finish this lemma, may need many cases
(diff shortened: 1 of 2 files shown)


quangvdao:
Thanks! Currently reviewing this, I think there is no need for a level of indirection with the `structure`. Instead it seems better to simply have:
```
@[reducible]
def UniPoly (R : Type*) := Array R
```
This way we can simplify many notations. I'm going over the lemmas and fixing them according to this definition.

quangvdao:
I can merge this now if there's no other comment.

PR #8, opened 2025-01-16
chore: fixing things that broke in recent VCV update

dtumad:
There's still a few things missing on the composition section because it seems like mathlib may have also broke some of the `ZKLib.Data.Math.Fin` stuff. I'd like to make some other larger changes based on the vcv changes, but this hopefully is close to fixing the build for now at least.


quangvdao:
I added back (most) of the extra materials from mathlib's zklib branch. Could you fix the rest of the broken VCVio stuff?

dtumad:
@quangvdao is there anything left to fix on this for now? Seems to build fine now at least.

PR #10, opened 2025-02-28
Progress on Proof Simplification

dtumad:
A couple remaining issues with `simp` pathing that force you to call `simp` multiple times before/after unfolding some definitions. Seems largely to be an issue with priorities on the lemmas.


ZKLib/ProofSystem/Sumcheck/Basic.lean
```diff
@@ -203,36 +203,63 @@ instance : Nonempty ((pSpec R d).FullTranscript) := by
...
+  -- Need `convert` because of some duplicate instances, should eventually track those down
+  convert (probEvent_eq_one_iff _ _).2 ⟨?_, ?_⟩
+  · simp only [Reduction.run, probFailure_bind_eq_zero_iff]
+    refine ⟨?_, ?_⟩
+    · simp -- There's still some pathing issue here w/ simp, need to simp in steps which is sub-par
+      unfold Prover.runAux
+      simp [Fin.induction, Fin.induction.go, reduction, prover]
+    · intro ⟨⟨r, ⟨(), st⟩⟩, log⟩ hx
+      simp -- Also some pathing issues, need to simp once before reducing `reduction`
+      simp [reduction, verifier, Verifier.run]
```
dtumad, line 264: Things like this would also be helped by having `simp` lemmas for each individual operation in the protocol, so you don't have to manually unfold e.g. `reduction` and `verifier`

PR #11, opened 2025-03-03
Unipoly conversion

mitschabaude:
this PR shows that conversion between `UniPoly` and mathlib's `Polynomial` satisfies many properties you'd expect:
* converting a `UniPoly` to mathlib and back yields the trimmed representative
* converting a mathlib poly to a `UniPoly` and back is the identity
* in particular, the conversion is a bijection between `UniPolyC` and `Polynomial`
* the conversion preserves evaluation at a point
* the conversion preserves addition

following mathlib, we define the coefficient at any Nat index (i.e. of any monomial) for UniPoly:
```lean
coeff : UniPoly R → ℕ → R
```
and we show that it coincides with mathlib's definition of `coeff`.

**Left for future PRs:**
* conversion preserves other poly operations like (scalar) multiplication
* I noticed the current definition of `eval` is suboptimal: $O(n^2)$ instead of $O(n)$, because the computation of powers $x ^ i$ is not cached. I still went with the current definition for now and will address this in a follow-up PR
* I think more effort should be spent to align definitions with mathlib. for example, `UniPoly.degree` behaves slightly different than `Polynomial.degree` (maps to `ℕ` instead of `ℕ | ⊥`)

### DIFF
--- a/ZKLib/Data/UniPoly/Basic.lean
+++ b/ZKLib/Data/UniPoly/Basic.lean
@@ -763,7 +763,7 @@ lemma toImpl_elim (p : Q[X]) :
   have hnat : p.degree = p.natDegree := degree_eq_natDegree (degree_ne_bot.mp hbot)
   simp [hnat]
 
-theorem ofPoly_toPoly {p : Q[X]} : p.toImpl.toPoly = p := by
+theorem toPoly_toImpl {p : Q[X]} : p.toImpl.toPoly = p := by
   ext n
   rw [coeff_toPoly]
   rcases toImpl_elim p with ⟨rfl, h⟩ | ⟨_, h⟩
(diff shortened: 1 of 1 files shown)


ZKLib/Data/UniPoly/Basic.lean
```diff
@@ -27,8 +27,10 @@ the same polynomial via zero-padding, for example `#[1,2,3] = #[1,2,3,0,0,0,...]
 def UniPoly (R : Type*) := Array R
 
 /-- Convert a `Polynomial` to a `UniPoly`. -/
-def Polynomial.toImpl {R : Type*} [Semiring R] (p : Polynomial R) : UniPoly R :=
-  .ofFn (fun i : Fin p.natDegree => p.coeff i)
+def Polynomial.toImpl {R : Type*} [Semiring R] (p : R[X]) : UniPoly R :=
+  match p.degree with
+  | ⊥ => #[]
+  | some d  => .ofFn (fun i : Fin (d + 1) => p.coeff i)
```
mitschabaude, line 33: the old definition was off by one, array length has to be `degree + 1` not `degree`

ZKLib/Data/UniPoly/Basic.lean
```diff
@@ -41,6 +43,9 @@ def coeffs {R : Type*} (p : UniPoly R) : Array R := p
 variable {R : Type*} [Ring R] [BEq R]
 variable {Q : Type*} [Ring Q]
 
+@[reducible]
+def coeff (p : UniPoly Q) (i : ℕ) : Q := p.getD i 0
```
mitschabaude, line 47: added this definition and used it in existing theorems where possible.

note that mathlib basically _defines_ polynomials as their `coeff` functions, so having the same notion here is useful for conversion

quangvdao:
Awesome, thanks & merged.

PR #12, opened 2025-03-11
fix `UniPoly` Tropical Equivalence

BoltonBailey:
Hey, I noticed the part about the equivalence with tropically bound polynomials is not right as stated (this is probably my fault as I think this part maybe originates with something I wrote).

The `UniPoly` type should be equivalent with the subsemiring where the polynomials respect the bound. I changed a few types around to make this happen, (and filled in one sorry as a bonus)

### DIFF
--- a/ZKLib/Data/UniPoly/Basic.lean
+++ b/ZKLib/Data/UniPoly/Basic.lean
@@ -1002,8 +1002,7 @@ noncomputable def Equiv.UniPoly.TropicallyBoundPolynomial {R : Type} [BEq R] [Ri
             simp
             exact rfl
         · simp [hi2]
-      right_inv := by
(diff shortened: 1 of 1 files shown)


quangvdao:
Nice, thanks! I should mention that we plan to prove an isomorphism between `Polynomial` and `UniPolyC`, the quotienting of `UniPoly` by trimming trailing zeros. Thus, going through `TropicallyBoundPolynomial` is not a priority for now.

In fact, we don't even need an isomorphism when using `UniPoly` in SNARK constructions. We only care that `Polynomial` has a split injection to `UniPoly`, and that `UniPoly.eval` is the same as `Polynomial.eval`, up to maps in both directions.

Issue #4, opened 2024-12-18
Definition & Security Proofs for Merkle Trees


quangvdao:
This issue tracks all development of Merkle trees. This is the basic vector commitment scheme that is required for all proof systems we care about (later development may bring in other schemes such as KZG, Bulletproofs, etc.)

We would like to prove completeness, extractability, and hiding for Merkle trees. These proofs (extractability and hiding) are non-trivial, and will require serious effort (from the core team or a serious outside contributor). The reference for this is in Chiesa & Yogev's recent zkSNARK textbook.

Relevant files in ZKLib:
- [Merkle Trees](https://github.com/quangvdao/ZKLib/blob/main/ZKLib/CommitmentScheme/MerkleTree.lean)
- [Basic Definitions of Commitment Schemes](https://github.com/quangvdao/ZKLib/blob/main/ZKLib/CommitmentScheme/Basic.lean)

Depends on progress in #4 to be able to state security properties for Merkle commitments.

Tasks:

- [ ] Flesh out the definitions of Merkle trees, especially the opening phase.
- [ ] Prove completeness of Merkle commitments.
- [ ] Initial development of the extractability proof.
- [ ] Initial development of the hiding proof.

References:
- Chapter 18 (Merkle commitment scheme) of [Building Cryptographic Proofs from Hash Functions](https://snargsbook.org/).
- Section 3 of the [original IOP paper](https://eprint.iacr.org/2016/116.pdf).

The textbook has more details but the paper might be an easier first read.

quangvdao:
@BoltonBailey just assigning you for tracking reason. Feel free to add updated TODOs here especially related to the inductively defined definition.

BoltonBailey:
For an update, this code now lives in VCVio, but we now have proof for single-index completeness and single-instance extractability. I have updated the above checklist to flesh out more points.

Abraxas1010:
I've opened Verified-zkEVM/VCV-io#473 addressing the **batch-index completeness** checkbox above (plus a batch analogue of the opening-uniqueness result).

[...]

PR #13, opened 2025-03-12
feat: desorry `buildMerkleTree`

BoltonBailey:
I was playing around with the library today, and saw there were some sorries in this file, so I changed some things around to get an unsorried (version of) the `buildMerkleTree` def. 

I also renamed a variable to make things a bit more readable to me, if that's not helpful feel free to tell me to revert that or ignore this PR entirely.

### DIFF
--- a/ZKLib/CommitmentScheme/MerkleTree.lean
+++ b/ZKLib/CommitmentScheme/MerkleTree.lean
@@ -46,6 +46,8 @@ lemma domain_def : (oracleSpec α).domain () = (α × α) := rfl
 @[simp]
 lemma range_def : (oracleSpec α).range () = α := rfl
 
+section
+
 variable [DecidableEq α] [Inhabited α] [Fintype α]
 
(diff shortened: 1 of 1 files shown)


BoltonBailey:
Partially addresses (but does not close) #4

quangvdao:
Thanks!

(copying from my message on Zulip)

An alternative implementation of Merkle tree that I think would be interesting to consider, is with respect to an inductive datatype (for example see this Rocq formalization https://github.com/logsem/clutch/blob/main/theories/eris/examples/merkle/merkle_tree.v). that might make it simpler to consider non-full binary trees, but require more reasoning about its size.

Another point is that, the `verifyMerkleTree` function was written when `OracleComp` did not have a failure option. Now that it does, one could utilize `failure` and let the result be `OracleComp spec Unit` instead. This has the advantage of `failure` propagating through many verification steps (though may come with the disadvantage of appearing more opaque).

PR #17, opened 2025-04-13
chore: Add API and changes to proof in MerkleTree

BoltonBailey:
This PR:

* Redefines `Cache.cons` in terms of the simpler existing Mathlib def `Fin.snoc`
* Defines `Cache.upper` and `Cache.leaves`, analogous to `head` and `tail` and proves some simple API about them
* Refactors `verifyProof` into the component that computes the putative root, and the component that compares this to the provided root`
* Refactors `generateProof` to be defined inductively, for more synergy with the rest of the proof style.
* Tries to make progress on the completeness theorem


BoltonBailey:
Is this change to the completeness theorem statement correct? I feel like `verifyProof α i leaf root proof = pure ()` should not be true in general, because if the Merkle tree is more than one layer, there will be hashes involved in the proof, and so the computation will not be pure.

For example, when I try to prove the n=1 case:

```
theorem completeness (leaves : List.Vector α (2 ^ 1)) (i : Fin (2 ^ 1)) :
    [ fun ⟨leaf, root, proof⟩ => verifyProof α i leaf root proof = pure () | do
      let cache ← buildMerkleTree α 1 leaves
      let proof := generateProof α i cache
      return (leaves[i], getRoot α cache, proof) ] = 1 := by
  simp [buildMerkleTree, buildLayer, generateProof, verifyProof, getPutativeRoot]
  intro x hx
  rw [ite_eq_iff]
  simp
  -- goal is now `False`
```

BoltonBailey:
As always feel free to change/revert parts.

PR #25, opened 2025-04-24
[DRAFT] Updating documentation

alexanderlhicks:
This is a draft PR aimed at updating documentation, more specifically:

- Addition of a CONTRIBUTING.md file [in progress]
- A pass over the ROADMAP.md file to identify identify things that are now being worked on, missing desirables, ... [todo]
- A pass of the README.md file to better indicate currently active lines of work as well as adding a reference to VCVio as it is an important dependency that is undergoing parallel development that can affect this library [in progress]

Comments and suggestions are welcome.


alexanderlhicks:
A few too many meetings the last few days but I should be able to finish it up tonight.
A helpful addition would be a basic overview/mini tutorial of how to represent arklib/vcv specific objects that wouldn't be covered by the mathlib style guide or map 1-1 from the pen & paper literature.
(See CONTRIBUTING.md but depending on the scope it could be a separate file.)

alexanderlhicks:
Ok, I did just a quick pass:
- Simplified the README so that it lists basic information, active formalizations and refers to CONTRIBUTING/ROADMAP/BACKGROUND for the rest. I've only listed sumcheck, FRI and STIR/WHIR as active right now, and we can add anything that is being worked on in the sense of "someone has committed to getting it done soon-ish". I haven't listed specific work on the underlying theory but this could also be added since it isn't documented elsewhere right now (other than the new-interaction-types branch). Likewise for the polynomial type work. Once some work is finished we should also create a list of finished work to distinguish from the other unfinished/very early work on many components of the library.
- CONTRIBUTING can be expanded later on with a more specific style guide, but I figured this would also depend on more general documentation so we can leave it as is for now -- I don't think we do anything that is not compatible with the mathlib style guide.
- Small changes to ROADMAP, primarily removing references to specific schemes that we don't currently have plans for. This is mostly to avoid having a roadmap document that overtime accumulates to work that may be less desired as time passes. We could add more specific references to the interactions types, polynomial type, and links to branches/folders with ongoing work for the active items of the roadmap. As the library evolves, we might opt to remove the reference to the old version of the roadmap as these will probably cease to make much sense.
- Renamed References to BACKGROUND, since there are references in the different .lean files as is.

I don't particularly strong opinions about these changes, so feel free to tweak anything you want, but I think this makes things a bit clearer for someone looking at the library for the first time.

quangvdao:
@lalalune could you stop posting AI slop on our PRs?

PR #31, opened 2025-05-06
Coding theory definitions and proofs

katyhr:
`DivergenceOfSets.lean`: Proved definition for divergence of an arbitrary set `U` from an arbitrary set `V`.

`InterleavedCodes.lean`:- Definition of the `κ`-interleaved code `IC` of a linear code `LC` over a semiring `F`. Definition of distances for interleaved codes and statement for the relation between the minimal distance of an interleaved code and its underlying linear code. Statements of proximity results for Reed-Solomon codes  (`Lemma 4.3`, `Lemma 4.4` and `Lemma 4.5` from Ligero) with proximity parameter less than the minimal code distance divided by `3`. 

`LinearCodes.lean`: Definition of a linear code, minimal distance of a linear code, length, dimension and rate. Linear codes defined by a generator matrix and a proof that the dimension of a linear code equals the rank of its generator matrix. Definition of the weight of a vector, statement and proof of that fact that the minimal distance of a linear code is equal to the minimal weight taken over the set of codewords.

`Prelims.lean `: Preliminary notions needed for the definitions and proofs in the FRI blueprint. Currently, the file only contains a couple of definitions (affine line, row span, set of non-equal columns). This will be expanded as the project progresses.

`ReedSolomon.lean `: Proposed generalisation of the existing ArkLib definition of Reed-Solomon code: we propose to change the field F to a semiring F. This matches the generality of the definitions and statements in the rest of the library as well as of mathlib's polynomial library. 

`RelativeHammingDistance.lean`: Proved definitions of relative Hamming distance and related notions for a linear code `LC`. This covers all corner cases, too.


ArkLib/Data/CodingTheory/DivergenceOfSets.lean
```diff
@@ -0,0 +1,49 @@
...
+open Classical
+
+variable {ι : Type*} [Fintype ι]
+         {F : Type*}
+         {U V C : Set (ι → F)}
+
+/--
+`d_U` is the set of achievable relative Hamming distances between elements `u ∈ U` and `V`.
+-/
+def d_U [Nonempty ι] (U V : Set (ι → F)) : Set ℚ :=
```
quangvdao: Could you make this more descriptive? and this seems generalizable as well. For instance:
```
def Set.possibleDistances {α β γ : Type*} (U : Set α) (V : Set β) (metric : α → Set β → γ) : Set γ :=
  { d : γ | ∃ u ∈ U, metric u V = d }

def d_U' [Nonempty ι] (U V : Set (ι → F)) : Set ℚ :=
  Set.possibleDistances U V RelativeHamming.distToCode
  ```

File: ArkLib/Data/CodingTheory/DivergenceOfSets.lean [hunk 2]
```diff
@@ -0,0 +1,49 @@
...
+  λ _ _ ↦ by aesop (add simp d_U)
+
+@[simp]
+lemma finite_d_U [Nonempty ι] : (d_U U V).Finite :=
+  Set.Finite.subset RelativeHamming.finite_relHammingDistRange d_U_subset_relHammingDistRange
+
+/--
+divergence of an arbitrary set `U` from an arbitrary set `V`.
+-/
+def div_U_from_V [Nonempty ι] (U : Set (ι → F)) (V : Set (ι → F)) : ℚ :=
```
quangvdao: Could you instead of hardcoding variable names U and V in the definition, name it by describing what it is doing? This applies to other definitions as well.

ArkLib/Data/CodingTheory/Prelims.lean
```diff
@@ -0,0 +1,44 @@
...
+
+/--
+affine line between vectors `u` and `v`.
+-/
+def line [Ring F] (u v : Fin ι → F) : Set (Fin ι → F) :=
+  {x | ∃ γ : F, x = γ • u + (1 - γ) • v}
+
+end Affine
+
+namespace Matrices
```
quangvdao: Put Matrix not Matrices

ArkLib/Data/CodingTheory/Prelims.lean
```diff
@@ -0,0 +1,44 @@
...
+open Classical
+
+noncomputable section
+
+variable {F : Type*} [Semiring F]
+variable {κ ι : ℕ}
+
+namespace Affine
+
+/--
```
quangvdao, line 20: Mathlib already has the definition:

https://leanprover-community.github.io/mathlib4_docs/Mathlib/LinearAlgebra/AffineSpace/AffineSubspace/Defs.html#spanPoints

quangvdao: You can prove that this mathlib definition is equivalent to your definition, but in general I prefer using what Mathlib already has.

ArkLib/Data/CodingTheory/RelativeHammingDistance.lean
```diff
@@ -0,0 +1,160 @@
...
+
+variable {ι : Type*} [Fintype ι]
+         {F : Type*}
+         {u v w c : ι → F}
+         {C : Set (ι → F)}
+
+@[simp]
+lemma zero_lt_card [Nonempty ι] : 0 < Fintype.card ι := by have := Fintype.card_ne_zero (α := ι); omega
+
+abbrev LinearCode.{u, v} (ι : Type u) (F : Type v) [Semiring F] : Type (max u v)
```
quangvdao: Could you put this in the `Basic.lean` file?

ArkLib/Data/CodingTheory/RelativeHammingDistance.lean
```diff
@@ -0,0 +1,160 @@
...
+-/
+open Classical
+
+variable {ι : Type*} [Fintype ι]
+         {F : Type*}
+         {u v w c : ι → F}
+         {C : Set (ι → F)}
+
+@[simp]
+lemma zero_lt_card [Nonempty ι] : 0 < Fintype.card ι := by have := Fintype.card_ne_zero (α := ι); omega
```
quangvdao: Rename to `Fintype.zero_lt_card`

ArkLib/Data/CodingTheory/RelativeHammingDistance.lean
```diff
@@ -0,0 +1,160 @@
...
+
+abbrev LinearCode.{u, v} (ι : Type u) (F : Type v) [Semiring F] : Type (max u v)
+  := Submodule F (ι → F)
+
+noncomputable section
+
+namespace RelativeHamming
```
quangvdao: Namespaces make invoking these definitions awkward. You should put this as `section`, and rename `dist` below to `relHammingDist`.

ArkLib/Data/CodingTheory/RelativeHammingDistance.lean
```diff
@@ -0,0 +1,160 @@
...
+
+/--
+`δᵣ(u,v)` denotes the relative Hamming distance between vectors `u` and `v`.
+-/
+notation "δᵣ(" u ", " v ")" => dist u v
+
+/--
+the range of the relative Hamming distance function.
+-/
+def distRange (ι : Type*) [Fintype ι] : Set ℚ :=
```
quangvdao: same with this, `relHammingDistRange`

ArkLib/Data/CodingTheory/RelativeHammingDistance.lean
```diff
@@ -0,0 +1,160 @@
...
+        ⟩⟩
+      ⟩
+
+@[simp]
+lemma finite_offDiag [Finite F] : C.offDiag.Finite := Set.Finite.offDiag (Set.toFinite _)
+
+/--
+The set of possible distances between distinct codewords of `C`.
+-/
+def d_C (C : Set (ι → F)) : Set ℚ :=
```
quangvdao:
```
def Set.possibleDistances {α γ : Type*} {s : Set α} {f : α → α → γ} : Set γ :=
  { d : γ | ∃ p ∈ s.offDiag, f p.1 p.2 = d }
```
This means we'll need to rename my above proposal for the divergence of set definitions. This feels like the "right" definition for within a set. The other one could be `possibleDistances₂`?

Also might be better if you put this definition in the `Basic.lean` file (and in the general formulation above). It is more general than just relative hamming distance.

ArkLib/Data/CodingTheory/RelativeHammingDistance.lean
```diff
@@ -0,0 +1,160 @@
...
+lemma finite_d_C [Nonempty ι] : (d_C C).Finite :=
+  Set.Finite.subset finite_relHammingDistRange d_C_subset_relHammingDistRange
+
+end
+
+/--
+relative Hamming Distance of a code `C`.
+-/
+def minDistCode [Nonempty ι] (C : Set (ι → F)) : ℚ :=
+  have : Fintype (d_C C) := @Fintype.ofFinite _ finite_d_C
```
quangvdao: You should use `haveI` for these settings, which would inline the `have` definition.

ArkLib/Data/CodingTheory/RelativeHammingDistance.lean
```diff
@@ -0,0 +1,160 @@
...
+
+/--
+`δᵣ C` denotes the minimum relative Hamming distance of a code `C`
+-/
+notation "δᵣ" C => minDistCode C
+
+/--
+`d_w` is the set of values of relHamming distance between a vector `w` and codewords `c ∈ C`.
+-/
+def d_w (w : ι → F) (C : Set (ι → F)) : Set ℚ :=
```
quangvdao: Again, this should go in `Basic.lean`, and be more descriptive. This is essentially a precursor to `distFromCode`, except that this is Rat-valued not (E)Nat-valued. The "right" way seems to be defining these notions with respect to an arbitrary `(E)Dist` on the function type `ι → F`. And relative Hamming distance is a special case.

This way the theorems you state below can be true automatically for other situations (regular Hamming distance, "weighted" Hamming distance for other weights, etc.)

ArkLib/Data/CodingTheory/RelativeHammingDistance.lean
```diff
@@ -0,0 +1,160 @@
...
+@[simp]
+lemma finite_d_w [Nonempty ι] : (d_w w C).Finite :=
+  Set.Finite.subset finite_relHammingDistRange d_w_subset_relHammingDistRange
+
+instance [Nonempty ι] : Fintype (d_w w C) := @Fintype.ofFinite _ finite_d_w
+
+/--
+relative Hamming distance between a vector `w` and a code `C`.
+-/
+def distToCode [Nonempty ι] (w : ι → F) (C : Set (ι → F)) : ℚ :=
```
quangvdao: Yeah, we definitely need to parametrize over an abstract distance like above, since this is essentially the same as `distFromCode` in `Basic.lean`, but with relative Hamming distance instead of (regular) Hamming distance.

ArkLib/Data/CodingTheory/LinearCodes.lean
```diff
@@ -0,0 +1,99 @@
...
+
+noncomputable section
+
+variable {ι κ : ℕ}
+         {F : Type*} [CommRing F]
+
+/--
+A linear code of length `ι` defined by right multiplication by a `κ x ι` generator matrix `G`.
+-/
+def mul_GenMat (G : Matrix (Fin κ) (Fin ι) F) : LinearCode ι F :=
```
quangvdao: Consider mathlib convention for naming, i.e. `mulGenMat`. Also, this seems to be better as `fromRowGenMat`? And the below `genMat_mul` can be `fromColGenMat`? Maybe you should double-check what should be row and what should be column, and clearly says this in the documentation.

ArkLib/Data/CodingTheory/LinearCodes.lean
```diff
@@ -0,0 +1,99 @@
...
+  distance of a linear code equals the minimal weight taken over the set of codewords.
+-/
+
+open Classical
+open Finset
+
+noncomputable def wt {F : Type*} [Semiring F] [Zero F] {ι : ℕ}
+  (v : Fin ι → F) : ℕ := #{i | v i ≠ 0}
+
+namespace LinearCodes
```
quangvdao: Don't put plurals in the namespace.

ArkLib/Data/CodingTheory/LinearCodes.lean
```diff
@@ -0,0 +1,99 @@
...
+Authors: Katerina Hristova, František Silváši, Julian Sutherland
+-/
+
+import Mathlib.Data.Matrix.Rank
+import Mathlib.InformationTheory.Hamming
+import Mathlib.LinearAlgebra.Matrix.ToLin
+import Mathlib.Algebra.Module.Submodule.Range
+import Mathlib.Algebra.Module.Submodule.Defs
+
+/-!
```
quangvdao, line 13: I'd like this to be fully merged into `Basic.lean`. That file should be quite large (1000+ lines) consisting of all the general coding theory definitions people might need. Lemmas can be in separate files. I can also rename that file to `Defs.lean` to emphasize that all definitions should go there.

quangvdao:
Left a bunch of comments. I'm a bit nitpicky about the right naming / organization for the definitions, so that things could be as general as possible, which would mean fewer theorems needed & people can use the definitions for more use cases.

katyhr:
> Left a bunch of comments. I'm a bit nitpicky about the right naming / organization for the definitions, so that things could be as general as possible, which would mean fewer theorems needed & people can use the definitions for more use cases.

Hey Quang, thanks! I'm sorry if this is a stupid question, but where exactly are the comments? I can't seem to find them. Or do you mean the comments you left in the telegram chat?

quangvdao:
If you look above or in the Files changed section, they should appear? I’m just commenting directly on your code

Julek:
@quangvdao I'm also not seeing any comments posted from you?

ArkLib/Data/CodingTheory/DivergenceOfSets.lean
quangvdao: @katyhr Can you see this?

ArkLib/Data/CodingTheory/RelativeHammingDistance.lean
quangvdao: @Julek How about this one?

quangvdao (changes requested):
Comments on the PR

ArkLib/Data/CodingTheory/DivergenceOfSets.lean
katyhr: Yes! Thanks :)

ArkLib/Data/CodingTheory/DivergenceOfSets.lean
katyhr: Sure, will implement across all files.

PR #32, opened 2025-05-06
Stir tools + main theorems v1

dpoulami:
contains 
* requisites from coding theory in Stir/ToCodingTheory
** ReedSolomon codes
** fractional Hamming distance
** Proximity bound parameters
** Johnson bound
* tools for STIR (section 4, ACFY25)
** folding univariate functions
** quotienting of polynomials and functions
** combining functions of varying degrees using degree correction
** out of domain sampling
* Stir main theorem (section 5, ACFY25)
* Stir round-by-round soundness theorem

This is still a work in progress.


ArkLib/ProofSystem/Stir/Combine.lean
```diff
@@ -0,0 +1,140 @@
...
+lemma geometric_sum_units {F : Type*} [Field F] [DecidableEq F] (r :  Units F) (a : ℕ) :
+  ∑ j : Fin (a + 1), (r ^ (j:ℕ) : F) =
+    if r = 1 then (a + 1 : F)
+    else (1 - r ^ (a + 1)) / (1 - r) := by sorry
+
+/-- Coefficients r_i as used in the definition of Combine(),
+r_0 := 1, r_i := r^{i + sum_{j < i}(d* - d_j)}
+for i > 0  (We range 0...m-1, not 1...m as in STIR)-/
+def ri {F : Type*} [Field F] {m : ℕ} (dstar : ℕ) (degs : Fin m → ℕ) (r : F) (i : Fin m) : F :=
+  if i.val = 0 then (1:F)
```
quangvdao: Pretty sure you don't need the special case, the general definition specializes to r^0 = 1.

ArkLib/ProofSystem/Stir/Combine.lean
```diff
@@ -0,0 +1,140 @@
...
+  {F : Type*} [Field F]
+  (L : Finset F)
+  (m : ℕ)
+  (dstar : ℕ)
+  (r : F)
+  (fs : Fin m → L → F)
+  (degs : Fin m → ℕ)
+  (hdegs : ∀ i, degs i ≤ dstar) : L → F :=
+    fun x =>
+      (Finset.univ : Finset (Fin m)).sum
```
quangvdao: Please use the bigoperator notation throughout for better readability

ArkLib/ProofSystem/Stir/Combine.lean
quangvdao:
```
  r^(i.val + ∑ j < i, (dstar - degs j))
```
works perfectly well

ArkLib/ProofSystem/Stir/Combine.lean
quangvdao:
```
    ∑ i, (ri dstar degs r i) * fs i x * (∑ l < (dstar - degs i + 1), (r * x.val)^l)
```

quangvdao: Do the same for all the below functions with sums.

ArkLib/ProofSystem/Stir/Combine.lean
```diff
@@ -0,0 +1,140 @@
...
+  (r : F)
+  (fs : Fin m → L → F)
+  (degs : Fin m → ℕ)
+  (hdegs : ∀ i, degs i ≤ dstar) : L → F :=
+    fun x =>
+      let q := r * x.val
+      (Finset.univ : Finset (Fin m)).sum
+      fun i =>
+        let di := degs i
+        let geom := if q = 1 then (dstar - di + 1 : F)
```
quangvdao: This term (1 - q^n)/(1-q) (and n if q = 1) should be made into its own definition with a good naming.

And come to think of it, this should just be `∑ j < n, q^j`, where `n = dstar - di + 1`. This is better since there is no more casing; you can then prove that if q is not 1 then this is equal to the division. Maybe no separate definition is needed, though you should prove lemmas about this.

ArkLib/ProofSystem/Stir/Combine.lean
```diff
@@ -0,0 +1,140 @@
...
+lemma combine
+  {F : Type*} [Field F] [Fintype F] [DecidableEq F]
+  {L : Finset F}
+  {dstar : ℕ}
+  {m : ℕ}
+  (Cstar : ReedSolomonCode F L dstar)
+  (fs   : Fin m → L → F)
+  (degs : Fin m → ℕ)
+  (hdegs : ∀ i, degs i ≤ dstar)
+  (δ    : {δ // 0 < δ ∧ δ < combineRange Cstar })
```
quangvdao: I think it's better to introduce `δ : Real` followed by two hypotheses about `δ`, namely `hδPos` and `hδLt`.

ArkLib/ProofSystem/Stir/Folding.lean
```diff
@@ -0,0 +1,181 @@
...
+   This is MonomialOrder.div from Mathlib.RingTheory.MvPolynomial.Groebner
+
+   Using the usual lexicographic order x₀ > x₁ is equal to proposition 6.3 in
+   https://people.csail.mit.edu/madhu/papers/2005/rspcpp-full.pdf under the
+   substitution z = x₀ and y = x₁, hence the following definition constructs
+   Q ∈ 𝔽[Z,Y] with P(z,y) = Q'(z,y) * R(z,y) + Q(z,y)
+-/
+/-- Given `P, P' ∈ 𝔽[Z,Y]`, `P' ≠ 0`, computes `Q ∈ 𝔽[Z,Y]`,
+with `P(z,y) = Q'(z,y) * P'(z,y) + Q(z,y)` for some `Q' ∈ 𝔽[Z,Y]` -/
+noncomputable def modBivar
```
quangvdao: Consider using `Polynomial (Polynomial F))` instead of `MvPolynomial (Fin 2) F`. https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Polynomial/Bivariate.html#Polynomial.Bivariate.termY

This makes expressing the division property a lot easier. And in general there's not much connection between `MvPolynomial` and `Polynomial`, which someone should pick up on. Even something as simple as `MvPolynomial (Fin 1) R` is isomorphic to `Polynomial R` is not in mathlib, as far as I can tell. I think I have defined this in ArkLib though, check the `MvPolynomial` folder.

So avoiding MvPolynomial if you can (esp. in this situation) is best.

ArkLib/ProofSystem/Stir/Folding.lean
```diff
@@ -0,0 +1,181 @@
...
+  -- Pbi(z,y):= P(z)
+  let Pbi : MvPolynomial (Fin 2) F := uni2bi P
+  -- P'(z,y) := (y - q(z))
+  let P' : MvPolynomial (Fin 2) F := (MvPolynomial.X 1) - uni2bi q
+  -- proof that leading coefficient f q is not zero
+  have h_unit : IsUnit ((MonomialOrder.lex).leadingCoeff P') := sorry
+  modBivar Pbi P' h_unit
+
+/-- Helper For Readability: Evaluate a bivariate polynomial Q at (a, b) ∈ F×F -/
+def evalBivar
```
quangvdao: If using `Polynomial (Polynomial F)`, which can be abbreviated as `F[X][Y]` once you open then `Bivariate` scope, then this has been defined in mathlib as `evalEval`.

ArkLib/ProofSystem/Stir/Folding.lean
```diff
@@ -0,0 +1,181 @@
...
+  (C1 : ReedSolomonCode F L d)
+  (C2 : ReedSolomonCode F (powDom L k) (d/k))
+  (δ : {x : ℝ // 0 < x ∧ x < foldingRange C1 f}) :
+    (PMF.uniformOfFintype F).toOuterMeasure { r |
+            fractionalHammingDistSet
+              (fold f k r)
+              (C2.code)
+              (C2.nonempty)
+            ≤ δ.val } > err' F (d/k) C1.rate δ k
+            -- Double check if this really is C1.rate not C2.rate
```
quangvdao: This looks painful & very opaque... you should look around mathlib for probability notation (or define one yourself) so that people can immediately understand what this is saying (I cannot).

ArkLib/ProofSystem/Stir/Folding.lean
```diff
@@ -0,0 +1,181 @@
...
+  (f : L → F)
+  (k : ℕ)
+  (x : powDom L k) : Polynomial F :=
+    let S := powFiber L k x
+    Lagrange.interpolate
+      S.attach
+      Subtype.val
+      fun i =>
+        let hL : i.1 ∈ L := (Finset.mem_filter.1 i.2).1
+        f ⟨i.1, hL⟩
```
quangvdao: Another pretty opaque definition. Why is it correct? Why do you need to put `attach`?

ArkLib/ProofSystem/Stir/Folding.lean
```diff
@@ -0,0 +1,181 @@
...
+  (r : F): Polynomial F :=
+    let q : Polynomial F := Polynomial.X ^ k
+    let hdeg_q_min : q.natDegree > 0 := sorry
+    let hdeg_q_max : q.natDegree < Fintype.card F := sorry
+  -- choose the unique bivariate lift Q
+    let Q : MvPolynomial (Fin 2) F := polyQ f q
+    MvPolynomial.eval₂Hom
+      (Polynomial.C : F →+* Polynomial F)
+      (fun i : Fin 2 => if i = 0 then Polynomial.X else Polynomial.C r)
+      Q
```
quangvdao: This will hopefully be a lot simpler with the `Polynomial Polynomial` change.

ArkLib/ProofSystem/Stir/Folding.lean
```diff
@@ -0,0 +1,181 @@
...
+/-- Fact 4.6.2 in STIR-/
+lemma degree_bound_bivariate
+  {F  : Type*} [Field F] [Fintype F]
+  (q : Polynomial F) (hdeg_q_min : q.natDegree > 0) (hdeg_q_max : q.natDegree < Fintype.card F)
+  {t : ℕ}
+  (Q : MvPolynomial (Fin 2) F)
+  (hdegX : MvPolynomial.degreeOf 0 Q < t)
+  (hdegY : MvPolynomial.degreeOf 1 Q < q.natDegree) :
+  (MvPolynomial.eval₂Hom (Polynomial.C : F →+* Polynomial F)
+      (fun i : Fin 2 => if i = 0 then q else Polynomial.X) Q).natDegree < t * q.natDegree :=
```
quangvdao: for Fin notation, you could use `![q, X]` for readability. But this is going away anyway with the change to `Polynomial Polynomial`

ArkLib/ProofSystem/Stir/IOPP.lean
```diff
@@ -0,0 +1,57 @@
...
+  /--**One-shot soundness**-/
+  ioppSoundness : soundness langIn langOut verifier ε_sound
+
+  /-- A *state function* used for round-by-round soundness. -/
+  stateFn :
+    StateFunction langIn langOut verifier
+
+  /--**round-by-round soundness** -/
+  ioppRBRSoundness : rbrSoundness langIn langOut verifier stateFn ε_rbr
+
```
quangvdao: See my branch for the updated definition: https://github.com/Verified-zkEVM/ArkLib/commit/e92dfde20d6a22ff68eeb3338d319d0d6f6b185a

ArkLib/ProofSystem/Stir/MainThm.lean
```diff
@@ -0,0 +1,158 @@
...
+
+/-- **Degree after `i` folds.**
+The starting degree is `d`; every fold divides it by `kⱼ (j<i)` to obtain `dᵢ`.
+We assume divisibility so the integer division is exact. -/
+def degreeᵢ {M : ℕ} (d : ℕ) (P : Params F M) (i : Fin (M + 1)) : ℕ :=
+  d / ∏ j : Fin i, P.k (Fin.castLT j (Nat.lt_trans j.isLt i.isLt))
+
+/-- **`Rateᵢ= degreeᵢ/|Lᵢ|`** of the Reed–Solomon code used in round `i`.
+A real number because most analytic bounds live in `ℝ`. -/
+noncomputable def rateᵢ {M : ℕ} (d : ℕ) (P : Params F M)
```
quangvdao: You can get that this is nonnegative for free though. This works:
```
noncomputable def rateᵢ {M : ℕ} (d : ℕ) (P : Params F M)
    (i : Fin (M + 1)) : ℝ≥0 :=
  (degreeᵢ d P i) / ((P.L i).card)
  ```
This saves you time later on when proving properties involving rate.

File: ArkLib/ProofSystem/Stir/ToCodingTheory/ReedSolomonCodes.lean [hunk 13]
```diff
@@ -0,0 +1,99 @@
...
+
+/-- The ReedSolomonCode structure, storing the codewords as `Finset (L → F)`. -/
+structure ReedSolomonCode
+  (F : Type*) [Field F] [Fintype F] [DecidableEq F]
+  (L : Finset F)
+  (d : ℕ) where
+    nonempty_L : L.Nonempty
+    L_ne_F     : L ≠ (Finset.univ : Finset F)
+    code       : Finset (L → F)
+    code_def   : code = (polynomialsUpTo F d).image (polynomialEvalOn F L)
```
quangvdao: Some problems with this definition. Why do you need L not equal to the whole field (and nonempty)? The fact that `L` is a Finset means it does not come with a canonical ordering, which would be bad when you want to reference that ordering (i.e. roots of unity)

I think it was agreed on at the meeting that you'll be switching to Katy's definition of Reed-Solomon code anyway, so the above are moot.

ArkLib/ProofSystem/Stir/ToCodingTheory/ProximityBound.lean
```diff
@@ -0,0 +1,24 @@
...
+ -- coerced to ENNReal, because it is exclusively used in probability inequalities and
+  we agreed to coerce ℝ → ENNReal not vice versa -/
+noncomputable def err' (F : Type*) [Fintype F] (d : ℕ) (ρ : ℝ) (δ : ℝ) (m: ℕ) : ENNReal :=
+  ENNReal.ofReal (
+  if δ ≤ (1 - ρ) / 2 then
+    ((m - 1) * d) / (ρ * (Fintype.card F))
+  else
+    let min_val := min (1 - (Real.sqrt ρ) - δ ) ((Real.sqrt ρ) / 20)
+    ((m - 1) * d^2) / ((Fintype.card F) * (2 * min_val)^7)
+)
```
quangvdao: Please use this definition, which gives you NNReal for free (and in particular proves that it is not infinity):
```
noncomputable def err' (F : Type*) [Fintype F] (d : ℕ) (ρ : ℝ≥0) (δ : ℝ≥0) (m: ℕ) : ℝ≥0 :=
  if δ ≤ (1 - ρ) / 2 then
    ((m - 1) * d) / (ρ * (Fintype.card F))
  else
    let min_val := min (1 - ρ.sqrt - δ) (ρ.sqrt / 20)
    ((m - 1) * d^2) / ((Fintype.card F) * (2 * min_val)^7)
 ```
 
 Some subtleties in the definition here. Subtraction over `NNReal` is truncating. This is not a problem in this definition since we only encounter subtraction in `m - 1`, which in practice is always nonnegative since `m>=1`, and in `min_val`, which is also not a problem since it's a min of two terms.
 
 Finally, if `min_val` is zero, then division by zero is defined in Lean to be zero (see `_root_.div_zero`).
 
 Please put all these caveats in the documentation.

File: ArkLib/ProofSystem/Stir/ToCodingTheory/FracHammingDist.lean [hunk 15]
```diff
@@ -0,0 +1,21 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Least Authority
+-/
+
+import Mathlib.Data.Real.Basic
+import Mathlib.InformationTheory.Hamming
+
+def fractionalHammingDist
+  {α β : Type*} [Fintype α] [DecidableEq β]
+  (f g : α → β) : ℚ := (hammingDist f g : ℚ) / Fintype.card α
```
quangvdao: Change to use Katy's definition once that PR is merged.

ArkLib/ProofSystem/Stir/ToCodingTheory/ErrCorrCodes.lean
```diff
@@ -0,0 +1,61 @@
...
+
+/-!  ###  Error–correcting codes over an alphabet  -/
+
+
+/-- An error correcting code of finite length `|ι|` over some finite alphabet `A`. -/
+structure ErrCorrCode
+  (A : Type*) [Fintype A] [DecidableEq A]
+  (ι : Type*) [Fintype ι] where
+    words : Finset (ι → A)
+
```
quangvdao: Change to follow ArkLib's definition.

ArkLib/ProofSystem/Stir/ToCodingTheory/ErrCorrCodes.lean
```diff
@@ -0,0 +1,61 @@
...
+namespace ErrCorrCode
+
+variable {A : Type*} [Fintype A] [DecidableEq A] {ι : Type*} [Fintype ι] [DecidableEq ι]
+
+
+/-- The set `Λ(C, f, δ)` of codewords within fractional Hamming distance `δ`
+of function `f: ι → A`. -/
+noncomputable def list (C : ErrCorrCode A ι) (f : ι → A) (δ : ℝ) : Finset (ι → A) :=
+  C.words.filter (fun c ↦ fractionalHammingDist f c ≤ δ)
+
```
quangvdao: I think it is fine to let this be a `Set` rather than a `Finset`. Then we can have hypotheses that these sets are finite in lemmas (or prove that they are finite).

ArkLib/ProofSystem/Stir/Quotienting.lean
```diff
@@ -0,0 +1,81 @@
...
+    let pV   := vanishingPoly F S
+    fun x =>
+      if hx : x.val ∈ S then
+        Fill ⟨x.val, hx⟩
+      else
+        (f x - pAns.eval x.val) / pV.eval x.val
+
+/-- We define the set T(f,L,S,Ans) as the set of all points x ∈ L that lie in S such that
+  the AnsPoly disagrees with f. Formally: T := { x ∈ L | x ∈ S ∧ AnsPoly x ≠ f x }. -/
+noncomputable def T
```
quangvdao: Please make this definition more descriptive. Something like `disagreementSet` is a lot better (and can be improved upon).

ArkLib/ProofSystem/Stir/Quotienting.lean
```diff
@@ -0,0 +1,81 @@
...
+  (Ans : S → F)
+  (Fill : S → F)
+  : L → F :=
+    let pAns := ansPoly F S Ans
+    let pV   := vanishingPoly F S
+    fun x =>
+      if hx : x.val ∈ S then
+        Fill ⟨x.val, hx⟩
+      else
+        (f x - pAns.eval x.val) / pV.eval x.val
```
quangvdao: Please add documentation. Why is there a Fill function here? And the definition should be more descriptive, i.e. `Polynomial.quotientByVanishingPoly`.

ArkLib/ProofSystem/Stir/Folding.lean
quangvdao: Okay, after looking I couldn't find any notation. I'll try to define something and refer you to it later.

ArkLib/ProofSystem/Stir/MainThm.lean
```diff
@@ -0,0 +1,158 @@
...
+/-- **Family of Reed–Solomon codes** actually expected by the verifier.
+The index `i` says *which* round we refer to; all codes are computed from
+the same source polynomial `f`. -/
+structure Code
+    {F : Type*} [Field F] [Fintype F] [DecidableEq F]
+    {d : ℕ} {M : ℕ} {P : Params F M} (i : Fin (M + 1)) where
+  codeᵢ : (j : Fin (M + 1)) → ReedSolomonCode F (P.L j) (degreeᵢ d P j)
+
+/-- Predecessor inside `Fin (n+1)` (requires `i ≠ 0`). -/
+def predFin {n : ℕ} (i : Fin (n + 1)) (h : i.val ≠ 0) : Fin (n + 1) :=
```
quangvdao: Please use `Fin.pred`

ArkLib/ProofSystem/Stir/MainThm.lean
```diff
@@ -0,0 +1,158 @@
...
+  ⟨Nat.pred i.val, by
+    have hpred : Nat.pred i.val < i.val := Nat.pred_lt h
+    exact Nat.lt_trans hpred i.isLt⟩
+
+/-- Max of `ε_fold`, `ε_fin`, all `ε_out` and `ε_shift`.  Uses `max'` so
+no `OrderBot ℝ≥0` instance is required. -/
+noncomputable def maxErrors {M : ℕ}
+    (ε_fold ε_fin : ℝ≥0) (ε_out ε_shift : Fin (M + 1) → ℝ≥0) : ℝ≥0 :=
+  (insert ε_fold
```
quangvdao: More compact notation:
```
  ({ε_fold} ∪ {ε_fin} ∪ univ.image ε_out ∪ univ.image ε_shift).max' (by simp)
```

ArkLib/ProofSystem/Stir/MainThm.lean
```diff
@@ -0,0 +1,158 @@
...
+  (pSpec : ProtocolSpec n) (oSpec : OracleSpec ι)
+  [∀ i, VCVCompatible (pSpec.Challenge i)] [oSpec.FiniteRange]
+  [∀ i, ToOracle (pSpec.Message i)]
+  (StmtIn WitIn : Type)
+  {ιₛₒ : Type} (OStmtOut : ιₛₒ → Type) [∀ i, ToOracle (OStmtOut i)]
+  (ε_sound : ℝ≥0) (ε_rbr : pSpec.ChallengeIndex → ℝ≥0) :
+  ∃ π : IOPP pSpec oSpec StmtIn WitIn OStmtOut ε_sound ε_rbr,
+    ε_rbr < 1 / 2 ^ secpar :=
+by
+  sorry
```
quangvdao: The theorem states the existence of some IOPP (with the structure given by `pSpec`, `oSpec`, `OStmtOut`) satisfying some `ε_rbr` round-by-round knowledge error (the soundness error should be omitted and derivable from round-by-round error).

This is not quite what we are looking for though. The issues are:

1. You need to define StmtIn WitIn OStmtOut to be the ones in the STIR protocol definition, and the relation need to be the STIR relation as well. And you can put `oSpec` to be `emptySpec` or just `[]\_0` with notation, to denote that this IOPP is not expected to require any auxiliary random oracles (which is the case for STIR).

Otherwise this is defined with respect to some nebulous parameters that may or may not have anything to do with STIR.

You may also want to specify pSpec to be the particular protocol structure of STIR. And the round-by-round knowledge error should be the quantity appearing in the theorem below.

2. Even then, you can still make trivial IOPPs (per the current definition) that passes this theorem. The trick is that, since the relation is defined _inside_ the IOPP, we can just state some trivial relation and provide some trivial verifier. Your definition of IOPP needs to bring out the relation as one of its parameters.

ArkLib/ProofSystem/Stir/Quotienting.lean
```diff
@@ -0,0 +1,81 @@
...
+lemma quotienting
+  {F : Type*} [Field F] [Fintype F] [DecidableEq F]
+  {L : Finset F}
+  {d : ℕ}
+  (S : Finset F) (hS_lt : S.card < d)
+  (C1 : ReedSolomonCode F L d)
+  (C2 : ReedSolomonCode F L (d - S.card))
+  (f : L → F)
+  (Ans Fill : S → F)
+  (δ : ℝ) (hδ : 0 < δ ∧ δ < 1)
```
quangvdao: Make delta be a NNReal for free positivity conditions

ArkLib/ProofSystem/Stir/ToCodingTheory/FracHammingDist.lean
quangvdao: Also, this should be NNRat to once again take advantage of free positivity.

ArkLib/ProofSystem/Stir/IOPP.lean
quangvdao: Actually, there is no need for the output relation (it should be trivial, just returning whether the output of the verifier is true or false). I'll make the definition available in ArkLib.

quangvdao:
Okay, I left a bunch of comments. The high-level ideas are:
1. Definitions should be more readable by using notation. The probability notation is not available yet, but we (and I mean myself) should define them.
2. We should use positive number types (NNRat, NNReal) instead of general types. This would simplify the statement & also help down the line with the security proofs.
3. The main theorem of STIR (and the definition of IOPP) needs to be refined to not be vacuous.
4. Consider switching to `Polynomial Polynomial` instead of `MvPolynomial (Fin 2)`.
5. Use ArkLib's current definitions & Katy's upcoming PR for some of the coding theory definitions.

We can take some time to make sure the PR is in good shape before merging.

ArkLib/ProofSystem/Stir/Folding.lean
quangvdao: Maybe something like this:
```
    (do let r ← PMF.uniformOfFintype F
        return (fractionalHammingDistSet (fold f k r) (C2.code) (C2.nonempty) ≤ δ.val)
    ).val True
      > err' F (d/k) C1.rate δ k
```
Do notation requires `Type` and not `Type*` though.

In the future it could be something like:
```
Pr_{ let r ←$ᵖ F }[ fractionalHammingDistSet (fold f k r) (C2.code) (C2.nonempty) ≤ δ.val ]
  > err' F (d/k) C1.rate δ k
```

quangvdao (changes requested):
Comments for PR

ArkLib/ProofSystem/Stir/Folding.lean
quangvdao: Also, when you define proofs in the middle of the definition, use `haveI` which will inline the proof. (you can check afterward by doing "#print [definition here]")

ArkLib/ProofSystem/Stir/Folding.lean
dpoulami: we agreed in the call with Quang that for now MVPolynomial should be fine, and we can keep a note about replacing it in the future. same is true for all Polynomial-related comments below.

ArkLib/ProofSystem/Stir/Folding.lean
dpoulami: we agreed to wait for a notation from Quang, and replace it when its ready.

ArkLib/ProofSystem/Stir/ToCodingTheory/ProximityBound.lean
dpoulami: @quangvdao min_val has a term which contains a subtraction, are you sure this will be always positive?

quangvdao: Well, perhaps not. But `err'` should be nonnegative regardless, right? There's no extra information to be gained about `err'` being negative, since what it is used for is in the probability statement `Pr_{r \gets F} [ some proximity claim ] \ge err'`. Then err' might as well be non-negative, and hence all the subtraction might as well be in `NNReal`.

PR #37, opened 2025-05-19
doc: add some uses commands

BoltonBailey:
This PR adds some `\uses` commands to the leanblueprint source, to try to avoid the blueprint dependency chart being a flat list of declarations.


quangvdao:
Thanks!

PR #42, opened 2025-05-24
feat: add blueprint details for sumcheck protocol / round protocol

BoltonBailey:
Adds details that should make the blueprint graph for sum-check more filled out


quangvdao:
Thanks! That section should really be fleshed out further, including all details. I'll try to get to it next week.

PR #44, opened 2025-05-26
feat: Inductive Merkle Trees with completeness theorem

BoltonBailey:
Define Merkle trees over an indexed inductive binary tree data type.

A reference for Merkle trees can be found in chapter IV of [the SNARGs book](https://github.com/hash-based-snargs-book/hash-based-snargs-book/blob/main/snargs-book.pdf). Note though, that this development generalizes to arbitrary binary tree structures.

The potential benefits of this are:

* We can commit to non-power-of-two sized tree with non-perfectly balanced Merkle trees
* The code is written in a more functional than one that uses `Vector`,  
* We perhaps avoid things like needing API in VCVio for monadic operations over vectors.

This PR includes:

* A low level file defining and providing API for indexed binary Merkle trees
* Definitions of the basic Merkle tree primitives for single-leaf proofs in the `OracleComp` monad
* The same primitives as pure functions with the hash passed explicitly
* Completeness theorem for single-leaf proofs
  * Currently depends on sorries relating `randomOracle` to `runWithOracle`. Not sure if VCVio is undergoing a change that will change what we need here.

I am marking this as ready for review, despite the presence of sorrys, since all the sorries are either not critical, or about VCVio definitions. Reviewers, please let me know if:

* Any sorries really ought to be finished before merging
* The PR should be split somehow


alexanderlhicks:
/review

External:
https://github.com/hash-based-snargs-book/hash-based-snargs-book/blob/main/snargs-book.pdf

Internal:

Comments:

alexanderlhicks:
Is there a good reference for inductive Merkle trees we could use here?

alexanderlhicks:
LGTM?

PR #45, opened 2025-05-26
feat: add theorem statements for MerkleTree completeness

BoltonBailey:
Adds theorems breaking down this execution.


BoltonBailey:
Note that I am also trying to work through the same logic for the inductively defined tree in #44

quangvdao:
Is this ready for merging?

BoltonBailey:
I have mainly been working on the inductive approach, but I feel this could simply be merged now, as it would be a good place to pick up if I decide I want to come back to this approach.

PR #51, opened 2025-06-02
Linear, Interleaved and Reed-Solomon codes + Berlekamp-welch.

Ferinko:
`CodingTheory/Basic.lean` - Integrated the definitions on linear codes and distances, including everything on relative Hamming distances, previously in `LinearCodes.lean` and `RelativeHamming.lean`. The generality of all statements has been adapted as per request. 

`CodingTheory/BerlekampWelch/*` - Implemented and proved correctness of the Berlekamp-Welch decoder with all relevant constructions, definitions and proofs.

`CodingTheory/DivergenceOfSets.lean` - Proved definition for divergence of arbitrary sets.

`CodingTheory/HammingDistance/Lemmas.lean` - Temporary place for a specific auxiliary statement that is pending restructuring.

`CodingTheory/InterleavedCodes.lean` - Interleaved codes: definitions, distances, relation between the notion of distance in an interleaved code and its underlying linear code. Statements of proximity results for Reed-Solomon codes (`Lemma 4.3`, `Lemma 4.4` and `Lemma 4.5` from Ligero) with proximity parameter less than the minimal code distance divided by `3`. 

CodingTheory/ListDecodability.lean - Proved definition of list decodability for codes.

`CodingTheory/Prelims.lean` - All needed preliminaries and misc. This might be moved elsewhere at some point. 

`CodingTheory/ReedSolomon.lean` - definitions and proofs involving non-square Vandermonde matrices and their ranks. Proof that the generator matrix of a Reed-Solomon code is a non-square Vandermonde matrix. Definition of length and dimension notions for RS codes. Statement and proof of the minimal distance theorem for Reed-Solomon codes, including definitions and proofs for needed constructions. Constructions on polynomials: eg obtaining polynomials from a vector of coefficients, extracting a vector of coefficients from a polynomial and basic simp lemmas ensuring everything works as expected.


Ferinko:
I've made many improvements to this branch in the past 2 days or so, albeit using 4.21.0-rc3. They can be found [here](https://github.com/NethermindEth/ZKLibFri/tree/Ferinko/moreCleanups).

I can sorry out some `aesop` proofs and push these improvements here if we don't want to bump `main` to a release candidate _and_ would like to merge this branch into main at some earlier point, but please do prompt me to do so upon review - my current strategy is to wait for main to use any version greater than or equal to 4.21.0-rc1, so that I don't have to 'break' some proofs only to then 'unbreak' them in the future.

PR #52, opened 2025-06-03
Bump to 4.20.0: **read description**

Ferinko:
This needs a bump to VCVio 4.20.0 which I executed on in a fork here: `https://github.com/Ferinko/VCV-io.git`.

This is reflected in `lakefile.lean` and needs changing to: `require VCVio from git "https://github.com/dtumad/VCV-io.git" @ "v4.20.0"`

assuming `VCVio` is modified first to include [this PR](https://github.com/dtumad/VCV-io/pull/72) + adding an appropriate tag on the commit `4a4934f796be6a44add080c439c875b99fec10e6`.


quangvdao:
Thanks, will do! This is the normal process for me anyway

PR #67, opened 2025-06-17
feat: define general inductive trees

quangvdao:
This PR refactors the MerkleTree file to add dedicated new files for inductively defined trees.

We define two variants: binary trees and general trees (arbitrary arities, represented by `List`). For each we also define a variant where the data are just at the leaves, and another variant where the data is present in each node.

The plan is to define Merkle trees on both kinds of trees, implementing an embedding for binary trees into general trees, and prove properties with regards to general trees.

@BoltonBailey this might mess with your current progress, but I hope not too much. I do think the general version is needed (i.e. to accommodate some Rust libraries with arity more than 2)


quangvdao:
Note: Claude 4 Sonnet wrote this entire PR. It managed to one-shot the general tree definitions.

PR #83, opened 2025-07-07
feat: Redefine `OracleVerifier` & lenses for oracle reductions

quangvdao:
1. We re-define the oracle verifier in an IOR, to have a general way of computing the output oracle statement, instead of hard-coding them to be a subset of the oracles so far.
- This hard-coding will no longer make sense when we switch to dependent oracles / poly functors, since then there won't be an indexed family of types (for oracle statement), just a single type
- More generally, we could think of two versions:
  - The minimal version is that, we only need to specify how oracle access to the output oracle statement can be simulated by oracle access to the input oracle statement & the prover's messages. Nothing needs to be materialized. We will need to generalize everything as well, including security definition, to account for the fact that the oracle verifier has no way to "see" or derive the underlying data of the output oracle statement
  - The more integrated version is to also provide a "reification" function, which is a mapping to output oracle statement _at the data level_, and also a proof that this reification is actually compatible with the simulation. This is the case for all IORs we know today (where the procedure is just taking subset, linear combination, etc. which are trivially reifiable)
- We chose to do reification by default. This simplifies a lot of definitions and proofs.


alexanderlhicks:
/review

External:

Internal:

Comments:

PR #85, opened 2025-07-12
[DRAFT] Fix: WHIR/STIR

alexanderlhicks:
Creating this PR to fix an incorrect statement in the WHIR formalisation:

- `lemma genMutualCorrAgreement_le_bound` (lemma 4.10 in the WHIR paper) is stated as 
`lemma genMutualCorrAgreement_le_bound
  (Gen : ProximityGenerator ι F) [Fintype Gen.parℓ]
  (BStar : ℝ) (errStar : ℝ → ENNReal)
  (C : Set (ι → F)) (hC : C = Gen.C)
  (h: genMutualCorrAgreement Gen BStar errStar) :
    BStar < min (1 - (δᵣ C) / 2 : ℝ) (Gen.B Gen.C Gen.parℓ)
    ∧
    errStar = Gen.err Gen.C Gen.parℓ := by sorry`
    This appears to allow arbitrary BStar and errStar for which mutually correlated agreement is assumed to hold, which does not match the paper. @katyhr, @Ferinko, and @ElijahVlasov, show that the negation of the lemma stated this way [here](https://github.com/NethermindEth/ArkLibFri/blob/63bd02c7eaf9dc8a8f3b490b1dc212692ebfb1fa/ArkLib/ProofSystem/Whir/GenMutualCorrAgreement.lean#L65).
- My first pass at fixing this is to instead state the lemma as 
`lemma genMutualCorrAgreement_le_bound_fixed
  (Gen : ProximityGenerator ι F) [Fintype Gen.parℓ]
  (C : LinearCode ι F) (hC : C = Gen.C) :
    genMutualCorrAgreement Gen
      (min (1 - (Code.minDist (C : Set (ι → F)) : ℝ) / 2) (Gen.B C Gen.parℓ))
      (fun δ => Gen.err C Gen.parℓ δ) := by sorry`
In this case, BStar and errStar are constrained by instantiating them using the definition of Gen (definition 4.9 in the paper). I've also replaced the minimum relative distance `δᵣ C` with the minimum distance `Code.minDist (C : Set (ι → F)`  -- there's possibly a nicer way of satisfying the typechecker -- as the paper refers to minimum distance.
- I'm opening this PR so that I don't forget to do so, and would like to resolve this asap (ideally ahead of or during Monday's regular call), but given it's 1am on Saturday there's a chance I've also misunderstood something and that my proposed fix might contain something dumb (I've just quickly looked at the lemma itself and `def genMutualCorrAgreement`). I'm also aiming to identify any other similar issues in the formalization (if any) asap.

Many thanks to @katyhr @Ferinko @ElijahVlasov and Petar for flagging this and explicitly demonstrating the issue.
cc @dpoulami


dpoulami:
Thank you for pointing out this bug in Lemma 4.10. As pointed out by Alex in the call, I suspect this bug to have propagated in the following lemmas too. I agree we should work on fixing this asap!

On a broader note, please note the stir/whir blueprint is still missing a thorough review (either internally from our team or externally) and cannot be considered completely polished! I and @MQuaresma will work on a thorough check of the repos asap. We also welcome any feedback from external contributors in the meantime.

cc @alexanderlhicks

PetarMax:
We have understood what the issue is. 

In the [WHIR paper](https://eprint.iacr.org/2024/1586.pdf) itself, there is a typo in Definition 4.9. In particular, it should be `\delta \in (0, 1 - B*(...))` instead of `\delta \in (0, 1 - B(...))`. This is correctly flagged by Lean through a warning that `BSter` is an unused variable.

We have confirmed this with Giacomo Fenzi, one of the authors of WHIR, who also directed us to their [WARP paper](https://eprint.iacr.org/2025/753.pdf), which should contain a typo-free, but slightly different versions of the WHIR definitions, which could be used for cross-checking the WHIR definitions.

alexanderlhicks:
Just pushing some changes to the branch as a checkpoint (sorry for the delay, a bit staggered in between other things).

dpoulami:
In the [blueprint latex](https://verified-zkevm.github.io/ArkLib/blueprint/chap-proof_systems.html#a0000000018) you can already see definition 161 has Bstar instead of B. There are also differences in indexing for (0,..,m-1) instead of (1,..,m) and a generalization of the definition of Block relative distance that was required for Lean formalization. I can add a note/color to highlight the differences from the paper more explicitly.

PetarMax:
Yes, I think that highlighting these details brings additional value to any formalisation.

alexanderlhicks:
Is there a list of identified typos/changes to the paper (beyond going through each definition/theorem to identify which have been changed)?

alexanderlhicks:
The Last bit of the rbr soundness theorem previously stated:
```
-- ε_fin ≤ (1 - δ_{M})^(repeatParam_{M})
ε_fin ≤ 1 - (h.δ (Fin.last M))^(P.repeatParam (Fin.last M))
```
The paper states:
`ε^fin ≤ (1 − δ_{M−1})^t_{M−1} `.

I've tentatively adjusted the Lean formalization accordingly, can you confirm whether this was a typo in the paper (and, therefore, intended to written down this way in Lean) or a typo in Lean?

dpoulami:
minimum distance `\delta_C` is always minimum relative Hamming distance, as confirmed with Giacomo. So it should `\delta_r C` in the code instead of `Code.minDist`

ArkLib/ProofSystem/Whir/GenMutualCorrAgreement.lean
```diff
@@ -88,53 +88,55 @@ noncomputable def gen_α (α : F) (parℓ : Type) (exp : parℓ → ℕ): F →
...
-    -- Fintype instance of parℓ_type
     hℓ := inferInstance,
-    -- Generator function
     Fun := gen_α α parℓ_type exp,
-    -- BStar
-    B := fun _ _ => (1 + LinearCode.rate (smoothCode φ m)) / 2,
-    -- errStar
-    err := fun _ _ δ => ENNReal.ofReal
-        ((Fintype.card parℓ_type - 1) * (2^m / (LinearCode.rate (smoothCode φ m) * (Fintype.card F)))),
+    -- B from proximity gap theorem
```
dpoulami: As confirmed by Giacomo, `Gen(l,\alpha)` is exactly as Definition 4.7 `Gen`, with the difference that `\alpha` is used to generate `r_0,..,r_{l-1}` in Gen. So the B and err should have a generic form like in Gen.

ArkLib/ProofSystem/Whir/GenMutualCorrAgreement.lean
```diff
@@ -54,6 +57,14 @@ lemma genMutualCorrAgreement_le_bound
     ∧
     errStar = Gen.err Gen.C Gen.parℓ := by sorry
 
+-- Possible fix
+lemma genMutualCorrAgreement_le_bound_fixed
+  (Gen : ProximityGenerator ι F) [Fintype Gen.parℓ]
+  (C : LinearCode ι F) (hC : C = Gen.C) :
+    genMutualCorrAgreement Gen
+      (min (1 - (Code.minDist (C : Set (ι → F)) : ℝ) / 2) (Gen.B C Gen.parℓ))
```
dpoulami: it should be `\delta_r C` instead of `Code.minDist` since we are always referring to the minimum *relative* Hamming distance (already discussed)

alexanderlhicks: Yeah I thought I had commited this already but it's in the stashed stuff I have (following from your comment last week). Thanks!

dpoulami:
I will adjust the blueprint at some point today, a lot of definitions were renamed leading to the blueprint build failing.

alexanderlhicks:
> I will adjust the blueprint at some point today, a lot of definitions were renamed leading to the blueprint build failing.

No need to adjust the blueprint until the last moment I think, in case other definitions change in the meantime.

dpoulami:
okay that makes sense.

PR #95, opened 2025-07-27
feat: proofs for MlPoly

chung-thai-nguyen:
[x] Basic.lean
[x] Equiv.lean


alexanderlhicks:
You should now be able to trigger a review @chung-thai-nguyen.

PR #96, opened 2025-07-30
Define unique decoding

alexanderlhicks:
This PR just adds a definition of unique decoding as a special case of list decoding, and is meant to serve as a quick test of the now added ai review workflow.


alexanderlhicks:
@quangvdao Can you test you have permission to run a review? You can just copy the command above and/or push some dummy PR with an error in it if you want to try anything out as well.

alexanderlhicks:
Ah yeah, I'm on my phone right now but I might just add the unique decoding radius lemmas tonight or merge it as is if I don't do it tonight.

PR #98, opened 2025-08-01
feat: new `Fin` functions with better def'eq, refactor `ProtocolSpec`, and more

quangvdao:
This PR gives a few quality-of-life improvements to the experience of composing oracle reductions, motivated by getting sumcheck to be more modular.

1. **Problem:** current `Fin` functions such as `Fin.append` offers almost NO definitional equality. Even something as simple as `![1] ++ ![2] = ![1, 2]` would fail. If not fixed, this would lead to an inordinate amount of casting everywhere in `OracleReduction`. We have to always cast the protocol specification, and more.

The immediate solution is to redefine al the `Fin` functions like `cons/snoc/append` so that definitional equality holds. This takes a _lot_ of effort, but finally I made this work.

2. As a consequence, the general sequential composition case now follows from the base case (`id`) plus the successor case (`append`).

3. However, there are problems with dependent types that do not go away no matter what, i.e. `Fin.append` is not associative on the nose, since it would require `m + (n + p) = (m + n) + p`.

An experimental but potentially promising solution to this is to redefine the type of `Nat` and `Fin` itself so that associativity holds definitionally. This is possible, at least for natural numbers, for a function-based definition of the natural numbers as `Nat \to Nat`, or a Church encoding.

The experimental stuff is still ongoing. Probably won't be merged within this PR, but we keep around the files for further development.

4. We unbundle the definition of `ProtocolSpec` to have two separate fields, one for the directions and one for the Type. This leads to better definitional equalities.

5. We slightly generalize the definition of the prover, in anticipation for the upcoming introduction of a session-type-based definition. In particular, the malicious provers in soundness & knowledge soundness do not require the full API of an honest prover.

6. We flesh out the casting infrastructure, stating theorems that they preserve execution semantics & security properties. But this is still TODO, a lot to be done.

7. We re-define the simple single round of sumcheck to be a sequential composition of multiple 1-round & 0-round oracle reductions. This is still ongoing.

This is a huge PR so I'm sure there's some changes I'm not mentioning here. The AI review bot should catch them.


alexanderlhicks:
/review

External:

Internal:
ArkLib/

Comments:

alexanderlhicks:
> Ah this is fantastic. I don't even need to write descriptions of PR!

It ensures all PRs will be marked as "excellent" :smile: but yes, the next thing I want to setup is a second bot to add documentation (with a specific style guide and some way to avoid it blabbing about too much) once a PR is ready to be merged, and it can also provide a nice commit/merge message (with Gemini CLI retrieving the last n PRs and having it summarise them is not bad UX).

alexanderlhicks:
> Can you add system prompts to the bot? Like "be direct and concise, don't flatter".

You can tweak the system prompt in scripts/review.py and experiment, asking it to be more concise runs the risk it will ignore some issues (i.e., it sometimes focus on obvious issues like sorrys and doesn't look at whether the actual statement is sensible).

quangvdao:
/review

External:

Internal:
ArkLib/

Comments:
Read the prior AI review first, do not mention anything that already gets said in the prior review. If anything has changed since then, do mention.

PR #106, opened 2025-08-25
`UniPoly` equivalence relation and quotient operations

dhsorens:
## Summary 
This PR:
- [x] proves the equivalence relation defined on `UniPoly` is transitive, using an (assumed) lemma analogous to the (also unproved) lemma used in the List case
- [x] defines functions `add`, `neg`, and `sub` on the resulting quotient and proves that the equivalent operations on `UniPoly` descend

From here, the main TODO would be to show that the remaining operations on `UniPoly` descend to the quotient.

## Related Issue
Related to #7

### DIFF
--- a/ArkLib/Data/UniPoly/Basic.lean
+++ b/ArkLib/Data/UniPoly/Basic.lean
@@ -897,7 +897,6 @@ open Trim
   intro h i
   exact Eq.symm (h i)
 
-open List in
 /-- Transitivity of the equivalence relation. -/
 @[simp] theorem equiv_trans {p q r : UniPoly Q} : Trim.equiv p q → equiv q r → equiv p r := by
   simp_all [Trim.equiv]


dhsorens:
Is any reason to keep the old definition around, possibly proving functional equivalence?

quangvdao:
Awesome! What is the old definition you're referring to?

ArkLib/Data/UniPoly/Basic.lean
```diff
@@ -875,32 +885,22 @@ lemma eval_trim_eq_eval [LawfulBEq R] (x : R) (p : UniPoly R) : p.trim.eval x =
 end ToPoly
 
 section Equiv
-
-/-- An equivalence relation `equiv` on `UniPoly`s where `p ~ q` iff one is a
-zero-padding of the other. -/
-def equiv (p q : UniPoly R) : Prop :=
-  match p.matchSize q 0 with
-  | (p', q') => p' = q'
```
dhsorens, line 883: @quangvdao the old definition is this one here -- two polynomials are equivalent if, when zero-padded into arrays of the same size, they are equal

quangvdao: Ah yeah. sounds good to me

Also we should think about which of `QuotientUniPoly` or `UniPolyC` should be "the" way to get equivalence with polynomial. The non-quotient version might be preferable?

dhsorens: From what I can tell, that probably depends on what the vision of this repo is (I'm familiar with it, but still new here!). Aesthetically, I quite like `QuotientUniPoly` -- but `UniPolyC` is likely more computation-friendly? Considering this is a repo directly aimed at snarkification/formal verification we might favor the computation (non-quotient) in the long run

dhsorens:
gh actions made a good point on the proofs of `add_descends` and `sub_descends` so I have improved them using `calc` notation and pushed. I would be very happy to address the proof of `add_coeff` next but it's not a theorem introduced in this PR so I would argue not related to this one (should ofc be done soon though)

PR #137, opened 2025-08-29
Model of the FRI-IOPP

Julek:


ArkLib/Data/GroupTheory/Smooth.lean
```diff
@@ -0,0 +1,11 @@
+import Mathlib.GroupTheory.SpecificGroups.Cyclic
+
+class IsCyclicC (G : Type) [Pow G ℤ] where
```
quangvdao: Can you add documentation for this type class and also `Smooth2`?

I'm a bit surprised this isn't already a type class in mathlib. Maybe the one in mathlib doesn't have the generator as a field in the type class?

quangvdao: Ah, seems like `IsCyclic` type class in mathlib bundles the generator in an existential statement.

We should clarify this. Also consider more descriptive naming like `IsCyclicWithGen`

ArkLib/Data/GroupTheory/Smooth.lean
```diff
@@ -0,0 +1,11 @@
...
+  gen : G
+  zpow_surjective : Function.Surjective (gen ^ · : ℤ → G)
+
+instance {G} [Pow G ℤ] [inst : IsCyclicC G] : IsCyclic G where
+  exists_zpow_surjective := ⟨inst.gen, inst.zpow_surjective⟩
+
+class Smooth2 (n : ℕ) (G : Type) [Pow G ℤ] [Monoid G] [inst : IsCyclicC G] where
```
quangvdao: Name suggestion to avoid numeral in the class: `SmoothPowerOfTwo`

ArkLib/ProofSystem/Fri/Domain.lean
```diff
@@ -0,0 +1,280 @@
...
+
+variable {F : Type} [NonBinaryField F]
+variable (D : Subgroup Fˣ) {n : ℕ} [DIsCyclicC : IsCyclicC D] [DSmooth : Smooth2 n D]
+
+namespace Fri
+
+namespace Domain
+
+@[simp]
+def evalDomain (i : ℕ) : Subgroup Fˣ :=
```
quangvdao, line 22: Please add documentation to all definitions.

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,272 @@
...
+/-!
+# The FRI protocol
+
+We describe the FRI oracle reduction as a composition of many single rounds, and a final
+(zero-interaction) query round where the oracle verifier makes all queries to all received oracle
+codewords. f` is the sum over `a : α`
+  of the probability of `a` under `p` times the measure of the set under `f a`. -/
+
+/-- The (default) oracle interface for a function `α → β`. -/
+instance {α β : Type*} : OracleInterface (α → β) :=
```
quangvdao: This is now already defined in `OracleInterface.lean` as `OracleInterface.instFunction`

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,272 @@
...
+namespace FoldPhase
+
+/-- The oracle interface for the `j`-th oracle statement of the `i`-th round of the FRI protocol.
+
+Since everything are functions right now, we just use the default oracle interface for functions. -/
+instance {i : Fin (k + 1)} : ∀ j, OracleInterface (OracleStatement D x i j) :=
+  fun _ => instOracleInterfaceForall
+
+/-- This is missing the relationship between the oracle statement and the witness. Need to define a
+  proximity parameter here. Completeness will be for proximity param `0`, while soundness will have
```
quangvdao: Flagging this so we don't merge before we properly define the relation

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,272 @@
...
+  the verifier. -/
+@[reducible]
+def pSpec : ProtocolSpec 2 := ⟨!v[.V_to_P, .P_to_V], !v[F, (evalDomain D x i.1) → F]⟩
+
+instance {i : Fin k} : ∀ j, OracleInterface ((pSpec D x i).Message j)
+  | ⟨0, h⟩ => nomatch h
+  | ⟨1, _⟩ => instOracleInterfaceForall
+
+/-- The prover for the `i`-th round of the FRI protocol. It first receives the challenge -/
+noncomputable def foldProver :
```
quangvdao: We need to add comprehensive docstring to the prover.

As of right now, it is hard to read & reason about - for instance, we need to spot that the prover updates the polynomial witness after receiving every challenge, by folding with the received challenge

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,272 @@
...
+          if h : j.1 < i.1
+          then by
+            simpa [OracleStatement, evalDomain] using o ⟨j.1, h⟩
+          else fun x => p.eval x.1.1
+      ⟩,
+      p
+    ⟩
+
+/-- The oracle verifier for the `i`-th round of the FRI protocol. -/
+noncomputable def foldVerifier :
```
quangvdao: Same with docstring for what the verifier does

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,272 @@
...
+  verifier := foldVerifier D x i
+
+end FoldPhase
+
+namespace QueryRound
+
+variable (l : ℕ)
+
+@[reducible]
+def pSpec (F : Type) : ProtocolSpec 1 := ⟨!v[.P_to_V], !v[Unit → F]⟩
```
quangvdao: I think the query round has the wrong semantic.

As written, the prover sends a single field element, and the verifier queries the shared oracle (configured to give randomness), then makes the queries & verify.

First, I think the field element sent by the prover is redundant - for the last round in the fold phase (if it's executed fully for n times), the prover already sends this element.

Actually, for the final round (if there is early stopping), the prover should send the final _polynomial_, not the RS codeword corresponding to that polynomial.

Second, there should be _no_ shared oracle in the FRI protocol. It's a pure IOP(P). In this final round, the verifier should send all the queries it would make to the prover.

By the execution semantic of the IOP, these queries will be random.

This way, the prover knows what queries the verifier makes. Without this knowledge, the verifier is private-coin and we cannot compile this into an interactive argument using the BCS transform.

quangvdao: So concretely, `QueryRound` should just be `\< !v[ .V_to_P ], !v[ Fin numRepetitions \to F] \>`.

And we should think about early stopping and the validation checks that would impose on the verifier (needs to make sure it is a Reed-Solomon codeword). Alternatively, the prover needs to send the underlying folded polynomial itself.

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
Julek: Understood, on it! :slightly_smiling_face:

ArkLib/ProofSystem/Fri/Domain.lean
Julek: :saluting_face: :saluting_face: :saluting_face: I will handle this before un-drafting the PR, for the moment I'm just focused on getting the semantics right :grin:

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
Julek: :pray:

Julek: @quangvdao Is this on the current `main` branch? I can't seem to find it.

quangvdao: Just pushed this morning, if you could merge the latest main onto your branch

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
Julek: Understood re: IOPP, will make the change. 

Also, we've already implemented early stopping, the model does `k` rounds, where `k ≤ n`. When it finalises, it produces an Oracle statement of the form `OracleStatement D x (Fin.last k)`, which only contains oracles for codewords up to `evalDomain D x (k - 1)`, see:
```
def OracleStatement (i : Fin (k + 1)) : Fin i.1 → Type :=
  fun j => evalDomain D x j.1 → F
```

So sending this final constant is necessary I think, as, in particular, `evalDomain D x (k - 1)` will never be the trivial subgroup of `Fˣ`. We could generalise this to return a polynomial of strictly less than degree `d`, which would then allow the proof system to prove that the original polynomial was of degree strictly less than `2^k * d`.

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
Julek: :pray:

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,274 @@
...
+    pure prevChallenges
+  embed :=
+    ⟨
+      fun (j : Fin k) =>
+        Sum.inl <| by simpa using j,
+      by intros _; aesop
+    ⟩
+  hEq := by
+    unfold OracleStatement pSpec
+    aesop
```
quangvdao, line 657: After the verifier makes all the queries, I think we can drop all the prior codewords (and same can be done on prover side).

In other words, verifier just returns an `OracleComp emptySpec Unit`.

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,274 @@
...
+  | ⟨0, _⟩ => fun ⟨⟨chals, o⟩, p⟩ => do
+      pure ⟨fun _ => p.coeff 0, ⟨⟨chals, o⟩, p⟩⟩
+  | ⟨1, h⟩ => nomatch h
+  receiveChallenge
+  | ⟨0, h⟩ => nomatch h
+  | ⟨1, _⟩ => fun x => pure <| fun _ => x
+
+  output := pure
+
+def sampleCodeword (i : Fin k) (d : evalDomain D x i.1) :
```
quangvdao: The name here seems a bit inaccurate: we are asking for the value of the `i`-th codeword at position `d`.

A better name would be something like `queryCodeword`.

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,274 @@
...
+    OracleComp [OracleStatement D x (Fin.last k)]ₒ F :=
+  OracleComp.lift (OracleSpec.query i d)
+
+def getConst : OracleComp [(pSpec D x l).Message]ₒ F :=
+  OracleComp.lift
+    (by simpa using
```
quangvdao: This is interesting; I wonder why you have to use `simpa` here

quangvdao: Just unfolding the definition might work?

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,274 @@
...
+  verify := fun prevChallenges roundChallenge => do
+    for m in (List.finRange l) do
+      let s₀ := roundChallenge ⟨1, by aesop⟩ m
+      discard <|
+        (List.finRange k).mapM
+              (fun (i : Fin k) =>
+                do
+                  let x₀ := prevChallenges i
+                  let s₀ : evalDomain D x i.1 := ⟨_, pow_2_pow_i_mem_Di_of_mem_D D x i s₀.2⟩
+                  let s₁ := (domain_neg_inst (n := n) (i := ⟨i.1, by omega⟩) D x).neg s₀
```
quangvdao: Can we define this better so that we can write just `- s₀`, `neg s₀` or something concise?

Julek: This will be changed completely as we move towards the generalised round consistency check :grin:

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
Julek: Wouldn't it be good to keep this to encode schemes like DEEP-FRI that will need access to these oracles later I believe?

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,777 @@
...
+  fun j =>
+    if j.1 = k + 1
+    then (Unit → F[X])
+    else (evalDomain D x (s * j.1) → F)
+
+/-- The FRI protocol has as witness the polynomial that is supposed to correspond to the codeword in
+  the oracle statement. -/
+@[reducible]
+def Witness (F : Type) [Semiring F] {k : ℕ} (s d : ℕ) (i : Fin (k + 2)) :=
+  F⦃< 2^(((k + 1) - i.1) * s) * d⦄[X]
```
quangvdao, line 78: I think the witness should not be dependent on the round number, it should stay the same throughout.

The reason is that we also append the oracle statements as we go along, and so the original one is always present.

You need the witness to correspond to the original oracle statement, and the i-th folded witness (derived from the original one) to correspond to the i-th oracle sent in the i-th round.

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,777 @@
...
+        )
+        (2 ^ (n - s * i))
+  let enum : Fin (2 ^ (n - ↑ind)) → ↑(evalDomain D x (s * ↑(Fin.last ↑i))) := by
+    simpa [ind] using (CosetDomain.domain_enum D x ind).1
+  {
+    ⟨⟨stmt, ostmt⟩, p⟩ |
+      statementConsistent cond stmt ostmt ∧
+      ∀ x, ostmt (Fin.last i.1) x = p.1.eval x.1.1 ∧
+      δᵣ(ostmt (Fin.last i.1) ∘ enum, code) < δ
+  }
```
quangvdao: Following the comment made above on the witness, we need to enforce proximity of the i-th folded witness to the i-th oracle statement for all i (including i = 0 i.e. initial witness).

Also, the condition `∀ x, ostmt (Fin.last i.1) x = p.1.eval x.1.1` seems to imply that the oracle statement is already _exactly_ the evaluation of the witness? Perhaps this should not be the semantics.

Finally, `δᵣ(ostmt (Fin.last i.1) ∘ enum, code) < δ` should be a `<=`, otherwise it never holds when `δ = 0`.

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
```diff
@@ -0,0 +1,777 @@
...
+        )
+        (2 ^ (n - s * i.succ))
+  let enum : Fin (2 ^ (n - ind)) → ↑(evalDomain D x (s * i.succ)) := by
+    simpa [ind] using (CosetDomain.domain_enum D x ind).1
+  {
+    ⟨⟨stmt, ostmt⟩, p⟩ |
+      statementConsistent cond stmt ostmt ∧
+      ∀ x, ostmt (Fin.last i.succ) x = p.1.eval x.1.1 ∧
+      δᵣ(ostmt (Fin.last i.succ) ∘ enum, code) < δ
+  }
```
quangvdao: Same comment as with `inputRelation`

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
quangvdao: I take back my suggestion that i-th folded witness needs to be close to the i-th oracle statement for all i. This might not work in the list-decoding regime I think.

The best way is to define the relation according to the definition of the doomed set (Lemma 5.2) in the FRI analysis paper: https://eprint.iacr.org/2023/1071.pdf.

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
Julek: Are you suggesting a formulation of the protocol where the next putative codeword is computed at each round based purely on the previous one, and the folded polynomials are not explicitly passed along? 

How would the previous codeword be passed along? I guess as part of the witness? As we would not want the prover to have to query the previous oracle at all points to construct the next putative codeword?

ArkLib/ProofSystem/Fri/Spec/SingleRound.lean
Julek: Fixed I believe :smile: Thanks for the review!

alexanderlhicks:
/review

External:
https://eprint.iacr.org/2022/1216.pdf
https://eprint.iacr.org/2021/582.pdf
https://eprint.iacr.org/2019/336.pdf
https://eprint.iacr.org/2020/654.pdf

Internal:
ArkLib/Data/CodingTheory/
ArkLib/ProofSystem/Fri
ArkLib/Data/Polynomial/
ArkLib/OracleReduction/

Comments:
Please review and identify any possible issues or possible improvements. Note that this is an attempt to formulate FRI using the framework of interactive oracle reductions.

PR #145, opened 2025-09-03
Section 5 of the Proximity Gap paper

ElijahVlasov:


ArkLib/Data/Polynomial/Bivariate.lean
```diff
@@ -1,252 +1,560 @@
...
+   individual `X`-degrees.
+   - We define and prove some basic properties about quotients of bivariate polynomials.
+   - We define and prove some basic properties of monomials of bivariate polynomials.
+
+-/
 
 open Polynomial
 open Polynomial.Bivariate
 
-namespace Polynomial
```
quangvdao, line 14: Why are we changing to `Bivariate` instead of `Polynomial.Bivariate` here?

ArkLib/Data/Polynomial/Bivariate.lean
```diff
@@ -1,252 +1,560 @@
...
+  : (f.eval (C x)).eval y = 0 := by
+  sorry
+
+open Univariate in
+/-- In the case of a bivariate polynomial we cannot easily use `discriminant`.
+   We are using the fact that the resultant in question is always
+   divisible by the leading coefficient of the polynomial.
 -/
-lemma degreeY_mul [IsDomain F] (hf : f ≠ 0) (hg : g ≠ 0)
+opaque discr_y {F : Type} [CommRing F] (f : F[X][Y]) : F[X] :=
```
alexanderlhicks: Can I ask why this ended up defined this way / opaque?
Since resultant_is_divisible_by_leadingCoeff makes use of Polynomial.resultant, couldn't this use [Polynomial.discr](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Polynomial/Resultant/Basic.html#Polynomial.discr) or adapt it in some way to avoid opaque + choose? (I see the comment right above this, but it isn't clear to me )

ArkLib/Data/CodingTheory/ProximityGap.lean
```diff
@@ -143,4 +162,435 @@ theorem correlatedAgreement_affine_spaces {k : ℕ} [NeZero k] {u : Fin (k + 1)
...
+structure ModifiedGuruswami
+  (m n k : ℕ)
+  (ωs : Fin n ↪ F)
+  (Q : F[Z][X][Y])
+  (u₀ u₁ : Fin n → F)
+  where
+  Q_ne_0 : Q ≠ 0
+  /-- Degree of the polynomial. -/
+  Q_deg : Bivariate.natWeightedDegree Q 1 k < D_X ((k + 1) / (n : ℚ)) n m
+  /-- Multiplicity of the roots is at least r. -/
```
alexanderlhicks: Should be "Multiplicity of the roots is at least m" in this case?

ArkLib/Data/CodingTheory/ProximityGap.lean
```diff
@@ -143,4 +162,435 @@ theorem correlatedAgreement_affine_spaces {k : ℕ} [NeZero k] {u : Fin (k + 1)
...
+  (h_gs : ModifiedGuruswami m n k ωs Q u₀ u₁)
+  : Finset F := (lemma_5_5 k h_gs (δ := δ)).choose
+
+/-- `S'` is indeed a subset of `S` -/
+lemma the_S'_subset_the_S
+  (h_gs : ModifiedGuruswami m n k ωs Q u₀ u₁)
+  : the_S' k ωs δ u₀ u₁ h_gs ⊆ the_S k ωs δ u₀ u₁ := by sorry
+
+/-- The equation 5.12 from the proximity gap paper. -/
+lemma eq_5_12
```
alexanderlhicks: Is there a nicer way to name eq_5_12, lemma 5_6, lemma 5_7, ...? maybe some llm can give helpful suggestions that will make these clearer in the long run (e.g. if used outside the context of the proximity gaps paper for whatever reason).

ArkLib/Data/Polynomial/Prelims.lean
```diff
@@ -0,0 +1,89 @@
...
+
+namespace Ideal
+section
+
+/-- A principal ideal is an ideal generated by a single element. -/
+def principalIdeal {F : Type} [Semiring F] (f : F) : Ideal F := Ideal.span {f}
+
+/-- A principal ideal in a polynomial ring is maximal if and only if its generator is
+an irreducble polynomial. -/
+lemma principal_is_maximal_iff_irred {F : Type} [CommRing F] (f : F[X]) :
```
alexanderlhicks: Should this be restricted to a field?

alexanderlhicks (commented):
First pass at reviewing some of the PR:

alexanderlhicks:
(if anyone has a recommendation for a freely available linear algebra textbook we could use as standard reference for the review bot, lmk)

ArkLib/Data/Polynomial/Bivariate.lean
ElijahVlasov: `opaque` is an artifact from the time this def was sorried out. `Polynomial.discr` isn't available in the mathlib version we're using and I'm not courageous enough to bump the mathlib version

alexanderlhicks: Right, the bump to 4.24 should take care of that then!

alexanderlhicks: Could you add a todo comment as I expect this PR might be merged before the bump to 4.24?

ElijahVlasov: Done!

PR #263, opened 2025-10-20
KZG Correctness + Function Binding

tobias-rothmann:
- [x] Update the KZG to computable polynomials
- [x] Update `CommitmentScheme` to use a KeyGen function
- [x] Instantiate the KZG as a CommitmentScheme
- [x] Prove Correctness for the KZG (wire-up to `CommitmentScheme` Correctness)
- [x] Formalize Function Binding in `CommitmentScheme`
- [x] Formalize ARSDH
- [x] Formalize the Reduction 
- [x] Prove Function Binding for the KZG


alexanderlhicks:
Thanks for the PR!
Just a heads up that we should be splitting off computable polynomials into their own library. Hopefully this should happen sooner rather than later and will then require you to update your dependencies accordingly.
We also have a review bot available which can take in both external references (html or pdf) and internal (specific sections of ArkLib) and a custom prompt. This is sometimes able to catch (mis-)formalization issues and/or make helpful style suggestions, and works well as an intermediary review step for work in progress pending a manual review. Let us know if/when that would be helpful.

ArkLib/Data/UniPoly/Basic.lean
```diff
@@ -882,6 +882,68 @@ theorem eval_toImpl_eq_eval [LawfulBEq R] (x : R) (p : R[X]) : p.toImpl.eval x =
 lemma eval_trim_eq_eval [LawfulBEq R] (x : R) (p : UniPoly R) : p.trim.eval x = p.eval x := by
   rw [← toImpl_toPoly, eval_toImpl_eq_eval, eval_toPoly_eq_eval]
 
+-- TODO is this the right place?
+/-- degree is the mathlib degree + 1 -/
```
quangvdao: This seems fine to me. Eventually we could have a `Lemmas.lean` file inside `UniPoly` folder for stuff like this

ArkLib/CommitmentScheme/KZG.lean
```diff
@@ -75,32 +77,101 @@ theorem commit_eq {a : ZMod p} (hpG1 : Nat.card G₁ = p)
...
 -- e ( C / g₁ ^ v , g₂ ) = e ( O , g₂ ^ a / g₂ ^ z)
 
 -- theorem correctness {g : G} {a : ZMod p} {coeffs : Fin n → ZMod p} {z : ZMod p} :
+theorem correctness (hpG1 : Nat.card G₁ = p) (n : ℕ) (a : ZMod p) (coeffs : Fin (n + 1) → ZMod p) (z : ZMod p) :
+  let poly : UniPoly (ZMod p) := UniPoly.mk (Array.ofFn coeffs)
+  let v : ZMod p := poly.eval z
+  let (psrs,vsrs) : Vector G₁ (n + 1) × Vector G₂ 2 := generateSrs (g₁:=g₁) (g₂:=g₂) n a
+  let C : G₁ := commit psrs coeffs
+  let opening: G₁ := generateOpening psrs coeffs z
+  verifyOpening pairing (g₁:=g₁) (g₂:=g₂) vsrs C opening z v := by
```
quangvdao: Thanks! Would be nice to wire this up as a special case of correctness for `CommitmentScheme`

alexanderlhicks:
For some reason some new dependency was required ("packaging", presumably stopped being important by something else) which was messing with the summary/review scripts but it should be fixed now.

tobias-rothmann:
/review

External:
https://www.iacr.org/archive/asiacrypt2010/6477178/6477178.pdf

Internal:
ArkLib/Data/UniPoly/Basic.lean
ArkLib/CommitmentScheme/
ArkLib/OracleReduction/

Comments:
Please review with a focus on the use of the ArkLib machinery to instantiate the KZG as a Commitment Scheme, especially the opening Reduction, and the correctness proof. Focus on the correctness of the formalization and the proof with respect to ArkLib's Reduction and the original paper, and indicate any potential issues or improvements that could be made.

ArkLib/CommitmentScheme/Basic.lean
```diff
@@ -57,29 +57,34 @@ noncomputable section
...
-    [ fun x => x.2.1 | do
-        let cm ← liftComp (scheme.commit data randomness) _
-        let ⟨result, _⟩ ← scheme.opening.run ⟨cm, query, O.answer data query⟩ ⟨data, randomness⟩
-        return result] ≥ 1 - correctnessError
+    ∀ data : Data,
+    ∀ randomness : Randomness,
+    ∀ query : O.Query,
+      let rel : Set ((Commitment × O.Query × O.Response) × (Data × Randomness))
+        := {((cm, query, answer), (data, randomness))|
+          answer = O.answer data query ∧ cm ∈ (scheme.commit data randomness).support}
```
quangvdao: I think that here, we need to define a combined game where:
- (honest) prover first commits to the data
- prover and verifier then runs the opening protocol for the functional query
At the end, prover & verifier needs to agree w.r.t output statement, which is to accept

To achieve this, you probably need to refactor the completeness definition. Hoist out the game into a new definition `completenessGame`, which then allows you to tack on the commitment phase before the game.

ArkLib/CommitmentScheme/KZG.lean
```diff
@@ -75,32 +111,237 @@ theorem commit_eq {a : ZMod p} (hpG1 : Nat.card G₁ = p)
...
+  The scheme takes a pregenerated srtuctured reference string (srs) for the
+  commiter and the verifier (generated by `generateSrs`).
+
+  - `commit` : a function that commits to an `n + 1`-tuple of coefficients `coeffs`
+  (corresponding to a polynomial of maximum degree `n`)
+  - `opening` : a non-interactive reduction (i.e. soly the committer sends a single
+  message) to prove the evaluation of the commited polynomial at a point `z`. The
+  message from the prover is the witness for the evaluation.
+-/
+def kzg (srs : Vector G₁ (n + 1) × Vector G₂ 2) :
```
quangvdao: Minor nit but you can make this all caps `KZG`

ArkLib/CommitmentScheme/Basic.lean
tobias-rothmann: That was my first thought as well, seems like I should've stuck to it. Will change it ASAP!

tobias-rothmann:
Finally ready for review @quangvdao @alexanderlhicks :) 
I think the PR summary does a pretty good job -- if you need more information or want a quick walk-through/chat, let me know. 
There are some lemmas that I think would belong in VCV-io or CompPoly. For now, I separated them into ToVCV-io/ToCompPoly, but if you agree/think this is fine, I'd make respective PRs to those repos. 
Also related: Is it still policy to put all hardness assumptions into VCV-io, or is there another place? What about very SNARG-specific assumptions like ARSDH?

quangvdao (changes requested):
Requesting changes on one substantive point before merge.

The ARSDH success condition currently does not require the denominator to be nonzero. In `ArkLib/CommitmentScheme/HardnessAssumptions.lean`, the experiment accepts

`S.card = D + 1 ∧ h₁ ≠ 1 ∧ h₂ = h₁ ^ (1 / Zₛ.eval τ).val`

but there is no condition like `Zₛ.eval τ ≠ 0` / `τ ∉ S`. Since division is total in Lean, this makes the zero-denominator case a real formal case rather than an excluded one.

This seems to matter in the new function-binding reduction: the “find τ in queries” branch in `ArkLib/CommitmentScheme/KZG/FunctionBinding.lean` constructs

`S := (Finset.range (n + 1)).image ((↑) : ℕ → ZMod p)`

and returns `g₁ ^ (1 / Zₛ.eval α₁).val` after proving `α₁ = τ`. The proof then closes the local `arsdhCond` using cardinality, `g₁ ≠ 1`, and the equality, but it never proves `Zₛ.eval τ ≠ 0`.

Could you tighten both the global ARSDH experiment and the local `arsdhCond` with the required well-formedness condition, then update this branch to choose/prove a set avoiding `τ`?

Separately, please also remove the accidental `sorry` in `Commitment.Security.hiding'` before merge. I don’t think that declaration is needed for this PR.

Posted by Codex on behalf of Quang Dao; AI-authored with Codex (GPT-5).

quangvdao:
Non-blocking follow-up suggestion: I think the proof would be much easier to maintain if the reduction shape were made a bit more paper-like.

The two biggest wins, in my view:

1. Factor the ARSDH success predicate into a single shared definition. Right now the global `arsdhExperiment` and the local `arsdhCond` restate the same condition. Sharing one predicate would reduce duplication and make it harder for the experiment and reduction-side condition to drift.

2. Split the main bridge lemma by reduction branch. `function_binding_cond_ext_output_maps_to_arsdh` is the real heart of the proof, but it currently contains all three cases inline. I think it would read much better as a dispatcher over named lemmas for:
   - the conflicting-evaluations branch,
   - the `query = τ` branch,
   - the interpolation branch.

I also think named structures for the extended transcript / ARSDH output would help readability. The current nested tuples make the proof harder to scan than the underlying math really is.

None of this needs to block the mathematical fix above; I’m suggesting it because this proof is important enough that future reviewers will probably need to navigate it again.

Posted by Codex on behalf of Quang Dao; AI-authored with Codex (GPT-5).

tobias-rothmann:
@quangvdao Added all your suggested changes. 
- Added the missing non-zero check to ARSDH. 
- refactored the security definitions and hardness assumptions into a more modular form, with reusable abbreviations for Condition, Game, Experiment (the probability of winning a single concrete run of a game), and the actual property/assumption (for all adversaries..., the error is...).
-  Split the main reduction up into your 3 suggested parts (eval break, srs secret in queries, interpolation)
-  introduced the suggested structures for the ARSDH/Extended Game outputs. 

Let me know if there's something else :)

quangvdao (changes requested):
Thanks for the update. The earlier ARSDH denominator issue looks fixed now, and the branch split makes the function-binding proof much easier to follow.

I think there is still one request-changes-level modeling issue: the SDH/ARSDH hardness games appear to sample the SRS trapdoor `τ` through the same cache-backed oracle state that is then passed to the adversary.

Concretely, `tSdhAdversary` / `arsdhAdversary` are arbitrary `StateT unifSpec.QueryCache ProbComp` computations, while `tSdhGame` / `arsdhGame` sample `τ` via `simulateQ randomOracle ($ᵗ (ZMod p))` and then run the adversary in that same state. Since `randomOracle` caches fresh samples with `cacheQuery`, and `QueryCache` is readable state, the adversary can in principle inspect the cache entry for the `τ` sample. That makes the hardness game nonstandard: the adversary is no longer just given the public SRS, it can recover the trapdoor used to generate it.

Could you change the hardness games so setup samples `τ` privately, outside the adversary-visible `QueryCache`, and only the public SRS is passed to the adversary? If the framework needs adversaries to remain in `StateT QueryCache ProbComp`, I think the setup randomness should live in a separate/private computation or state component.

Two smaller but related follow-ups:

1. `functionBindingCondition` currently requires `Function.Injective queryOf`, so the theorem is a distinct-query function-binding theorem. That is narrower than CGKY25 Definition 3.17, and it also makes the duplicate-query conflict branch unreachable from the actual winning event. Either the injectivity condition should be removed, or the theorem/docstring should explicitly say this is distinct-query function binding and relies on a separate theorem for duplicate-query conflicts.

2. KZG setup and ARSDH currently sample `τ : ZMod p`, while CGKY25 samples `τ ∈ Z_p \ {0}`. If this is intentional, it would be good to document that the formal theorem is for an all-field setup distribution; otherwise the sampling should probably be nonzero in both places.

Posted by Cursor/Codex assistant (model: GPT-5.5) on behalf of the user (Quang Dao) with approval.

tobias-rothmann:
Good catch! Updated the PR according to all your suggestions @quangvdao 
- Separated the SRS generation oracle from the oracle that the adversary has access to
- removed injectivity from the FB condition (probably a misrefactor)
- enforced $\tau \neq 0$ for the srs secret (in the hardness assumptions and the KZG srs)

Let me know if you find anything else :)

quangvdao:
Thanks, the earlier correctness issue looks addressed to me now. I think the remaining blockers are maintainability/API-boundary issues rather than a new proof-strategy concern.

Could you please do a narrow cleanup pass with the following changes?

1. Canonicalize the powers-of-τ SRS construction.

Right now `towerOfExponents` / `generateSrs` are defined both in `CommitmentScheme/HardnessAssumptions.lean` and in `CommitmentScheme/KZG/Basic.lean` with the same bodies. Please make this a single shared API, e.g.

```lean
namespace Groups.PowerSrs

def tower ...
def generate ...

end Groups.PowerSrs
```

This should represent the monomial powers-of-τ SRS

```text
(g₁, g₁^τ, ..., g₁^(τ^n)), (g₂, g₂^τ)
```

used by KZG, t-SDH, and ARSDH. Then update `HardnessAssumptions`, `KZG.Basic`, `KZG.Binding`, and `KZG.FunctionBinding` to use this canonical `Groups.PowerSrs.generate`.

I would keep genuinely generic group-power lemmas in `namespace Groups`, SRS-shape lemmas in `namespace Groups.PowerSrs`, and KZG-specific lemmas in `namespace KZG`.

2. Split `KZG/FunctionBinding.lean` into a small number of files.

A four-file split seems enough; I’m not asking for a broad refactor or a new proof strategy:

```text
ArkLib/CommitmentScheme/KZG/Algebra.lean
ArkLib/CommitmentScheme/KZG/FunctionBinding/Support.lean
ArkLib/CommitmentScheme/KZG/FunctionBinding/Branches.lean
ArkLib/CommitmentScheme/KZG/FunctionBinding.lean
```

Suggested boundaries:

- `KZG/Algebra.lean`: shared group/SRS support, including the `Groups.PowerSrs` API and reusable group-order/power lemmas such as `gpow_div_eq`, `gpow_val_mul_eq`, `zmod_eq_zero_of_gpow_eq_one`, etc.
- `KZG/FunctionBinding/Support.lean`: function-binding-specific auxiliaries, such as `findConflict`, `queryReps`, `findA`, `findS`, vanishing/interpolation conversion lemmas, and ARSDH output helper structures/constructors.
- `KZG/FunctionBinding/Branches.lean`: the three main branch proofs: conflicting evaluations, query equals `τ`, and interpolation/no-data.
- `KZG/FunctionBinding.lean`: reduction/game definitions, probability/game-hop lemmas, and the final `function_binding` theorem.

3. Fix the import boundary with `KZG/Binding.lean`.

`KZG/Binding.lean` should not need to import the whole function-binding theorem just to use lower-level group/SRS helper lemmas. After moving those helpers into `KZG/Algebra.lean`, please make `Binding.lean` import that support file instead.

4. Clean up the long-file/style exception.

`FunctionBinding.lean` currently needs a long-file linter override, and the style exception entry appears stale after the latest changes. After the split, please remove or regenerate the relevant style exception rather than leaving old long-file metadata around.

5. Add a short private-setup clarification near the game definitions.

Since setup/keygen randomness is intentionally sampled outside the adversary-visible oracle state now, please add a brief comment/docstring near the generic binding/function-binding games or the KZG extended games clarifying that the setup trapdoor/keygen randomness is private and the adversary only receives the public SRS.

I’m asking for this as a mechanical maintainability cleanup: one canonical powers-of-τ SRS API, cleaner imports, a smaller final theorem file, and accurate style metadata. The current proof strategy can remain the same.

Posted by Codex assistant (model: GPT-5.5) on behalf of the user (Quang Dao) with approval.

tobias-rothmann:
Updated the files (almost) according to your changes @quangvdao 
- Canonicalized the powers-of-τ SRS construction into one Algebra file 
- Split KZG/FunctionBinding.lean into a small number of files (and accordingly the import in Binding and long file exception)
- Add a short private-setup clarification near the game definition

The only thing I didn't follow closely was the file structure when splitting the function-binding file. I don't think it makes particularly much sense to divide into a dedicated support and branches file (what is the objective here? Separating (mostly) definitions from proofs? Outsourcing only a few lemmas?). Instead, I split per branch into files (so one file for the evaluation binding conflict case, one for the $\tau$ is in queries case, and one for the degree failure case, with a support file for reused helpers and a file that pulls the whole proof together/does the game lifting part).

Let me know if this is okay or if you have anything else :)

quangvdao (dismissed):
Thanks Tobias, this addresses my requested changes. The `Groups.PowerSrs` canonicalization, the function-binding split, the `KZG.Binding` import cleanup, the long-file/style-exception cleanup, and the private-setup clarification all look good to me. Approving from my side.

Only two non-blocking nits, in case you want to do one final cleanup commit:

- `KZG.Basic` now imports `KZG.HardnessAssumptions` mostly for `Groups.sampleNonzeroZMod`. If you want a cleaner dependency boundary later, that sampler could live closer to `KZG.Algebra` / `Groups.PowerSrs` or in a tiny shared sampling-support file.
- The private-setup note is duplicated for t-SDH and ARSDH. That is harmless, but it could be consolidated if you want to keep the file a bit tighter.

No further changes requested from me. Thanks for doing all the requested cleanup.

Posted by Codex assistant (model: GPT-5.5) on behalf of the user (Quang Dao) with approval.

tobias-rothmann:
Added the one final cleanup commit @quangvdao :)

quangvdao (approved):
Posted by the assistant on behalf of Quang Dao. AI-authored by Codex (GPT-5).

Re-approving after the latest cleanup commit. The new commit is a narrow extraction of `Groups.sampleNonzeroZMod` and its `simulateQ` lemma into `KZG.Sampling`, with corresponding import cleanup and private-setup-note deduplication. That addresses the dependency-boundary nit from my prior approval without changing the substance of the KZG correctness/function-binding development. CI is green.

PR #279, opened 2025-12-17
Refactoring Coding Theory

alexanderlhicks:
Opening up this PR/branch to refactor and clean up some of the contents of our coding theory formalization.

Initial commits:

- Moved contents of STIR and WHIR into CodingTheory, fixed what seemed to be some issues in STIR along the way.
- Refactored ProximityGaps slightly, renaming files in terms of their contents rather than the reference paper authors, and moving the contents of AHIV22/Ligero into the other two (now ReedSolomon and Interleaved) for greater clarity if new results from more recent papers are formalized.


alexanderlhicks:
> For Interleaved.lean, most (if not all) proximity gap definitions/theorems can be written from the interleaved code POV. Similarly for ReedSolomon.lean, a lot of papers state proximity gaps for RS codes. 1 theorem might be put in both Interleaved.lean & ReedSolomon.lean if we don't have clear distinction in terms of purpose of the files
Yep, beyond what's been done on this PR now (which is really just moving stuff across, particularly the STIR/WHIR contents), I think a next step (in this PR or a follow up) will be to see where we can clean up and unify results (STIR and WHIR alone have/had overlap given parts of the WHIR paper are literally copy pasted from the STIR paper), but this first requires setting up the file structure so that we know what boxes we to put things in.

> If we don't split helper lemmas, the files would look pretty long & distracted (e.g. the DG25 file is 2k lines and it is only about proximity gaps for multilinear combination, but its main theorems are short). But idk how should we do it atm. What do you think?

Yep, I agree about the bloat. One option is to have small files for key results and split auxilliary/helpful lemmas across other files according to what they cover. See for example the edits I made to the proof of Polishchuk Spielman here: https://github.com/alexanderlhicks/ArkLib/pull/1/files, which keeps the proof of the main result in PolishchukSpielman.lean and splits intermnediary results across three files based on what they cover (if what they cover is not easily categorized, they can just be split into Aux.lean  or Aux1.lean, Aux2.lean if they are too long...).
This makes it easy to browse important theorems whilst ignoring intermeidary results, and speeds up build times if working on any individual file.

alexanderlhicks:
closing as stale.

PR #311, opened 2026-02-21
feat(guruswami-sudan): constructive decoder witness/filter correctness

### LINKED ISSUE #213: Proof obligation for `decoder` in `ArkLib/Data/CodingTheory/GuruswamiSudan/GuruswamiSudan.lean`
A proof in `ArkLib/Data/CodingTheory/GuruswamiSudan/GuruswamiSudan.lean` contains a `sorry`.

**🤖 AI Analysis:**
### Statement Explanation
This is an `opaque` definition for the Guruswami-Sudan list decoder. It defines a function that takes the parameters of a Reed-Solomon code and a received word, and returns a list of candidate decoded polynomials.

-   **Goal**: The function `decoder` is intended to implement the Guruswami-Sudan list-decoding algorithm. Given a received word `f`, it should output a list of all polynomials `p` of degree less than `k` such that the Hamming distance between the received word `f` and the codeword generated by `p` (`p.eval ∘ ωs`) is within a certain error threshold `e`.
-   **Hypotheses (Parameters)**:
    -   `k`: The message length, corresponding to the maximum degree of the message polynomial (`deg(p) < k`).
    -   `r`: The multiplicity parameter for the interpolation step.
    -   `D`: The weighted degree bound for the auxiliary bivariate polynomial `Q`.
    -   `e`: The maximum number of errors the decoder is expected to handle. The distance between the received word and a valid decoded codeword should be at most `e`.
    -   `ωs`: The `n` distinct evaluation points defining the Reed-Solomon code.
    -   `f`: The received word, which is a sequence of `n` field elements.

The return type is `List F[X]`, signifying that this is a list-decoder: it can return multiple possible candidates instead of a single unique one.

### Context
This `decoder` is the central piece of the Guruswami-Sudan algorithm implementation in this file. Its behavior is formally specified by the two theorems that follow it:
-   `decoder_mem_impl_dist`: This is a correctness property. It asserts that any polynomial `p` returned by the decoder is indeed "close" to the received word `f` (the Hamming distance is at most `e`).
-   `decoder_dist_impl_mem`: This is a completeness property. It asserts that any polynomial `p` that is "close" to `f` will be found by the decoder.

The algorithm itself consists of two main stages:
1.  **Interpolation**: Find a non-zero bivariate polynomial `Q(X, Y)` that satisfies the `GuruswamiSudan.Condition` structure defined earlier. This means `Q` must have a low weighted degree and pass through each point `(ωs i, f i)` with at least multiplicity `r`. The lemma `guruswami_sudan_for_proximity_gap_existence` establishes that such a `Q` exists under certain parameter choices.
2.  **Root-Finding**: Find all factors of `Q(X, Y)` that are of the form `Y - p(X)`, where `p(X)` has a degree less than `k`. The lemma `guruswami_sudan_for_proximity_gap_property` provides the theoretical guarantee that if a message polynomial `p` corresponds to a nearby codeword, then `Y - p(X)` must be a factor of `Q`.

### Proof Suggestion
Since this is an `opaque` definition, the goal is not to prove a theorem but to provide an implementation. A high-level plan to implement the Guruswami-Sudan algorithm would be:

1.  **Interpolation Step (Finding `Q`)**:
    -   The conditions in `GuruswamiSudan.Condition` on `Q` translate to a system of homogeneous linear equations. The variables of this system are the unknown coefficients of `Q(X, Y)`.
    -   The constraints `Bivariate.rootMultiplicity Q (ωs i) (f i) ≥ r` for each `i` mean that `Q` and its partial derivatives up to a certain order must vanish at `(ωs i, f i)`. Each of these evaluations yields a linear equation.
    -   You need to set up this system of equations. The number of variables depends on the weighted degree bound `D`, and the number of equations depends on `n` and `r`.
    -   Use a linear algebra solver (e.g., Gaussian elimination) to find a non-trivial solution for the coefficients of `Q`. The existence of such a `Q` is guaranteed if the number of variables is greater than the number of equations.

2.  **Root-Finding Step (Factoring `Q`)**:
    -   Once you have the polynomial `Q(X, Y)`, treat it as a polynomial in `Y` with coefficients in the ring `F[X]`.
    -   Use a polynomial factorization algorithm to find its roots in `F[X]`. Specifically, you are looking for factors of the form `Y - p(X)`.
    -   Collect all polynomials `p(X)` from these factors such that `p.natDegree < k`.
    -   The list of these polynomials `p(X)` is the final output of the decoder.

**Goal:** Replace the `sorry` with a complete proof.

[Link to the sorry on GitHub](https://github.com/Verified-zkEVM/ArkLib/blob/master/ArkLib/Data/CodingTheory/GuruswamiSudan/GuruswamiSudan.lean#L57)

**Code Snippet:**
```lean
opaque decoder (k r D e : ℕ) (ωs : Fin n ↪ F) (f : Fin n → F) : List F[X] := sorry

```

eliasjudin:
This PR adds proofs autoformalised by @Aristotle-Harmonic.

Implements constructive Guruswami-Sudan witness extraction in `computeGsWitness` via a normalized linear system solved by `BerlekampWelch.linsolve`, removing the brute-force fallback from the decoder path. It checks multiplicity by Hasse-derivative vanishing and performs computable CompPoly root filtering (`Q(X, p(X)) = 0`), with decoder soundness captured by `decoder_mem_impl_dist`; closes #213.

Co-authored-by: Aristotle (Harmonic) <aristotle-harmonic@harmonic.fun>

### DIFF
--- a/ArkLib/Data/CodingTheory/GuruswamiSudan/GuruswamiSudan.lean
+++ b/ArkLib/Data/CodingTheory/GuruswamiSudan/GuruswamiSudan.lean
@@ -1,37 +1,53 @@
 /-
 Copyright (c) 2024-2025 ArkLib Contributors. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: František Silváši, Ilia Vlasov, Stefano Rocca
+Authors: František Silváši, Ilia Vlasov, Stefano Rocca, Elias Judin
 -/
+
 import Mathlib.Algebra.Field.Basic
 import Mathlib.Algebra.Polynomial.Basic
 import Mathlib.Data.Real.Sqrt
+import Mathlib.RingTheory.Polynomial.Basic
 
+import ArkLib.Data.CodingTheory.BerlekampWelch.Sorries
 import ArkLib.Data.CodingTheory.GuruswamiSudan.Basic
-/-! # Guruswami-Sudan Decoder -/
+import ArkLib.Data.CodingTheory.ReedSolomon
+import ArkLib.Data.Polynomial.Bivariate
+import ArkLib.Data.Polynomial.Interface
 
+import CompPoly.Univariate.Lagrange
 
+/-!
+# Guruswami-Sudan Decoder
 
-open Finset Finsupp Polynomial Polynomial.Bivariate ReedSolomon
+This module keeps the abstract Guruswami-Sudan specification decoder alongside
+constructive candidate generation for Reed-Solomon codes.
 
---Let `F` be a field (finite).
-variable {F : Type} [Field F] [DecidableEq F]
---Let `k + 1` be the **dimension** of the code.
-variable {k : ℕ}
---Let `n` be the **blocklength** of the code.
-variable {n : ℕ}
---Let `m` be a natural number, serving as the **multiplicity parameter**.
-variable {m : ℕ}
---Let `ωs` be the **domain of evaluation**, i.e. the interpolation points.
-variable {ωs : Fin n ↪ F}
---Let `f` be the **received word**, possibly corrupted.
-variable {f : Fin n → F}
+The witness search is implemented by `computeGsWitness`, which solves a linearized
+system of Hasse-derivative constraints with a normalization equation. Candidate
+message polynomials are then filtered by a computable root check for
+`$Q(X, p(X)) = 0$` using CompPoly arithmetic.
+
+## References
+
+* [Bafna, P., Chiesa, A., Ishai, Y., Khurana, D., and Spooner, N.,
+    *On the Proximity Gap of Reed-Solomon Codes*][BCIKS20]
+-/
 
 namespace GuruswamiSudan
 
+variable {F : Type} [Field F] [DecidableEq F]
+variable {k : ℕ}
+variable {n : ℕ}
+variable {m : ℕ}
+variable {ωs : Fin n ↪ F}
+variable {f : Fin n → F}
+
+open Finset Finsupp Polynomial Polynomial.Bivariate ReedSolomon
+
 variable (k m) in
 /--
-Guruswami–Sudan conditions for the polynomial searched by the decoder.
+Guruswami–Sudan conditions for the polynomial searched by the specification decoder.
 
 These conditions characterize a nonzero bivariate polynomial `Q(X,Y)`
 with bounded weighted degree that vanishes with sufficiently high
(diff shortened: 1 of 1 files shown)


ArkLib/Data/CodingTheory/GuruswamiSudan/GuruswamiSudan.lean
```diff
@@ -53,8 +53,20 @@ structure Condition
...
-/-- Guruswami-Sudan decoder. -/
-opaque decoder (k r D e : ℕ) (ωs : Fin n ↪ F) (f : Fin n → F) : List F[X] := sorry
+/--
+Guruswami-Sudan decoder.
+
+If there exists a witness polynomial `Q` satisfying `Condition k r D ωs f Q`, the decoder
+returns the roots of `Q` filtered to degree `< k` and distance `≤ e` from `f`.
+If no such witness exists, it returns `[]`.
+-/
+noncomputable opaque decoder (k r D e : ℕ) (ωs : Fin n ↪ F) (f : Fin n → F) : List F[X] :=
```
quangvdao: Is there a way to make this definition computable? and can you also try to prove something with it, to make sure it's the right definition?

eliasjudin: `decoder` uses `Classical.choose` so it stays `noncomputable`; full computability needs a constructive GS procedure.

Removed `opaque` so the body is visible. Proved `decoder_mem_impl_dist`: if `p ∈ decoder k r D e ωs f` then `Δ₀(f, p.eval ∘ ωs) ≤ e`.

Proof assisted by @Aristotle-Harmonic

eliasjudin: I'm working on a constructive GS.

eliasjudin: Added `[Fintype F]` and replaced `Classical.choose`-based decoder with an algorithmic GS pipeline: Gaussian elimination for the interpolation system, `Finset.univ.filter IsRoot` for root-finding, and evaluation-interpolation for bivariate polynomial roots.

Definitions remain `noncomputable` because Mathlib's `Polynomial` and `Finset` types have no executable code (`Polynomial.C`, `Polynomial.X`, `Finset.toList` all use `Classical`). True computability would require a different polynomial representation.

`decoder_mem_impl_dist` is proved: if `p ∈ decoder k r D e ωs f` then `Δ₀(f, p.eval ∘ ωs) ≤ e`.

ArkLib/Data/CodingTheory/GuruswamiSudan/GuruswamiSudan.lean
```diff
@@ -53,8 +53,20 @@ structure Condition
...
-/-- Guruswami-Sudan decoder. -/
-opaque decoder (k r D e : ℕ) (ωs : Fin n ↪ F) (f : Fin n → F) : List F[X] := sorry
+/--
+Guruswami-Sudan decoder.
+
+If there exists a witness polynomial `Q` satisfying `Condition k r D ωs f Q`, the decoder
+returns the roots of `Q` filtered to degree `< k` and distance `≤ e` from `f`.
+If no such witness exists, it returns `[]`.
+-/
+noncomputable def decoder (k r D e : ℕ) (ωs : Fin n ↪ F) (f : Fin n → F) : List F[X] :=
```
quangvdao: Can you add a docstring here noting its noncomputability & outline a way to get computability in the future?

eliasjudin: sure will do, I tried a couple methods and they all ended up relying on `Classical.choose` so would require new polynomial methods not in mathlib, I'll do some more research into this

eliasjudin:
> Thanks for the PR! I'd like to merge #300 and then this one on top of it if you could keep an eye out for that and facilitate this.

@alexanderlhicks I'll wait for that and then update! Leaving the proximity generators stuff to @katyhr

alexanderlhicks:
> > Thanks for the PR! I'd like to merge #300 and then this one on top of it if you could keep an eye out for that and facilitate this.
> 
> @alexanderlhicks I'll wait for that and then update! Leaving the proximity generators stuff to @katyhr

Checking in on this too sinc eit's marked as draft but I see you made some commits after the comment.

alexanderlhicks:
Closing in favour of #436 which builds on some of your work in this PR so your (co-)authorship is preserved in #436's history.

PR #313, opened 2026-02-22
chore: update to v4.28.0

dtumad:
This PR updates Mathlib, CompPoly, and VCV to `v.4.28.0` and fixes downstream proofs that break.

The only remaining regression is in `QueryPhase.lean` where the specification `Prop` now compiles very slowly so I've enabled `debug.skipKernelTC` there for now.


dtumad:
I think this should be set to merge as soon as someone gets a chance to review and approve it.

ArkLib/ProofSystem/Binius/BinaryBasefold/QueryPhase.lean
```diff
@@ -58,6 +58,8 @@ def extractNextSuffixFromChallenge (v : sDomain 𝔽q β h_ℓ_add_R_rate ⟨0,
   simp only [Fin.val_zero, zero_add] at val
   exact val
 
+-- NOTE(dtumad): Compilation time without this is painfully slow right now
+set_option debug.skipKernelTC true in
```
dhsorens: does this do what I think it does-  skip the type checker? is this a soundness problem for `proximityChecksSpec`?

ArkLib/Data/MvPolynomial/Degrees.lean
```diff
@@ -39,10 +39,6 @@ def CEmbedding : R ↪ MvPolynomial σ R := ⟨C, C_injective σ R⟩
 
 section Support
 
-theorem support_C {r : R} [h : Decidable (r = 0)] :
-    (@C R σ _ r).support = if r = 0 then ∅ else { 0 } := by
-  rw [←monomial_zero', support_monomial]
-
```
dhsorens, line 45: just making a note that this was removed bc it is now in mathlib

ArkLib/Data/CodingTheory/ProximityGap/DG25.lean
```diff
@@ -413,10 +414,6 @@ variable {F : Type} [CommRing F] [Fintype F] [NoZeroDivisors F] [DecidableEq F]
 variable (MC : ModuleCode ι F A) [Nontrivial MC]
   (C : Set (Word A ι)) [Nonempty C] -- todo: change to Nontrivial if needed
 
-instance : NoZeroSMulDivisors (R := F) (M := A) := Module.Free.noZeroSMulDivisors F A
-
-instance : NoZeroSMulDivisors (R := F) (M := Word A ι) := _root_.Function.noZeroSMulDivisors
-
```
dhsorens, line 419: what's the story here? I noticed that reintroducing them gave me errors, looks like some of the definitions in mathlib changed. are there alternative definitions? if we don't need them, I'm happy to have them removed too

dhsorens (commented):
This looks mostly fine to me, everything compiles locally for me fine. Most of these changes are fixing proofs, there are a couple introduced TODOs (in comments), no introduced sorry's afaict, and a couple deletions. The only real concern I have that may affect soundness is the introduction of the line `set_option debug.skipKernelTC true in` in the definition of `proximityChecksSpec`.

Reviewed for:
- correctness
- soundness 
- added/resolved TODOs and `sorry`s
- proof repair

ArkLib/ProofSystem/Binius/BinaryBasefold/QueryPhase.lean
alexanderlhicks: yep
(I assume in this case it does actually go through without this option and Devon added it for the sake of compiling faster?)

chung-thai-nguyen: I will make many big updates on Binius later and I will fix type checking/slow compilation problems on my side when I do rebase, one of the reason might be due to the strict index design. So you can go ahead with some temporary solutions that you have to get the PR merged if that's acceptable, like #exit or so.

ArkLib/Data/CodingTheory/ProximityGap/DG25.lean
dtumad: It is because of [this change to mathlib](https://github.com/leanprover-community/mathlib4/commit/290f644e69925cec5f98aea3e950cb5218583b54) which seems to let `Module.Free` produce the required instances on its own now I think.

ArkLib/ProofSystem/Binius/BinaryBasefold/QueryPhase.lean
dtumad: Well not exactly not type-checked, the interpreter still checks it and you can step through it in VSCode for example. The only reason this is ever useful is that there is an error in how the interpreter models the kernel, which is why this error seemed very strange to me.

I've pushed a fix now, I have no idea why it works, it should be exactly equivalent but for some reason the kernel accepts this version. Some combination of not de-structuring the sub-domain element explicitly, using the change statement, and not using omega for the bounds was the problem. Seems like potentially something to do with having subtype elements as `Fin` arguments but I can't get any kind of MWE of the problem.

quangvdao (approved):
Woops! There ya go

lalalune:
## Update: FinalSumcheck fixed (verified) — but the Binius cone has extensive in-flight #317 API-drift breakage beyond it

**FinalSumcheck.lean: FIXED and verified.** Commit `633a0526b` — the 4 HEq errors (279/290/364/743) are gone; `lake env lean ArkLib/ProofSystem/Binius/BinaryBasefold/Steps/FinalSumcheck.lean` compiles **0 errors, 0 sorry**, via a verified `getFirstOracle_heq` bridge lemma (`HEq (getFirstOracle 𝔽q β oStmt) (oStmt ⟨0,h⟩)`) reused at all 4 sites.

**But the focused cone build still fails** — a real `lake build` of `BinaryBasefold.General`/`FRIBinius.General`/`BBFSmallFieldIOPCS` now surfaces **~1000 errors across 40+ files**, all from the **active #317 refactor mid-flight** (commits explicitly tagged "[UNVERIFIED pending cone build]" / "WIP, pre-compile"). Root causes (not random — signature/API drift not yet propagated to callers):

| count | error class | example |
|---|---|---|
| 13 | renamed function arg | `Invalid argument name ϑ for queryPhaseLogicStep` |
| 12 | dependent-type drift | `Invalid field notation` / `Application type mismatch` |
| 5 | renamed arg | `Invalid argument name y for iterated_fold` |
| 3 | dead identifier | `Unknown identifier k_succ_mul_ϑ_le_ℓ` |

Worst-hit files: `Prelude.lean` (147), `QueryPhase.lean` (139), `Soundness/Incremental.lean` (84), `ReductionLogic.lean` (81), `Relations.lean` (54), `Soundness/SuffixAlignCore.lean` (52), `ExtractMLPCorrectness.lean` (44), `Steps/Fold.lean` (42), `Basic.lean` (35)…

**Honest disposition of closeout item 1:** the cone build is **not** a stale-cache artifact (my first audit was wrong) **nor** a single break (my FinalSumcheck-only framing was incomplete) — it is the **in-progress #317 cone refactor**, currently broken across the whole tree as the renamed-signature changes propagate. Fixing it is the active #317 effort (multiple agents committing pre-compile WIP), not closeable by isolated patches while the refactor is mid-flight. My FinalSumcheck fix is one verified file repaired; the remaining ~40 files are the live #317 propagation. Recommend #313 item 1 be tracked as "blocked on #317 cone refactor completion + a green `lake build` of the three front-door modules," with a CI gate added once #317 lands so the cone can't silently re-break.

PR #363, opened 2026-02-24
List Decodability Correction

katyhr:
Edits of some coding theory inconsistencies/repeats.

- `ListDecodability.lean` : Edit the definition of Hamming ball and Hamming ball with respect to the relative Hamming distance (previously those were a list of codewords close to a given vector). This are the definition of closeCodewords and closeCodewordsRel now. Change these throughout the repo
- Remove lemmas that no longer make sense


alexanderlhicks (approved):

alexanderlhicks:
Thank you!!

PR #383, opened 2026-03-03
feat: completeness and rbrKnowledgeSoundness of FRI-Binius protocols

chung-thai-nguyen:
[x] Completeness/Soundness unfolding tools & snippets - mainly tools for converting the monadic defs into the logical defs (Completeness.lean, ReductionLogic.lean, Simulation.lean, Lemmas.lean) + new cast definition of oracle reduction (OracleReduction/Cast.lean) with completeness/rbrks compatibility
[x] Completeness & rbrKnowledgeSoundness for all FRI-Binius protocols: Binary Basefold, Ring-switching, FRI-Binius, simple ring-switching construction (BBFSmallFieldIOPCS), completeness of BBFSmallFieldIOPCS
[x] Minor changes: reintroduce AdditiveNTT.lean with index changes, will be migrated to CompPoly later

Status: All PR-owned files are sorry-free; the top-level composed results inherit sorryAx from OracleReduction's composition layer

Built with the help of Codex, Claude, Cursor, Gemini.
---
- Archived PRs: https://github.com/Verified-zkEVM/ArkLib/pull/270 https://github.com/Verified-zkEVM/ArkLib/pull/296 https://github.com/Verified-zkEVM/ArkLib/pull/417


alexanderlhicks:
/review

External:
https://eprint.iacr.org/2024/504.pdf
https://eprint.iacr.org/2023/1784.pdf

Internal:
ArkLib/Data/CodingTheory/
ArkLib/Data/Polynomial

Comments:
Please review for correctness with respect to ArkLib internals and the external reference papers.

ArkLib/ProofSystem/Binius/BinaryBasefold/Steps/FinalSumcheck.lean
```diff
@@ -0,0 +1,1796 @@
...
+  UDRCodeword 𝔽q β (h_ℓ_add_R_rate := h_ℓ_add_R_rate)
+    (i := finalOracleNextIdxOrig (ℓ := ℓ) (ϑ := ϑ)
+      (h_ℓ_add_R_rate := h_ℓ_add_R_rate) t ht)
+    (h_i := by
+      exact oracle_index_add_steps_le_ℓ ℓ ϑ (i := Fin.last ℓ) (j := ⟨t, Nat.lt_of_succ_lt ht⟩))
+    (f := finalOracleNextRaw (𝔽q := 𝔽q) (β := β) (ℓ := ℓ) (ϑ := ϑ)
+      (h_ℓ_add_R_rate := h_ℓ_add_R_rate) oStmtOut t ht)
+    (h_within_radius := h_close)
+
+private axiom finalOracleDecoded_next_heq
```
alexanderlhicks: could we avoid the reliance on axioms?

ArkLib/ProofSystem/Binius/BinaryBasefold/Steps/FinalSumcheck.lean
```diff
@@ -0,0 +1,1796 @@
...
+      (h_ℓ_add_R_rate := h_ℓ_add_R_rate) oStmtOut t ht)
+    (h_close_next_final : finalOracleClose (𝔽q := 𝔽q) (β := β) (ℓ := ℓ) (ϑ := ϑ)
+      (h_ℓ_add_R_rate := h_ℓ_add_R_rate) oStmtOut (t + 1) ht) :
+    HEq
+      (finalOracleNextCodeword (𝔽q := 𝔽q) (β := β) (ℓ := ℓ) (ϑ := ϑ)
+        (h_ℓ_add_R_rate := h_ℓ_add_R_rate) oStmtOut t ht h_close_next)
+      (finalOracleDecodedAt (𝔽q := 𝔽q) (β := β) (ℓ := ℓ) (ϑ := ϑ)
+        (h_ℓ_add_R_rate := h_ℓ_add_R_rate) oStmtOut (t + 1) ht h_close_next_final)
+
+private axiom finalDecodedPrefixAt_succ_heq
```
alexanderlhicks: could we avoid the reliance on axioms?

ArkLib/ProofSystem/Binius/BinaryBasefold/Prelude.lean
```diff
@@ -678,361 +1117,1340 @@ theorem qMap_total_fiber_disjoint
...
+  rfl
+
+/-- Evaluates polynomial P on the domain S⁽ʲ⁾.
+    This function is index-agnostic: logic doesn't change based on the round. -/
+def polyToOracleFunc {domainIdx : Fin r} (P : L[X]) :
+    (sDomain 𝔽q β h_ℓ_add_R_rate domainIdx) → L :=
+  fun y => P.eval y.val
+
+omit [CharP L 2] in
+/-- **Lemma 4.13** : if f⁽ⁱ⁾ is evaluation of P⁽ⁱ⁾(X) over S⁽ⁱ⁾, then fold(f⁽ⁱ⁾, r_chal)
```
alexanderlhicks: It would be good to refer to a specific version of the paper here (in the file header) as this is lemma 4.14 in https://eprint.iacr.org/2024/504.pdf

ArkLib/ProofSystem/Binius/FRIBinius/General.lean
alexanderlhicks, line 19: duplicate ## References

ArkLib/ToVCVio/Simulation.lean
```diff
@@ -0,0 +1,2455 @@
...
+by rfl
+
+end TranscriptLemmas
+
+section SupportPreservation
+
+variable {ι : Type} {spec : OracleSpec ι} [spec.Fintype] {α β : Type}
+  {m : Type → Type} -- [AlternativeMonad m] [LawfulAlternative m]
+
+#check support_liftM
```
alexanderlhicks: can remove this

ArkLib/ToVCVio/Simulation.lean
```diff
@@ -0,0 +1,2455 @@
...
+@[simp]
+lemma OptionT.simulateQ_failure
+    {ι ι' : Type*} {spec : OracleSpec ι} {spec' : OracleSpec ι'}
+    (so : QueryImpl spec (OracleComp spec')) :
+    simulateQ so (failure : OptionT (OracleComp spec) α) =
+      (failure : OptionT (OracleComp spec') α) := by
+  rfl
+
+@[simp]
+lemma OptionT.simulateQ_failure' {α : Type u}
```
alexanderlhicks: This appears to be a duplicate of lemma ptionT.simulateQ_failure above

ArkLib/ToVCVio/Lemmas.lean
```diff
@@ -8,4 +8,889 @@ universe u v w
...
+
+/-- Simplify a bind following a pure initialization of state. -/
+@[simp]
+theorem run'_bind_pure [Monad m] [LawfulMonad m] {s : σ} (x : α) (f : α → StateT σ m β) :
+    (StateT.run' (pure x >>= f) s) = (StateT.run' (f x) s) :=
+by simp [StateT.run']
+
+alias run'_pure_bind := run'_bind_pure
+
+-- @[simp]
```
alexanderlhicks: Can remove all these commented out lemmas if not useful

ArkLib/ProofSystem/Binius/FRIBinius/General.lean
```diff
@@ -171,52 +183,166 @@ noncomputable def fullOracleProof :
...
+              (aOStmtIn :=  (BinaryBasefoldAbstractOStmtIn (β := β)
+        (ϑ := ϑ) (h_ℓ_add_R_rate := h_ℓ_add_R_rate))))
+            (h₂ := coreInteractionOracleVerifier_rbrKnowledgeSoundness
+              (κ := κ) (L := L) (K := K) (β := β) (ℓ := ℓ) (ℓ' := ℓ') (h_l := h_l)
+              (𝓡 := 𝓡) (ϑ := ϑ) (h_ℓ_add_R_rate := h_ℓ_add_R_rate) (𝓑 := 𝓑)))
+      (h₂ := QueryPhase.queryOracleVerifier_rbrKnowledgeSoundness K β γ_repetitions
+        (h_ℓ_add_R_rate := h_ℓ_add_R_rate) (ϑ := ϑ) init impl)
+  dsimp only [ChallengeIdx, fullOracleVerifier, batchingCorePspec,
+    BinaryBasefold.pSpecCoreInteraction, batchingCoreVerifier, MessageIdx] at res ⊢
+  exact res
```
alexanderlhicks, line 336: It may be interesting to cover the concrete soundness (p46 of https://eprint.iacr.org/2024/504.pdf), I think everything that's needed for this is already in place?

chung-thai-nguyen: @alexanderlhicks Thanks for the thorough reviews! I've just addressed all of your comments and added proofs of knowledge soundness for all the Binius protocols (e.g., `fullOracleVerifier_knowledgeSoundness` in `FRIBinius/General.lean`). I'll migrate this PR to the new OracleReduction framework once it rolls out, so I'm leaving it open for now.

alexanderlhicks:
Just checking since you marked this as draft, what changes are pending?
(edit: I see https://github.com/Verified-zkEVM/ArkLib/pull/455)

chung-thai-nguyen:
> Just checking since you marked this as draft, what changes are pending? (edit: I see #455)

@alexanderlhicks I was waiting for https://github.com/Verified-zkEVM/ArkLib/pull/433 to be completed so I could migrate to it, but since it may take some time and adopting the new framework could introduce significant changes, I'm thinking about wrapping up this PR as a checkpoint for the security proofs and handling the migrations for computable Binius and the Interaction framework gradually in the later PRs.

alexanderlhicks:
Thanks! Will quickly test the update review workflow on this to make sure it handles larger PRs (else I'll tweak token limit parameters).

edit: token limits in the workflow setting got in the way, removed them.

ArkLib/ProofSystem/Binius/BinaryBasefold/General.lean
```diff
@@ -149,18 +155,19 @@ variable {σ : Type} {init : ProbComp σ} {impl : QueryImpl []ₒ (StateT σ Pro
 
 /-- Round-by-round knowledge soundness for the full Binary Basefold oracle verifier -/
 theorem fullOracleVerifier_rbrKnowledgeSoundness :
```
alexanderlhicks, line 157: > `theorem fullOracleVerifier_rbrKnowledgeSoundness :`

Consolidating everything axiom-related into one comment.

`#print axioms` on this theorem gives `[propext, sorryAx, Classical.choice, Quot.sound]`, and the
same holds for all twelve top-level results across `BinaryBasefold/General`, `FRIBinius/General`
and `BBFSmallFieldIOPCS`. So the PR body's "Status: **Fully proved** (no sorrys remaining)" holds
for your files but not for these results.

A full census (`Lean.collectAxioms` over every non-internal constant in the 37 PR-owned modules)
puts it at **2,013 axiom-clean / 89 tainted**, with **no `sorryAx` originating in this PR**. All
nine origins are pre-existing stubs in `main`:

*Sorry-defined objects* — `OracleVerifier.append` (`Composition/Sequential/Append.lean:148,158`,
`verify` field), reached because `fullOracleVerifier` at line 47 *is* an append;
`OracleVerifier.liftContext` (`LiftContext/OracleReduction.lean:53,57,58` — all three fields),
reached through the ctx-lens instances; `OracleVerifier.cast` (`Cast.lean:128`, `verify` field).

*Sorry-proved theorems* — `OracleVerifier.append_toVerifier` (`:191`),
`Verifier.append_rbrKnowledgeSoundness` (`:506`), `OracleReduction.append_{completeness,
perfectCompleteness}` (`:473`, `:486`), `Reduction.append_completeness` (`:439`),
`Verifier.rbrKnowledgeSoundness_implies_knowledgeSoundness` (`Security/Implications.lean:102`).

**A second, independent origin for the scalar-KS theorems.**
`fullOracleVerifier_knowledgeSoundness` (`:309`, and its `FRIBinius`/`BBFSmallFieldIOPCS`
counterparts) additionally routes through `rbrKnowledgeSoundness_implies_knowledgeSoundness`. So
even if the whole append layer were filled, those three would still be hollow — and they are the
ones a reader is most likely to cite, because they carry the concrete DP24 bound. Worth considering
whether to state them at all until `Implications.lean` is filled, versus stopping at
`rbrKnowledgeSoundness` plus `fullRbrKnowledgeError_sum_le_concrete`, which is the real content and
is untainted by `Implications.lean`.

**Your own `Cast.lean` additions are clean**, and that should be on record: `castInOut`,
`castOutSimple` and their four completeness plus two rbrKS compatibility theorems (lines 225–570)
are all axiom-clean. The 8 tainted declarations in that file are the pre-existing `cast` and the
theorems stated directly about it.

Worth knowing for planning: **the ABF26 framework work does not close these.** PR #668 removes all
three `Security/RoundByRound.lean` sorries and adds a sorry-free `Security/RbrGame.lean`, but
neither it nor #505 touches `Append.lean`, `Implications.lean`, `LiftContext/` or `Cast.lean`.

**Actionable:**
1. Adjust the PR body — e.g. "all PR-owned files are sorry-free; the top-level composed results
   inherit `sorryAx` from `main`'s composition layer".
2. Add a one-line disclosure to the docstring of each of the twelve top-level theorems naming the
   inherited stub.
3. Nothing for you on the composition layer itself — **I have filed
   https://github.com/Verified-zkEVM/ArkLib/issues/676** to track it. 89 declarations in this PR alone are waiting on it,
   which makes it the highest-leverage gap in the library right now, and it is substrate work
   rather than something a protocol PR should carry.

This is not a one-off. PR #503 (LogUp) landed in exactly the same position: `LogUp/**` sorry-free,
but `#print axioms logup_{soundness,completeness}` returning `sorryAx` inherited from the same
family of framework stubs — `append_*`, the `Security/Implications.lean` implications, and
`liftContext_*` — none of which that PR touched either. Two independent protocol formalizations
hitting the identical wall is the argument for prioritising the composition layer over any
individual protocol PR, and for making the `#print axioms` disclosure a standing convention rather
than something asked for case by case.

Everything below the append layer is clean; none of this touches the mathematics.[^1]

[^1]: Flagging one asymmetry only because it affects what the docstrings can say: the three
sorry-defined objects are not in the same position as the sorry-proved lemmas. For the latter,
"proof pending" is accurate. For the former the statement has no subject yet, so a disclosure
should say the result is not yet a claim about a constructed verifier rather than that its proof
is outstanding.

ArkLib/Data/FieldTheory/AdditiveNTT/AdditiveNTT.lean
```diff
@@ -0,0 +1,3090 @@
+/-
+Copyright (c) 2024-2025 ArkLib Contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Chung Thai Nguyen, Quang Dao
+-/
+
```
alexanderlhicks, line 6: > `import CompPoly.Fields.Binary.AdditiveNTT.NovelPolynomialBasis`

This file reinstates a module #307 ("Refactoring CompPoly Out of Arklib") deleted, and re-derives
62 of its 88 declarations from CompPoly modules it does not import — in the **same
`namespace AdditiveNTT`**, i.e. the same qualified names with different signatures. They coexist
only because the two are never imported together; the first consumer needing both hits an
ambiguity wall.

Having gone through it declaration by declaration, though, this is not carelessness and the "with
index changes" in your PR description is not a style preference. Two things came out of the dig:
**it works around a real limitation in CompPoly**, and **the overlap is the residue of parallel
development**. CompPoly gained `Domain`, `Intermediate`, `Algorithm` and `Correctness` between
v4.28.0 (#455's pin) and v4.30.0 (this PR's pin) — exactly the four modules re-derived here — while
#455 was independently adding an `AdditiveNTT/Domain.lean` inside ArkLib. Two repos in the same org
built the same layering at the same time. This PR has already partly corrected course by importing
CompPoly's `NovelPolynomialBasis`, the module #455 still vendors.

Full analysis with numbers is in the review notes; the short version:

**The plug-in surface is small.** Binius + RingSwitching reference only **28** of this file's 86
declarations (the other 58 are internal scaffolding). Of those 28: **11 are character-identical to
CompPoly** — including `sDomain`, the hottest at 312 references — accounting for ~80 % of all
usage; **11 are repackaged** (dominated by `iteratedQuotientMap` at 46 refs and
`intermediateEvaluationPoly` at 27); **6 are genuinely new**.

**The root cause is that CompPoly uses three index types for one concept:**

| CompPoly declaration | round-index type |
| --- | --- |
| `sDomain`, `qMap`, `sBasis`, `normalizedW` | `Fin r` |
| `intermediateNormVpoly`, `intermediateNovelBasisX`, `intermediateEvaluationPoly` | `Fin (ℓ + 1)` |
| `iteratedQuotientMap`, `NTTStage_correctness` | `Fin ℓ` |

Binius carries `Fin r` everywhere (`sDomain : Fin r → …`, `β : Fin r → L`, `OracleStatement` is
`Fin r`-indexed), so each crossing into the polynomial layer means rebuilding `⟨i.val, proof⟩` and
then relating `sDomain ⟨i.val, _⟩` back to `sDomain i`. Your file unifies all three on `Fin r` plus
a side condition, and passes the destination index explicitly (`destIdx` + `h_destIdx`) rather than
computing it in the result type. That is what removes the cast burden.

**And one of the differences is a genuine defect, not a preference.**
`CompPoly.iteratedQuotientMap (i : Fin ℓ)` **cannot express `i = ℓ`**. Binius needs exactly that:
`fiberwiseDisagreementSet` (`Code.lean:129`) is stated with `h_destIdx_le : destIdx ≤ ℓ`, not `<`,
and `fiberwiseDisagreementSet_steps_zero_eq_disagreementSet` instantiates
`iteratedQuotientMap (i := destIdx) (k := 0)` at that bound; the final fold lands on domain index
exactly `ℓ`. Your `(i : Fin r)` with `i.val + k ≤ ℓ` admits it. **CompPoly should absorb this
regardless of what happens to this PR.**

So the ask here is not "stop duplicating" — it is "please push the fix upstream".

**Actionable:**

1. **Now, in this PR (one line, unblocks everything else):** rename to
   `namespace Binius.AdditiveNTT`. The same-qualified-name collision is the only part of this that
   is actively dangerous today, and it is independent of any cross-repo work.
2. **The CompPoly side is on me — I have filed https://github.com/Verified-zkEVM/CompPoly/issues/281.** It asks for:
   widening `iteratedQuotientMap` to `Fin r` + `i.val + k ≤ ℓ` with an explicit `destIdx` —
   strictly more general, no consumer breakage; then migrating `intermediateNormVpoly` /
   `intermediateNovelBasisX` / `intermediateEvaluationPoly` from `Fin (ℓ+1)` to `Fin r` + `i ≤ ℓ`
   with deprecated wrappers (one-liners: `destIdx := ⟨i+k, _⟩`, `h_destIdx := rfl`). You do not
   need to drive that; please just follow along and say if the proposed signatures would not in
   fact suit Binius.
3. **Then** move your 6 new declarations upstream and delete this file. Your proofs already compile,
   so the bulk is a move; the real work is re-basing four spec↔computable bridges
   (`computableAdditiveNTT_eq_additiveNTT`, `computableAdditiveNTTFast_eq_additiveNTT`,
   `computableNTTStage_eq_NTTStage`, `computableTwiddleFactor_eq_twiddleFactor`).
4. **Independently, and worth double-checking rather than taking from me:** consider dropping
   `[NeZero ℓ]` where it is not needed. Your variable block adds it and CompPoly's does not, and 8
   shared declarations carry it in their elaborated type. Two are clearly vestigial (CompPoly's
   versions take `Fin ℓ`, which already forces `ℓ > 0`). For the other six — notably
   `additiveNTT_correctness` and `initial_tiled_coeffs_correctness`, whose statements are otherwise
   character-identical to CompPoly's — it looks like a silent narrowing off `ℓ = 0`.

   Caveat on that: I inferred the six are provable *without* the hypothesis from the fact that
   CompPoly proves the same statements without it; I did not actually strip it and rebuild. So
   please treat this as "worth a look" rather than a confirmed defect — if one of them genuinely
   needs `ℓ > 0`, the interesting question is why CompPoly's version does not.

ArkLib/ProofSystem/Binius/FRIBinius/General.lean
```diff
@@ -200,28 +208,300 @@ theorem fullOracleReduction_perfectCompleteness :
...
+ring-switching batching (`κ/|L|`), core-interaction sumcheck + fold
+(`2·ℓ'/|L| + 2^(ℓ'+𝓡)/|L|`, Props 4.23), and query-phase proximity (`(…)^γ`, Prop 4.24).
+This formalization proves the stronger knowledge-soundness statement with the same scalar. -/
+
+/-- Single-repetition proximity testing error `1/2 + 1/(2·2^𝓡)` (third factor of DP24 §5.2 (43)). -/
+def querySingleRepetitionError : ℝ≥0 :=
+  (1 / 2 : ℝ≥0) + 1 / (2 * 2 ^ 𝓡)
+
+/-- Concrete KS upper bound for full FRI-Binius (DP24 §5.2 eq. (43) / Construction 5.1). -/
+def concreteFRIBiniusKnowledgeError : ℝ≥0 :=
```
alexanderlhicks, line 350: > `def concreteFRIBiniusKnowledgeError : ℝ≥0 :=`

Checked against the paper and the arithmetic is **faithful** — recording that explicitly so it is
not re-litigated. DP24 §5.2 eq. (43) reads

```
(κ + 2·ℓ')/|L|  +  2^(ℓ'+R)/|L|  +  (1/2 + 1/(2·2^R))^γ
```

with the paper attributing the summands to ring-switching+sumcheck, Proposition 4.23, and
Proposition 4.24 respectively. This definition matches it term for term, and `BinaryBasefold`'s
`concreteBinaryBasefoldKnowledgeError` matches it with the `κ` ring-switching term dropped, exactly
as its docstring states. Prop 4.23's proof gives `≤ 2^{ℓ+R}/|L|`, matching the second summand. Good.

The accuracy issue is the *notion*, not the number. DP24 says: "It follows essentially from the
proofs of Theorems 3.5 and 4.17 that Construction 5.1's concrete **soundness** error is bounded
from above by (43)." This file states `knowledgeSoundness` with the same scalars. That is a
**strengthening** of the cited result, not a transcription of it.

I think the strengthening is defensible — you build genuine RBR extractors per step, so knowledge
soundness is earned rather than assumed — but the justification is *your* extractor construction,
not Theorems 3.5/4.17. As written, a reader chasing the citation will not find the knowledge claim
in the paper.

**Actionable:** extend the existing audit note (it already flags this, which is good practice) to
say where the knowledge claim actually comes from — i.e. that each per-round charge is discharged
by an explicit `Extractor.RoundByRound` in `Steps/*`, and that DP24 is cited only for the
*numerical* bound. One or two sentences.

**I audited whether the strengthening is legitimate, and it is — with a bonus.** ArkLib's
`Verifier.rbrKnowledgeSoundness` is the *averaged* notion: the transcript prefix is prover-sampled
and the probability ranges over prefix and challenge together. #668 adds
`rbrKnowledgeSoundnessWorstCase` and notes in its docstring that worst-case-per-prefix is "the
standard literature shape", and that the averaged form does **not** imply it. So the notion your
top-level theorems are stated in is weaker than the literature's.

Your proofs, however, discharge the *worst-case* obligation. Both helpers you route through —
`rbrKnowledgeSoundness_of_{2msg_PtoV,1msg_VtoP}_uniformChallenge` in your `Completeness.lean`
(`:1228`, `:1326`) — take `hbound` universally quantified over the statement **and the whole
prefix**, with probability over the fresh challenge only. That is exactly
`rbrKnowledgeSoundnessWorstCase`'s inner obligation, and all four per-step results
(`Steps/Fold`, `QueryPhase`, `BatchingPhase`, `SumcheckPhase`) supply it in that shape. So the
mathematics is at literature strength; only the packaging is the weaker notion, and only because
that is what `main` currently offers.

**Cheap follow-up, once #668 lands:** restate the per-step results in
`rbrKnowledgeSoundnessWorstCase` and get the averaged form from
`rbrKnowledgeSoundnessWorstCase_implies_rbrKnowledgeSoundness`. You already prove the stronger
hypothesis, so it is close to free and upgrades the advertised guarantee.

One speculation, flagged as such: the unproved `append_rbrKnowledgeSoundness` may be hard *because*
the ambient notion is averaged — under composition the second protocol's prefix distribution is
induced by the first's execution, which a worst-case bound survives uniformly and an averaged one
does not obviously. #668 ships both the worst-case notion and the mixture master bounds in
`RbrGame.lean`, which is the machinery such a proof would want. I have not shown averaged RBR-KS
fails to compose; I mention it because it suggests #668 may be the enabler for the very theorem
this PR is blocked on.

---

ArkLib/ProofSystem/RingSwitching/Profile.lean
```diff
@@ -91,6 +91,33 @@ structure RingSwitchingProfile (B L : Type*) (κ : ℕ)
   decomposeRows_spec : ∀ z : A, z = ∑ u, φ₀ (decomposeRows z u) * φ₁ (basis u)
   /-- **Column reconstruction law**: the right/`φ₁`-action dual of `decomposeRows_spec`. -/
   decomposeColumns_spec : ∀ z : A, z = ∑ v, φ₁ (decomposeColumns z v) * φ₀ (basis v)
+  /-- **Column additivity**: `decomposeColumns` is additive in its `A`-argument (it is a genuine
+  coordinate/`repr` map, not a law-free function). Binius: `map_add` of `baseChangeRight.repr`;
+  Hachi: linearity of `ψ⁻¹` (Theorem 2, `ψ` linearly homomorphic). -/
+  decomposeColumns_add : ∀ (z z' : A) (v : Fin κ → Fin 2),
```
alexanderlhicks, line 97: > `decomposeColumns_add : ∀ (z z' : A) (v : Fin κ → Fin 2),`

**Credit first, because this is the most valuable non-Binius thing in the PR.** The five laws added
here (`decomposeColumns_add`, `decomposeColumns_tmul`, `decomposeRows_add`, `decomposeRows_tmul`,
`decomposeColumns_injective`) close the "law-free profile" gap that was raised against the generic
ring-switching scaffold: with only `decompose{Rows,Columns}_spec`, nothing forced these to be
genuine coordinate maps, and the batching completeness proof could not be closed generically. Your
docstrings give the Binius *and* Hachi discharge for each, and the laws are load-bearing rather than
decorative — `decomposeColumns_injective` feeds `batchingMismatchPoly_nonzero_of_ne`
(`BatchingPhase.lean:1034`) which feeds the Schwartz–Zippel step of
`batchingOracleVerifier_rbrKnowledgeSoundness` (`:1297`). Adding `[IsDomain K] [IsDomain L]` to the
batching theorems addresses the second recorded blocker. `BatchingPhase` is 90 declarations, all
axiom-clean (verified by `#print axioms`).

Also good, and worth stating: this PR **does not fork** the abstraction. It parameterizes over
`RingSwitchingProfile` throughout (`Spec`, `General`, `Prelude`, `BatchingPhase`, `SumcheckPhase`)
and instantiates via `binaryTowerProfile`/`bbfProfile`. That is exactly the intended use.

Worth crediting separately: this PR **promotes** `RingSwitching/*` out of
`ArkLib/ProofSystem/Binius/RingSwitching/` to top-level `ArkLib/ProofSystem/RingSwitching/`. That is
the right direction for a profile that Binius and Hachi are both meant to instantiate, and it is
what makes the #615 generalization coherent. (#455 predates the move and still uses the nested
layout — another reason it needs a rebase.)

**The coordination problem, which is mine as much as yours.** My draft PR #615 is generalizing this
same structure, and the two PRs touch all six `RingSwitching/*.lean` files — this one adds +4,609
lines there, #615 adds +1,884, and `General.lean` already conflicts on merge with `main`. Whoever
lands second faces a substantial and *semantic* — not textual — reconciliation, because both change
what the profile is required to guarantee.

**Actionable:** let's settle a landing order before either merges — I'm happy for this one to go
first and to rebase #615 onto it, given this is the PR with the proofs attached. Two things would
help me either way:

1. Say in the PR body which of the five laws you consider stable API.
2. Tell me whether you think they survive a generalization. They are stated for the DP24 tensor
   formulation; whether they hold in the form #615 needs is a design question I would rather answer
   with your input than guess at. If they do not survive as stated, that is worth knowing now,
   because it means the profile needs a different decomposition rather than five more fields.

---

ArkLib/ProofSystem/Binius/BinaryBasefold/Steps/Fold.lean
```diff
@@ -0,0 +1,1762 @@
...
+      exact h_rel
+    · constructor
+      · rw [verStmtOut_eq, prvStmtOut_eq]; rfl
+      · rw [verOStmtOut_eq, prvOStmtOut_eq];
+        exact h_agree.2
+
+open scoped NNReal
+
+open Classical in
+/-! Definition of the per-round RBR KS error for Binary FoldFold.
```
alexanderlhicks, line 352: > `- err_SC = 2/|L| (Schwartz-Zippel for degree 1)`

The bound is right, the justification in the comment is not. The sumcheck round polynomial here is
`L⦃≤ 2⦄[X]` (see `sumcheckConsistencyProp`, `Relations.lean:220`, which takes
`H : L⦃≤ 2⦄[X Fin k]`), so this is degree-**2** Schwartz–Zippel and `2/|L|` is exactly right.
Degree 1 would give `1/|L|` and would make the stated error look loose by a factor of two when it
is in fact tight.

**Actionable:** `(Schwartz-Zippel for degree 2)`.

---

ArkLib/ProofSystem/Binius/BinaryBasefold/CoreInteractionPhase.lean
```diff
@@ -38,9 +41,11 @@ We define `(P, V)` as the following IOP, in which both parties have the common i
 namespace Binius.BinaryBasefold.CoreInteraction
 
 noncomputable section
-open OracleSpec OracleComp ProtocolSpec Finset AdditiveNTT Polynomial MvPolynomial
+open OracleSpec OracleComp ProtocolSpec Finset AdditiveNTT Polynomial MvPolynomial Equiv
 open scoped NNReal
 
+set_option linter.style.longFile 2600
```
alexanderlhicks, line 47: > `set_option linter.style.longFile 2600`

This opt-out is below the file's actual length (2,868 lines), so the linter still fires. The build
log shows eight PR files over the 1,500-line cap:

| file | lines | opt-out |
| --- | ---: | --- |
| `BinaryBasefold/QueryPhase.lean` | 2,882 | none |
| `BinaryBasefold/CoreInteractionPhase.lean` | 2,868 | 2600 — too low |
| `BinaryBasefold/Prelude.lean` | 2,857 | none |
| `ToVCVio/Simulation.lean` | 2,485 | none |
| `BinaryBasefold/Basic.lean` | 1,936 | none |
| `BinaryBasefold/Steps/FinalSumcheck.lean` | 1,924 | none |
| `BinaryBasefold/Soundness/QueryPhaseSoundness.lean` | 1,837 | none |
| `BinaryBasefold/Steps/Fold.lean` | 1,763 | none |

Alongside these, the build emits on PR-owned files: **280** lines over the 100-character limit,
**263** unused `simp` arguments, **116** unused auto-bound section variables, **95** flexible-tactic
warnings, **21** deprecated-lemma uses (e.g. `Fin.lt_iff_val_lt_val` → `Fin.lt_def` at
`QueryPhaseSoundness.lean:1521`), **14** `maxHeartbeats` bumps without the required explanatory
comment, and **12** uses of a tactic the linter says should only mark intermediate goal states.

None of this is a correctness concern, and CI passes because these are warnings. But
`CLAUDE.md`/`AGENTS.md` set the 1500 cap as the norm and the repo asks for justified opt-outs.

**Actionable:** prefer splitting the four ~2.9k files along the seams they already have (e.g.
`QueryPhase` → spec / prelims / soundness) over raising the caps; where a cap is genuinely right,
set it above the real line count and add the one-line justification the linter asks for. The
line-length and unused-`simp` sweeps are mechanical and would remove ~550 warnings.

---

ArkLib/ProofSystem/Binius/BinaryBasefold/General.lean
```diff
@@ -34,6 +37,7 @@ variable (β : Fin r → L) [hβ_lin_indep : Fact (LinearIndependent 𝔽q β)]
   [h_β₀_eq_1 : Fact (β 0 = 1)]
 variable {ℓ 𝓡 ϑ : ℕ} (γ_repetitions : ℕ) [NeZero ℓ] [NeZero 𝓡] [NeZero ϑ] -- Should we allow ℓ = 0?
 variable {h_ℓ_add_R_rate : ℓ + 𝓡 < r} -- ℓ ∈ {1, ..., r-1}
```
alexanderlhicks, line 39: *(anchored at line 39; the declaration in question is `variable [hdiv : Fact (ϑ ∣ ℓ)]` at line 41, which is unchanged by the diff and so not directly commentable.)*

> `variable [hdiv : Fact (ϑ ∣ ℓ)]`

DP24 §5.2 is explicit that this hypothesis is dispensable: *"This measure further allows us to drop
our assumption ϑ | ℓ′, which we picked up merely for notational convenience in the first place."*
Hard-wiring it as a `Fact` instance means the formalization covers a strictly smaller parameter
range than the paper's Construction 5.1.

That is a perfectly reasonable simplification for a first pass and I would not block on it, but it
should be visible to a reader rather than discovered.

**Actionable:** note the restriction in the module docstring, with the DP24 §5.2 pointer, and add a
TODO for the early-termination variant that removes it.

---

ArkLib/ToVCVio/Simulation.lean
```diff
@@ -0,0 +1,2484 @@
+/-
```
alexanderlhicks, line 1: 2,484 + 839 lines destined for VCVio. Overlap against the pinned VCVio is low — 7 shared
declaration names out of 141 — but I checked each rather than just counting, and two are genuine
redundancies you can drop today:

**Redundant — VCVio already has these, more generally:**

- `liftComp_self` (`Lemmas.lean:546`) is `abbrev liftComp_self [IsUniformSpec spec] (oa) :
  liftComp oa spec = oa := liftComp_id oa`. VCVio proves `liftComp_self`
  (`OracleComp/Coercions/SubSpec.lean:244`) with **no** `IsUniformSpec` hypothesis. So this is a
  thin alias that is also strictly weaker than the thing it shadows.
- `probFailure_liftComp` (`Lemmas.lean:488`) states `Pr[⊥ | liftComp oa superSpec] = Pr[⊥ | oa]`
  under `[IsUniformSpec spec] [IsUniformSpec superSpec]`. VCVio proves the identical statement with
  neither hypothesis (`SubSpec.lean:316`).

**Probably redundant, worth one look:** `support_bind` (`Lemmas.lean:198`) and
`mem_support_bind_iff` (`:205`) are your `OptionT`-specialised versions of VCVio's
`EvalDist/Monad/Basic.lean:112,117`, which are already generic in `m` given
`[MonadLiftT m SetM] [LawfulMonadLiftT m SetM]`. If `OptionT m` carries those instances — and
VCVio ships `EvalDist/Instances/OptionT.lean` — yours should follow from VCVio's directly.

**Not a duplicate, and arguably the wrong way round:** `simulateQ_liftComp`. Yours
(`Simulation.lean:171`) is general in `{spec₁ spec₂}` with `[MonadLift (OracleQuery spec₁)
(OracleQuery spec₂)]`; VCVio's (`QueryTracking/RandomOracle/Simulation.lean:61`) is specialised to
the random-oracle `hashSpec`. Same name, different lemmas — and yours is the more general one, so
it is a good upstream candidate rather than something to delete.

`probFailure_liftM` I did not compare; VCVio has declarations of that name in both
`EvalDist/Instances/ErrorT.lean` and `.../OptionT.lean`.

**On upstreaming generally:** worth opening the VCVio PR now rather than after merge. The precedent
is concrete and still live — `ArkLib/ToVCVio/OracleComp/SimSemantics/SimulateQ.lean` was meant to be
deleted once the corresponding VCVio work landed, and it is still sitting on `main` today. The
longer this material lives in `ToVCVio/`, the more likely it acquires the same fate.

ArkLib/ProofSystem/Binius/BinaryBasefold/Relations.lean
```diff
@@ -0,0 +1,923 @@
...
+    ⟩
+    Binius.BinaryBasefold.foldingBadEvent 𝔽q β (h_ℓ_add_R_rate := h_ℓ_add_R_rate)
+      (i := curOracleDomainIdx) (steps := ϑ) (destIdx := destIdx) (by rfl) (by dsimp only [destIdx]; simp only [oracle_index_add_steps_le_ℓ]) (f_i := f_k) (r_challenges :=
+        getFoldingChallenges (r := r) (𝓡 := 𝓡) stmtIdx challenges (k := j.val * ϑ) (h := by
+        simp only [curOracleDomainIdx] at hj
+        exact hj
+      ))
+  else False
+
+lemma foldingBadEventAtBlock_snoc_castSucc_eq (i : Fin ℓ)
```
alexanderlhicks, line 262: > `  else False`

Recording this because it was a known defect and it is now fixed. An earlier audit of the Binius
relations flagged `foldingBadEventAtBlock`'s guard falling through to `True`, which made
`roundRelation i` degenerate to `⊤` for `i < ℓ` and would have made the RBR knowledge-soundness
statements vacuous. The guard now yields `False`.

I also checked the remaining vacuity route and it is closed: `masterKStateProp` is
`badEventExists ∨ good`, so if `badEventExists` were satisfiable at `stmtIdx = 0` the input relation
would be `⊤`. It is not — at `stmtIdx = 0` the block index gives `k = min ϑ 0 = 0`, and your
`incrementalFoldingBadEvent_of_k_eq_0_is_false` (`Compliance.lean:275`, axiom-clean) proves the
event false there: the close branch compares a set to itself, and the else branch asserts exactly
the negated guard. `relIn` is therefore the honest `good`.

---

ArkLib/ProofSystem/Binius/BinaryBasefold/General.lean
```diff
@@ -98,41 +102,43 @@ noncomputable def fullOracleProof :
...
 /-- Perfect completeness for the full Binary Basefold protocol (reduction) -/
-theorem fullOracleReduction_perfectCompleteness :
+theorem fullOracleReduction_perfectCompleteness (hInit : NeverFail init) :
   OracleReduction.perfectCompleteness
     (oracleReduction := fullOracleReduction 𝔽q β γ_repetitions (ϑ:=ϑ)
-      (h_ℓ_add_R_rate := h_ℓ_add_R_rate) )
-    (relIn := roundRelation (mp := BBF_SumcheckMultiplierParam) 𝔽q β (ϑ:=ϑ)
-      (h_ℓ_add_R_rate := h_ℓ_add_R_rate) 0)
+      (h_ℓ_add_R_rate := h_ℓ_add_R_rate) (𝓑:=𝓑))
+    (relIn := strictRoundRelation (mp := BBF_SumcheckMultiplierParam) 𝔽q β (ϑ:=ϑ)
```
alexanderlhicks, line 119: > `    (relIn := strictRoundRelation (mp := BBF_SumcheckMultiplierParam) 𝔽q β (ϑ:=ϑ)`

**Completeness and knowledge soundness for the same protocol are stated over two different input
relations, and nothing in the PR relates them.**

- `fullOracleReduction_perfectCompleteness` (this line) uses `strictRoundRelation … 0`
- `fullOracleVerifier_rbrKnowledgeSoundness` (`:157`) uses `roundRelation … 0`

These are genuinely different definitions. `strictRoundRelationProp` is
`sumCheckConsistency ∧ witnessStructuralInvariant ∧ strictOracleFoldingConsistency`, where the
strict clause ties **every** oracle back to `f₀`, the encoding of `wit.t` (`Relations.lean:799`).
`roundRelationProp` goes through `masterKStateProp`, which is
`badEventExists ∨ (localChecks ∧ structural ∧ initial ∧ oracleFoldingConsistency)` — the non-strict
clause ties **consecutive** oracles to each other by folding.

I grepped the tree: there is no `strictRoundRelation i ⊆ roundRelation i`, no equality at `0`, and
no bridge lemma of any kind. `strictRoundRelation` appears only in completeness-side files
(`General`, `ReductionLogic`, `FRIBinius/CoreInteractionPhase`) and `roundRelation` only on the
soundness side.

Why it matters: as stated, the two theorems do not compose into a security statement. Completeness
holds on the smaller relation and extraction lands in the larger one, so a reader cannot conclude
"if the verifier accepts, the prover knows a witness of the kind completeness assumed" — the two
guarantees are about different sets and the gap between them is unquantified.

I do **not** think this is a soundness hole. At `i = 0` there is exactly one oracle
(`toOutCodewordsCount ℓ ϑ 0 = 1`), the bad-event disjunct is provably false there (comment 9), and
the relaxed relation's `initial` clause still pins the first oracle to the encoding of `wit.t` — so
the extracted witness does look meaningful. But that argument is mine, not the library's, and it
should not have to be reconstructed by a reader.

**Actionable:** add `strictRoundRelation … 0 ⊆ roundRelation … 0` and state the composed guarantee
in the module docstring. I got most of the way there, so here is the skeleton rather than just the
ask — this compiles against your tree with one `sorry`:

```lean
theorem strict_subset_round_zero :
    strictRoundRelation (mp := mp) (𝓑 := 𝓑) 𝔽q β (ϑ := ϑ)
        (h_ℓ_add_R_rate := h_ℓ_add_R_rate) 0
      ⊆ roundRelation (mp := mp) (𝓑 := 𝓑) 𝔽q β (ϑ := ϑ)
        (h_ℓ_add_R_rate := h_ℓ_add_R_rate) 0 := by
  rintro ⟨⟨stmt, oStmt⟩, wit⟩ h
  simp only [strictRoundRelation, strictRoundRelationProp, Set.mem_setOf_eq,
    strictOracleWitnessConsistency] at h
  obtain ⟨hSum, hStruct, hStrictFold⟩ := h
  simp only [roundRelation, roundRelationProp, masterKStateProp, Set.mem_setOf_eq]
  right
  refine ⟨hSum, hStruct, ?_, ?_⟩
  · -- `initial` — the one real obligation, see below
    have hcount : toOutCodewordsCount ℓ ϑ
        (OracleFrontierIndex.mkFromStmtIdx (0 : Fin (ℓ + 1))).val = 1 := by
      rw [OracleFrontierIndex.val_mkFromStmtIdx]; exact toOutCodewordsCountOf0 ℓ ϑ
    have h0 := hStrictFold ⟨0, by omega⟩
    simp only [firstOracleWitnessConsistencyProp, getFirstOracle]
    sorry
  · -- `oracleFoldingConsistencyProp` at i = 0 — vacuous, and this branch is DONE
    intro j hj
    exfalso
    have hcount : toOutCodewordsCount ℓ ϑ
        (OracleFrontierIndex.mkFromStmtIdx (0 : Fin (ℓ + 1))).val = 1 := by
      rw [OracleFrontierIndex.val_mkFromStmtIdx]; exact toOutCodewordsCountOf0 ℓ ϑ
    omega
```

Two things this establishes. The decomposition is right — `right; refine ⟨hSum, hStruct, ?_, ?_⟩`
typechecks, so the relaxed relation's `localChecks` and `structural` are *literally* the strict
one's, and only `initial` and `oracleFoldingConsistency` need work. And the second branch is
**closed**: at `i = 0` there is one oracle, so `oracleFoldingConsistencyProp`'s `j + 1 < count`
premise is unsatisfiable.

That leaves `initial` as the single obligation, and it is mathematically trivial —
`hStrictFold` at `j = 0` says `oStmt 0` *is* `f₀ = polyToOracleFunc P₀`, the encoding of `wit.t`,
so the two are at distance 0 and `pair_UDRClose` reduces to `0 < BBF_CodeDistance 0`. What stopped
me is plumbing, not mathematics: reducing `steps := 0 * ϑ` to `0` so `iterated_fold_zero_steps`
applies. `simp only [zero_mul]` will not fire, because the numeral sits in a dependent position
with `h_destIdx`/`h_destIdx_le` depending on it, so the motive is ill-typed. You own the
index-congruence idiom for exactly this (`sDomain_eq_of_eq`, `iteratedQuotientMap_congr_k`), so
this is a few minutes for you and was not for me.

If the containment turns out to fail rather than merely be fiddly, that is much more interesting
than the lemma and should be documented as a known gap instead.

---

alexanderlhicks (changes requested):
Thanks for this — it is a very large and, in the parts that are yours, a very solid piece of work.
I built it at your pins (clean, 4056 jobs) and ran a full axiom census over all 37 PR-owned
modules (`Lean.collectAxioms` on every non-internal constant). Headline number:

**2,013 axiom-clean declarations / 89 `sorryAx`-tainted (4.2 %). No `sorryAx` originates in this
PR.**

What is genuinely proved is substantial and I want to be explicit about it:

- The DP24 soundness core is real and axiom-clean: `prop_4_21_bad_event_probability`,
  `lemma_4_24_dist_folded_ge_of_last_noncompliant`, and all of
  `Soundness/{BadBlocks,FoldDistance,Incremental,Lift,Proposition4_21,QueryPhasePrelims,
  QueryPhaseSoundness}` (127 decls), `QueryPhase` (61), `Relations` (61), `ReductionLogic` (95).
- All four per-step reductions (`Steps/{Fold,Commit,Relay,FinalSumcheck}`, 157 decls) are clean,
  with **real** extractors — `foldRbrExtractor` reconstructs the mid-round witness via
  `projectToMidSumcheckPoly`/`getMidCodewords`, it is not a constant or `default`.
- Both concrete bounds match **DP24 §5.2 eq. (43)** exactly (checked against the paper, below).
- The relations are **not** vacuous: `incrementalFoldingBadEvent_of_k_eq_0_is_false`
  (`Compliance.lean:275`) closes the bad-event escape hatch at `stmtIdx = 0`, and the old
  `foldingBadEventAtBlock` `else True` defect is fixed to `else False` (`Relations.lean:262`).
- `BatchingPhase` (90 decls, clean) resolves both blockers recorded against the generic
  ring-switching scaffold — see the `Profile.lean` comment.

The blocking items are (1) the PR body's proof claim, (2) the AdditiveNTT/CompPoly duplication,
(3) merge mechanics. Details inline.

**On #455 and sequencing.** I looked at #455 ("computable FRI-Binius protocols") since it is stacked
on this branch, and I want to be clear that it is *not* an argument for holding this PR. #455
currently carries **229 sorries** with its "Proof migration" box unchecked, while this PR has 0 and
real proofs — blocking proved content behind an unproved refactor would be the wrong ordering. #455
is also 113 commits behind this branch, pins CompPoly v4.28.0 against your v4.30.0, and still uses
the pre-move `Binius/RingSwitching/` layout. Its rebase cost grows the longer this sits, so landing
this promptly serves #455 too. Your April framing — checkpoint the security proofs, migrate
gradually — looks right to me and is corroborated rather than undermined by #455's current state.

---

---

Better than the `CONFLICTING` label suggests — **5 conflicting files**, not 56:
`BinaryBasefold/CoreInteractionPhase.lean`, `BinaryBasefold/QueryPhase.lean`,
`BinaryBasefold/Steps.lean`, `Component/SendClaim.lean`, `RingSwitching/General.lean`.

The toolchain gap is the bigger job: this PR is on `leanprover/lean4:v4.30.0` while `main` is on
`v4.31.0`, with matching mathlib / VCVio / CompPoly / aesop pin bumps. Expect the `Fin`-heavy
dependent-cast proofs (`Prelude.lean`, `QueryPhasePrelims.lean`) to need attention, and note the 21
already-deprecated lemmas will likely become errors or new warnings across the bump.

**Actionable:** rebase onto current `main` and bump to v4.31 in a separate commit from any content
change, so the reviewer can see the bump breakage in isolation.

**Landing order.** #455 is based on `completeness-of-binius`, so it inherits whatever happens here.
Once this rebases, #455 will need to rebase onto the new head (or onto `main`) — it is currently 113
commits, a CompPoly v4.28→v4.30 bump, and the `Binius/RingSwitching/` → `RingSwitching/` relocation
adrift. Worth doing that before more work goes into #455, and worth enabling a build check on it:
its last CI run was `summarize` only, so nothing has compiled it.

ArkLib/ProofSystem/Binius/BinaryBasefold/Code.lean
```diff
@@ -0,0 +1,1118 @@
...
+              (k := steps) (h_k := by omega)) = y_of_x := by
+          simp only [iteratedQuotientMap, XX, y_of_x];
+        have h_eval_diff : f ⟨XX, by omega⟩ ≠ g ⟨XX, by omega⟩ := by
+          unfold XX
+          simp only [Subtype.coe_eta, ne_eq, hx_in_ΔH, not_false_eq_true]
+        simp only [h_forward_iterated_qmap, Subtype.coe_eta, h_eval_diff,
+          not_false_eq_true, and_self]
+      simp only [h_elemenet_Y_bad, true_and]
+      set qMapFiber := qMap_total_fiber 𝔽q β (i := i) (steps := steps)
+        h_destIdx h_destIdx_le (y := y_of_x)
```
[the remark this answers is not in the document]
chung-thai-nguyen: Addressed locally. Renamed the two scaled-distance facts to `h_fiberwise_scaled_le` and `h_scaled_distance_eq`, removing the local-name shadowing. Please verify in the next push and resolve if this addresses the clarity concern.

ArkLib/ProofSystem/Binius/BinaryBasefold/Soundness/BadBlocks.lean
```diff
@@ -0,0 +1,468 @@
...
+-/
+
+namespace Binius.BinaryBasefold
+
+set_option maxHeartbeats 400000
+
+open OracleSpec OracleComp ProtocolSpec Finset AdditiveNTT Polynomial MvPolynomial
+  Binius.BinaryBasefold
+open scoped NNReal
+open ReedSolomon Code BerlekampWelch Function
```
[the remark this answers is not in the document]
chung-thai-nguyen: Addressed locally. Added concise comments explaining the `400000` heartbeat budget in the bad-block, lift, and query-phase soundness files. Please verify in the next push and resolve if this is sufficient.

ArkLib/ProofSystem/Binius/BinaryBasefold/Soundness/FoldDistance.lean
```diff
@@ -0,0 +1,261 @@
...
+    intro x hx
+    by_contra h_neq
+    exact h_not_in_fiber_diff ⟨x, (by simp only [SetLike.coe_mem]), (by simp only [Subtype.coe_eta]; constructor; exact hx; exact h_neq)⟩
+  have h_fold_eq := fold_agreement_of_fiber_agreement 𝔽q β i steps h_destIdx h_destIdx_le f g (r_challenges := r_challenges) (y := y) h_agree_on_fiber
+  exact hy_mem h_fold_eq
+
+/-- **Lemma 4.25**
+For `i*` where `f^(i)` is non-compliant, `f^(i+ϑ)` is UDR-close, and the bad event `E_{i*}`
+doesn't occur, the folded function of `f^(i)` is not UDR-close to the UDR-decoded codeword
+of `f^(i+ϑ)`. -/
```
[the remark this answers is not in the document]
chung-thai-nguyen: Addressed locally. Renamed the DP24-derived declaration to match its cited result, Lemma 4.25, and updated its references. Please verify in the next push and resolve if this looks right.

ArkLib/ProofSystem/Binius/BinaryBasefold/Soundness/Lift.lean
```diff
@@ -0,0 +1,821 @@
...
+def getLiftPoly (i : Fin ℓ) (steps : ℕ) {destIdx : Fin r}
+    (h_destIdx : destIdx.val = i.val + steps) (h_destIdx_le : destIdx ≤ ℓ)
+    (V_codeword : ((BBF_Code 𝔽q β (h_ℓ_add_R_rate := h_ℓ_add_R_rate) destIdx)
+      ^⋈(Fin (2 ^ steps)))) : L⦃<2^(ℓ-i)⦄[X] := by
+  have h_ℓ_lt_r : ℓ < r := by
+    have h_pos : 0 < 𝓡 := Nat.pos_of_neZero (n := 𝓡)
+    exact lt_trans (Nat.lt_add_of_pos_right (n := ℓ) (k := 𝓡) h_pos) h_ℓ_add_R_rate
+  have h_i_lt_r : (i : Nat) < r := lt_trans i.isLt h_ℓ_lt_r
+  let iR : Fin r := ⟨i, h_i_lt_r⟩
+  refine ⟨intermediateEvaluationPoly 𝔽q β (h_ℓ_add_R_rate := h_ℓ_add_R_rate)
```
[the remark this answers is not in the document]
chung-thai-nguyen: Addressed locally. Renamed this result to `lemma_4_22_interleaved_word_UDR_far` to agree with the cited paper result, and updated its call sites. Please verify in the next push and resolve if this looks right.

ArkLib/ProofSystem/Binius/BinaryBasefold/Steps/Commit.lean
```diff
@@ -0,0 +1,524 @@
...
+  · -- Peel off monadic layers to reach the core verifier logic
+    dsimp only [commitOracleProver, commitOracleVerifier, OracleVerifier.toVerifier,
+    FullTranscript.mk1]
+    simp only [probFailure_bind_eq_zero_iff]
+    conv_lhs =>
+      simp only [liftComp_eq_liftM, liftM_pure, probFailure_eq_zero]
+    rw [true_and]
+    intro inputState hInputState_mem_support
+    simp only [Fin.isValue, Message, Matrix.cons_val_zero, Fin.succ_zero_eq_one, ChallengeIdx,
+      Challenge, liftComp_eq_liftM, liftM_pure, support_pure,
```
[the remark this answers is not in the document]
chung-thai-nguyen: Addressed locally. Removed the obsolete commented proof lines from `Commit.lean`. Please verify in the next push and resolve if this is sufficient.

ArkLib/ProofSystem/Binius/BinaryBasefold/Steps/Fold.lean
```diff
@@ -0,0 +1,1762 @@
...
+                    calc
+                      j.val * ϑ + cId.val < j.val * ϑ + (i.val + 1 - j.val * ϑ) :=
+                        Nat.add_lt_add_left h_cId_lt (j.val * ϑ)
+                      _ = i.val + 1 := Nat.add_sub_of_le h_block_le⟩)
+              afterSlice := by
+          apply fin_fun_heq_of_cast hk_after
+          intro cId
+          dsimp [afterSlice]
+        exact HEq.symm h_afterSlice_heq
+    have h_bad_after_full :
```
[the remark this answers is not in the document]
chung-thai-nguyen: Addressed locally. Removed the unused plural `foldKStateProps` definition; the used singular `foldKStateProp` remains. Please verify in the next push and resolve if this looks right.

ArkLib/ToVCVio/Simulation.lean
```diff
@@ -0,0 +1,2484 @@
...
+    Pr[⊥ | simulateQ so oa] = 0 := by
+  simp only [probFailure_of_liftM_PMF]
+
+/--
+Safety preservation: A simulated protocol is safe if and only if the original
+protocol is safe. This requires:
+1. The implementation itself never fails (h_so).
+2. The implementation doesn't return "illegal" values outside the spec (h_supp).
+-/
+@[simp]
```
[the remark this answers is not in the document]
chung-thai-nguyen: Addressed locally. Updated the docstring to state the actual zero-failure property of `OracleComp`. Please verify in the next push and resolve if this is accurate.

ArkLib/ToVCVio/Simulation.lean
```diff
@@ -0,0 +1,2484 @@
...
+    {oSpec : OracleSpec ι} [IsUniformSpec oSpec] {σ : Type}
+    (impl : QueryImpl oSpec (StateT σ ProbComp))
+    {α : Type} (oa : OracleComp oSpec α) (s : σ) :
+    Pr[⊥ | (simulateQ impl oa).run' s] = 0 ↔ Pr[⊥ | oa] = 0 := by
+  simp only [probFailure_of_liftM_PMF]
+
+/-- **Safety Preservation Lemma for Stateless Implementations**
+
+If an oracle implementation is safe and support-faithful, then simulation preserves safety
+from the specification level to the implementation level (stateless version).
```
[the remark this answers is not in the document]
chung-thai-nguyen: Addressed locally. Updated the docstring to describe the actual zero-failure statement and removed references to nonexistent hypotheses. Please verify in the next push and resolve if this is accurate.

ArkLib/ToVCVio/Simulation.lean
```diff
@@ -0,0 +1,2484 @@
...
+      (mx := Vector.mapM (m := OptionT (OracleComp spec)) f v)
+  have step2 :
+      (simulateQ so
+          ((Vector.toArray <$> Vector.mapM (m := OptionT (OracleComp spec)) f v :
+            OptionT (OracleComp spec) (Array β))) :
+              OptionT (OracleComp superSpec) (Array β)) =
+      (simulateQ so (Array.mapM (m := OptionT (OracleComp spec)) f v.toArray) :
+          OptionT (OracleComp superSpec) (Array β)) := by
+    rw [Vector.toArray_mapM]
+  have step3 :
```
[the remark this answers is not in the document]
chung-thai-nguyen: Addressed locally. Removed the obsolete `simulateQ_array_mapM` TODO; the short docstring now matches the one-line proof. Please verify in the next push and resolve if this is sufficient.

ArkLib/ProofSystem/Binius/FRIBinius/General.lean
chung-thai-nguyen: Addressed locally. Clarified that DP24 is cited for the numerical soundness bound, while the knowledge-soundness strengthening is discharged by the explicit round-by-round extractors in the core and query phases. Please verify in the next push and resolve if the provenance is now clear.

ArkLib/ProofSystem/Binius/BinaryBasefold/Steps/Fold.lean
chung-thai-nguyen: Addressed locally. Corrected the `foldKnowledgeError` comment to degree-2 Schwartz–Zippel; `2 / |L|` was already the correct bound. Please verify in the next push and resolve if this is accurate.

ArkLib/ProofSystem/Binius/BinaryBasefold/General.lean
chung-thai-nguyen: Addressed locally. Documented the current `ϑ ∣ ℓ` / `ϑ ∣ ℓ′` restriction with the DP24 §5.2 pointer and a TODO for the early-termination variant. Please verify in the next push and resolve if this is sufficient.

ArkLib/ToVCVio/Simulation.lean
chung-thai-nguyen: Addressed locally. Removed the exact local duplicates of upstream VCVio results: `probFailure_liftComp`, its duplicate `_eq` alias, and `liftComp_self`. The specialized OptionT support lemmas remain because they are used throughout the local proofs. Please verify in the next push and resolve if this looks right.

ArkLib/ProofSystem/Binius/BinaryBasefold/General.lean
chung-thai-nguyen: Addressed locally. Added `strictRoundRelation_subset_roundRelation_zero` and documented the bridge: at round zero the strict first-oracle equality yields distance-zero UDR closeness, and the consecutive-fold condition is vacuous because the frontier has one oracle. Please verify in the next push and resolve if this is the intended bridge.

ArkLib/ProofSystem/Binius/BinaryBasefold/General.lean
chung-thai-nguyen: Addressed. I updated the PR description to point out that PR-owned files are sorry-free while there are still sorrys from OracleReduction's composition layer.

ArkLib/Data/FieldTheory/AdditiveNTT/AdditiveNTT.lean
chung-thai-nguyen: Addressed. I got PR https://github.com/Verified-zkEVM/CompPoly/pull/285 merged (handling your issue https://github.com/Verified-zkEVM/CompPoly/issues/281) so we will be able to integrate it when ArkLib migrates to v4.32.

ArkLib/ProofSystem/Binius/BinaryBasefold/Relations.lean
chung-thai-nguyen: Thanks for the clarification on the previous detection. Nothing needs to be fixed here.

ArkLib/ProofSystem/RingSwitching/Packing/Profile.lean
```diff
@@ -91,6 +91,33 @@ structure RingSwitchingProfile (B L : Type*) (κ : ℕ)
   decomposeRows_spec : ∀ z : A, z = ∑ u, φ₀ (decomposeRows z u) * φ₁ (basis u)
   /-- **Column reconstruction law**: the right/`φ₁`-action dual of `decomposeRows_spec`. -/
   decomposeColumns_spec : ∀ z : A, z = ∑ v, φ₁ (decomposeColumns z v) * φ₀ (basis v)
+  /-- **Column additivity**: `decomposeColumns` is additive in its `A`-argument (it is a genuine
+  coordinate/`repr` map, not a law-free function). Binius: `map_add` of `baseChangeRight.repr`;
+  Hachi: linearity of `ψ⁻¹` (Theorem 2, `ψ` linearly homomorphic). -/
+  decomposeColumns_add : ∀ (z z' : A) (v : Fin κ → Fin 2),
```
chung-thai-nguyen: Thanks — after checking #615 against DP24 §2.5/§3 and Hachi §3, I think this PR should land first, and #615 should rebase onto the proof-bearing concrete profile.

The five laws are stable for the DP24/Binius tensor adapter, not as fields of the universal generic carrier. They exactly express the two-sided tensor-coordinate calculation used by the present Binius batching/completeness proofs, so I would retain them as derived Binius-specific compatibility theorems/adapter laws.

They do not generalize literally to Hachi. Hachi has a packing isomorphism `ψ : (R_q^H)^(d/k) ≃ R_q`, but its opening algebra is `R_q^H` with a singleton basis; packing and opening ranks are therefore different. Its Theorem 2 proves `ψ` bijective and a traced inner-product identity, not the two coordinate-wise product-extraction equalities. Thus the additivity/injectivity content belongs to basis-derived coordinates (`repr`), while the old `..._tmul` laws are DP24-tensor-specific.

The common stable surface should be #615’s two bases, basis-derived recombination/coordinates, and `bridge_eqTilde`, with batching chosen separately. That data layer fits both Binius and Hachi. I would scope the Hachi claim to that layer: `R_q` is non-domain and the singleton opening basis makes its relocation deterministic, so it does not instantiate the existing Schwartz–Zippel batching/soundness path unchanged. Please verify and resolve if this matches the intended decomposition.

alexanderlhicks:
The state-aware sequential-composition consolidation is now open at https://github.com/Verified-zkEVM/ArkLib/pull/885. Its proved interfaces and nine Hachi migrations passed independent review and full validation.

We are now preparing a separate caller/API-retirement PR covering the current-main FRIBinius and BinaryBasefold completeness callers, alongside Sumcheck and Packing. The intended migration uses explicit guarded verifier forms, pure prover outputs for the empty ambient oracle, and suffix completeness specialized at the actual shared state. Exact elaborated caller types are captured to check that current public assumptions are preserved.

This does not resolve this PR's independent cast, RBR/error-transport, lifting, or general security dependencies. I will link the validated migration when it is ready; your downstream contributions remain credited to this PR.

alexanderlhicks:
The current-main FRIBinius and BinaryBasefold completeness migration is complete in https://github.com/Verified-zkEVM/ArkLib/pull/887, stacked on https://github.com/Verified-zkEVM/ArkLib/pull/885.

All maintained caller types match their original elaborated types exactly. Independent compiled auditing finds no remaining dependency on the eight retired false completeness contracts, and full validation passed. Existing independent component/lens admissions remain disclosed; this is not a claim that every top-level Binius security theorem is clean.

This PR's additional cast and RBR/error-transport work remains separate. The current source changes and preserved public assumptions provide the concrete completeness migration to reuse when rebasing.

chung-thai-nguyen:
Update on the PR:
1. We keep the strict relation separate for completeness; this is likely unavoidable.
2. DP24’s ring-switching and FRI-Binius require laws beyond those in the default ring-switching profile, so we use a CoordinateLaws precondition for now. We may be able to remove it if PR #615 can generalize them.

PR #469, opened 2026-04-19
feat(dsfs): statements of Section 5

chung-thai-nguyen:
## Summary

This PR formalizes the Section 5 definitions of the duplex-sponge Fiat-Shamir transformation. It expands the current  specifications. It provides the executable constructions, trace machinery, bad-event infrastructure, and hybrid experiment definitions needed by the later security proofs.

- Define reusable oracle-distribution sampling and the single-salt Fiat--Shamir construction.
- Formalize the Section 5 protocol, codec, and nondegeneracy interfaces.
- Define the trace data structures and implement lookahead, backtracking, and candidate extraction.
- Implement the prover and trace transformations used by the duplex-sponge-to-single-salt reduction.
- Define the Section 5 bad events and prove the structural exclusion results corresponding to Lemmas 5.10, 5.12, 5.14, and 5.16.
- Define the endpoint experiments and hybrid chain `Hyb_0` through `Hyb_4`.
- Add `θStar`, `ηStar`, the per-transition bound expressions, and the required query-bound interfaces.
- Expose the new modules through ArkLib and update the axiom baseline and repository map.

## Authors

- Chung Thai Nguyen
- Michele Orrù
- Yuxi Zheng

## Proof boundary

This PR intentionally provides the Section 5 definitions and supporting infrastructure without proving the probabilistic hybrid claims, Lemma 5.8, or Lemma 5.1. Abort Analysis lemmas, Claims 5.21–5.24 and the final Key Lemma proof remain deferred to a later proof-focused PR.

The single-salt Fiat--Shamir completeness theorem is stated here, while its proof is also left outside the scope of this PR.

## Follow-up PR

#848 is stacked directly on this PR. It adds the NARG and state-restoration security definitions, proves canonical Fiat--Shamir soundness and straightline knowledge soundness, and derives the DSFS results corresponding to Theorems 6.1 and 6.2.

## Verification

- Ran `./scripts/validate.sh --axioms --docs` successfully using Python 3.12.
- Built the library and API documentation successfully.
- Verified that the refactored commit sequence preserves the final implementation and authorship attribution.


alexanderlhicks (commented):
## Independent review of §5/§6 vs CO25 (ePrint 2025/536)

Thanks for this substantial formalization — the transform/codec plumbing and the §5.6 bad-event predicates are largely faithful, and the file is `axiom`-free. The comments below focus on **sorried statements that are false or unfaithful as written** (so that whoever closes them isn't misled), validated against the paper and existing ArkLib abstractions. Highlights:

- **Quantitative (HIGH):** the state-restoration soundness/KS hypotheses of Thms 6.1/6.2 are un-parametrized (`∀` prover, bare `ε`, no `θ★(t)`/`n`), so they cannot carry CO25's bound — the θ★ query bound needed is already proven in `lemma_5_1.2` but discarded. Root cause is the un-parametrized SR notion in `StateRestoration.lean` (pre-existing on `main`).
- **False as written (HIGH):** `claim_5_19`, `claim_5_20`, `lemma_5_18` — each has an explicit counterexample in the inline comments; `theorem_5_19` is a *proven* `theorem` that is nonetheless a false Prop (rests on the false `lemma_5_18`). These §5.7 results currently have **no consumers**, so nothing is unsound today.
- **Definitions/constants (MED):** `E_func` is strictly larger than paper Eq 27 (watch `lemma_5_8`); `claim5_24Bound` is tighter than Claim 5.24.
- **Nits:** a systematic −1 equation-citation shift; a `<j` vs `≤j` docstring typo.

Full write-up with fix ordering is available separately. Flagged as comments (not a formal request-changes) — all are actionable for closing the sorries faithfully.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/Soundness.lean
```diff
@@ -1,16 +1,1212 @@
...
+            (Option.map (fun out : DSFSGameOutput (oSpec := oSpec) (StmtIn := StmtIn)
+              (StmtOut := StmtOut) (pSpec := pSpec) (U := U) (δ := δ) => (out.1, out.2.1))) |
+          dsfsGameDist hyb0Init (hyb0Impl oSpecImpl) V maliciousProver ] := by rw [probEvent_map]
+    _ = Pr[ dsfsRawEvent langIn langOut |
+          dsfsGameDist hyb0Init (hyb0Impl oSpecImpl) V maliciousProver ] := by rw [hev2]
+
+/-- **Theorem 6.1** — Soundness of the duplex-sponge Fiat–Shamir scheme.
+For a query-bounded malicious prover, its false-acceptance probability `ε_NARG`
+is at most `ε_sr + η★`. -/
+theorem theorem_6_1_soundness
```
alexanderlhicks, line 730: **[HIGH] SR-soundness hypothesis carries no quantitative content.**
The `h_IP_SR_sound : …soundnessWithCoins … ε_sr` hypothesis quantifies over **all** SR provers with a bare `ε_sr : ENNReal` — no move budget and no instance size. CO25 Thm 6.1 (+Eq 58) only needs SR-soundness against a **`θ★(t)=t`-query prover at size `n`**: `εNARG(λ,t,n) ≤ ε_sr_IP(δ★, θ★(t), n) + η★`. Because the SR challenge oracle is a fixed pre-sampled random function (`srInitDIP`), an *unbounded* prover can grind across prefixes, so the only valid `ε_sr` over the full `∀` is `≈1` for any IP with soundness error in `(0,1)` ⇒ `ε_sr + η★ ≥ 1` is trivial. The theorem is thus non-trivially instantiable only for perfectly-sound IPs.
Note the fix is already in reach: `lemma_5_1.2` proves `IsTotalQueryBound (d2sAlgoTransform …) (θStar …)` for the induced prover, and `hyb4_falseAccept_le_nargSoundness` instantiates the hypothesis at exactly that one bounded prover — but line 768 takes only `.1` and discards the bound. Parametrize the SR notion by `(t,n)` (Def 3.13) and instantiate at `θ★(t)/n`; the existing proof should go through. (`η★` itself is faithful.)

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/Soundness.lean
```diff
@@ -1,16 +1,1212 @@
...
+    simp [OptionT.run_mk, simulateQ_map, map_bind, bind_map_left,
+      bind_assoc]
+  rw [hdist]
+
+-- Needed because unfolding `adaptiveNARGKnowledgeSoundnessExp` and `dsfsKSGameDist` generates large
+-- proof states
+/-- **Theorem 6.2** — Straightline knowledge soundness of the duplex-sponge Fiat–Shamir scheme.
+For a query-bounded malicious prover, the extraction-failure probability
+is at most `ε_sr + η★`. -/
+theorem theorem_6_2_straightline
```
alexanderlhicks, line 1100: **[HIGH] Same one-sided defect on the SR-KS side.**
`knowledgeSoundnessWithCoins … ε_sr` has no `θ★(t)`/`n` parametrization, so it cannot be instantiated with CO25's `κ_IP^sr(δ★, θ★(t), n)` (Thm 6.2, straightline case). Extractor wiring (Construction 6.3 via `theorem_3_19`) and the `η★` term look right; the gap is purely the query/instance-bound quantification of the hypothesis — same fix as `theorem_6_1_soundness`.

ArkLib/OracleReduction/Security/StateRestoration.lean
```diff
@@ -140,6 +186,40 @@ def soundness
...
+      let ⟨transcript, stmtIn⟩ ← srSoundnessGameWithCoins srProver
+      let stmtOut ← liftComp (verifier.run stmtIn transcript) _
+      return (stmtIn, stmtOut))).run' (← init)
+  ]
+
+/-- **Coin-bearing** state-restoration soundness: identical to `soundness`, but the prover may use
+private coins `auxSpec` (answered at game time by the sampler `auxImpl`). The challenge oracle is
+still answered by `srChallengeQueryImpl'` (the pre-sampled function in `init`), the IP's shared
+oracle by `impl`. Taking `auxSpec := []ₒ` recovers `soundness`. -/
+def soundnessWithCoins {κ : Type} (auxSpec : OracleSpec κ)
```
alexanderlhicks, line 214: **[HIGH — root cause of the two §6 theorems] SR error is un-parametrized.**
`soundnessWithCoins` (and the base `soundness` @176, which pre-exists on `main`) uses a single `srSoundnessError : ENNReal` with `∀ srProver` and no move-budget `t` / size `n`, unlike CO25 Def 3.12/3.13 (a `t`-move malicious prover at bound `n`, error `ε_sr_IP(δ,t,n)`). The new coin variant faithfully mirrors the existing shape — but it is what hollows out Thms 6.1/6.2. Suggest restricting the game prover to `t` moves and making `srSoundnessError` a function of `(δ,t,n)`.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/AbortAnalysis.lean
```diff
@@ -11,12 +11,301 @@ import ArkLib.OracleReduction.FiatShamir.DuplexSponge.Security.BadEvents
...
+    (state : CanonicalSpongeState U)
+    (S_BT : S_BT trace state)
+    (hInv : ¬ BadEventDS.E_inv trace state S_BT)
+    (hPrp : ¬ BadEventDS.E_prp trace)
+    (hFork : ¬ BadEventDS.E_fork trace state S_BT) :
+    BackTrackNoAbort (δ := δ)
+      (T_H := T_H) (T_P := T_P) (StmtIn := StmtIn) (n := n) (pSpec := pSpec) (U := U)
+      (depthBound := trace.length + 1) (trace := trace) (trΔ := trΔ) (h_trΔ := h_trΔ)
+      (state := state) := by
+  sorry
```
alexanderlhicks, line 160: **[HIGH] `claim_5_19` is false as written.**
`BacktrackSequenceFamily` (Backtrack.lean:296) carries only `seqFamily` + a `maximality` antichain closure — it does **not** encode CO25 Def 5.3's completeness (`S_BT(tr,s)` = the set of *all* maximal valid sequences, which also excludes self-loops via `sC,in,ι ≠ sC,out,ι`). So `⟨∅, by simp⟩` is a legal witness, making `E_fork := seqFamily.card > 1` and `E_inv` (over `J_BT ∅ = ∅`) vacuously false. Meanwhile the executable `backTrack` (which never sees the family) hits `forkErr` on a self-loop p-entry (Backtrack.lean:653).
**Counterexample:** `trace = [(‘p’, s_in, s_out)]` with `s_in.cap = s_out.cap = state.cap`; `trΔ.p` that entry, `trΔ.h = ∅`; `S_BT = ⟨∅, by simp⟩`. All of `hInv/hPrp/hFork` hold, yet `backTrack … = err`. As Backtrack.lean:1002 notes, the scan-time fork is *"a strict over-approximation of E_fork,p ∪ E_fork,h,p"* — that breaks the `¬E ⇒ no abort` direction. Fix: encode canonicity/completeness of `S_BT` and reconcile the self-loop err with `E_inv`.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/AbortAnalysis.lean
```diff
@@ -11,12 +11,301 @@ import ArkLib.OracleReduction.FiatShamir.DuplexSponge.Security.BadEvents
...
+    (trace : QueryLog (duplexSpongeChallengeOracle StmtIn U))
+    (trΔ : TraceNabla T_H T_P StmtIn U)
+    (state : CanonicalSpongeState U)
+    (i : pSpec.ChallengeIdx)
+    (hPrp : ¬ BadEventDS.E_prp trace) :
+    LookAheadNoAbort
+      (T_H := T_H) (T_P := T_P)
+      (StmtIn := StmtIn) (pSpec := pSpec) (U := U)
+      trΔ state i := by
+  sorry
```
alexanderlhicks, line 178: **[HIGH] `claim_5_20` is false as written — missing the `trΔ`↔`trace` binding.**
`claim_5_19` (@150) carries `h_trΔ : trΔ.IsSubsetOfQueryLog trace`; this lemma does not. Its conclusion `LookAheadNoAbort … trΔ state i` depends only on `trΔ.p`, so `trace`/`hPrp` are dead and `trΔ.p` is a free variable. CO25 Claim 5.20 explicitly says *"let `tr∇` be **its corresponding** data structure"* — that is the dropped binding.
**Counterexample:** `trace = []` (so `¬E_prp []` holds), `trΔ.p = {(state,a),(state,b)}` with `a≠b` ⇒ `successorCandidates` returns two entries ⇒ `forkErr` ⇒ `err ∈ support`. Fix: add `h_trΔ` (as in `claim_5_19`) and reconcile the self-loop err mode.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/AbortAnalysis.lean
```diff
@@ -11,12 +11,301 @@ import ArkLib.OracleReduction.FiatShamir.DuplexSponge.Security.BadEvents
...
+    (initM : M)
+    (tr_A : QueryLog (duplexSpongeChallengeOracle StmtIn U))
+    (h_tr_A_mem_support : some tr_A ∈ support (duplexSpongeTrace (δ := δ) (T_H := T_H) (T_P := T_P)
+        gImpl A initM).run)
+    (hE : ¬ BadEventDS.E tr_A) :
+    D2SQueryNoAbort (δ := δ) (T_H := T_H) (T_P := T_P) (StmtIn := StmtIn) (n := n)
+      (pSpec := pSpec) (U := U) gImpl A initM
+    -- This also means tr_A is the full trace that we can collect from
+      -- the non-aborted D2SQuery-simulated computation A
+    := by sorry
```
alexanderlhicks, line 258: **[HIGH] `lemma_5_18` is a quantifier/scope inversion — and it makes the *proven* `theorem_5_19` (@266) a false Prop.**
`D2SQueryNoAbort … := none ∉ support (d2fRaw …).run` is a **whole-distribution** property (@113). The hypothesis `h_tr_A_mem_support` only fixes **one** support trace `tr_A` with `¬E tr_A`. One clean path does not imply the whole run never aborts. CO25 Lemma 5.18 is **per-execution** (*"let `tr_A` be the trace resulting from **the** execution … if `E(tr_A)=0` then `A^{D2SQuery}` does not abort"*).
**Counterexample:** a probabilistic `A` with a clean path (trace `tr1`, `¬E tr1`) and an aborting path (`d2fRaw ↦ none`); take `tr_A := tr1`. Both hypotheses hold, conclusion `none ∉ support` is false — the normal regime, since the analysis bounds `Pr[E] ∈ (0,1)`.
`theorem_5_19_d2sQuery_abort_implies_badEvent` has no `sorry` in its body but is proven from this lemma, and the same counterexample falsifies its statement — a false Prop a `sorry`-grep would miss. Fix: state Lemma 5.18 per-path and re-derive 5.19.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/AbortAnalysis.lean
```diff
@@ -11,12 +11,301 @@ import ArkLib.OracleReduction.FiatShamir.DuplexSponge.Security.BadEvents
...
+    [∀ i, Fintype (pSpec.Message i)]
+    {T_H T_P : Type} {Salt : Type} [SaltCodec U δ Salt]
+    [LawfulTraceNablaImpl T_H T_P StmtIn U]
+    (log : TaggedQueryLog (oSpec + duplexSpongeChallengeOracle StmtIn U))
+    (hE : ¬ BadEventDS.E (dsTraceOfLog (oSpec := oSpec) (StmtIn := StmtIn) (U := U)
+      (TaggedQueryLog.untagged log))) :
+    D2STraceNoAbort (T_H := T_H) (T_P := T_P) (δ := δ) (Salt := Salt)
+      (oSpec := oSpec) (StmtIn := StmtIn) (pSpec := pSpec) (U := U)
+      log := by
+  sorry
```
alexanderlhicks, line 201: **[MED] `lemma_5_17` statement is faithful, but its advertised proof route is blocked.**
The statement matches CO25 Lemma 5.17 and is *plausibly true* — note that `¬E` **does** exclude a self-loop forward p-step, via the `≤ j` self-branch of `isDuplicatedPriorCapacity` (BadEvents.lean:405) which fires `E_p` (⇒ `E_dup` ⇒ `E`). However the docstring's plan is "apply Claim 5.19 / Claim 5.20", both of which are false as stated (see comments above), so that route is unavailable. (Minor: docstring says `D2STrace`; the paper procedure here is `StdTrace`.)

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
```diff
@@ -1,117 +1,996 @@
...
+/-- CO25 Claim 5.22. Statistical-distance bound for `Hyb_1` vs `Hyb_2` (Eq. 53):
+`θ★(t_h, t_p, t_{p⁻¹}) · max_i ε_{cdc,i} + ∑_i ε_{cdc,i}`. -/
+def claim5_22Bound
+    (tₕ tₚ tₚᵢ : ℕ) (εcodec : CodecBias (pSpec := pSpec)) : ℝ :=
+  (θStar tₕ tₚ tₚᵢ : ℝ) * iSup (fun i => (εcodec i : ℝ))
+    + ∑ i, (εcodec i : ℝ)
+
+/-- CO25 Claim 5.24. Statistical-distance bound for `Hyb_3` vs `Hyb_4` (Eq. 55):
+`(7·L·(2·t_h + 2 + 2·t_p + L + 2·t_{p⁻¹})) / (2·|Σ|^c) − 5·(L+1) / |Σ|^c`. -/
+def claim5_24Bound (U : Type) [SpongeUnit U] [Fintype U]
```
alexanderlhicks, line 582: **[MED] `claim5_24Bound` constant deviates from CO25 Claim 5.24 (and is strictly tighter).**
Paper (p.51): `7(L+1)(2tₕ + 1 + 2tₚ + L + 2tₚ⁻¹)/(2|Σ|^c) − 5(L+1)/|Σ|^c`. Here: `7·L·(2tₕ + 2 + 2tₚ + L + 2tₚ⁻¹)/(2·cardPow) − 5(L+1)/cardPow` — two deviations (`L` vs `L+1`, `+2` vs `+1`). `paper − lean = 7(2tₕ+2tₚ+2tₚ⁻¹+1) > 0`, so this asserts a value **strictly tighter** than the paper proves ⇒ likely unprovable by the paper's argument. Currently harmless (`claim_5_24`/`claim5_24Bound` have zero downstream consumers; `ηStar` is a separate hardcoded formula), but please fix the constant. Also the `lemma_5_1_inner` docstring (@909) overstates that the `claim_5_21..24` triangle is wired into it — it is `sorry` and invokes none of the claims.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/BadEvents.lean
```diff
@@ -1,343 +1,1660 @@
...
+```
+
+Note: `E_func(tr)` never holds for a true permutation `p` and its inverse `p⁻¹`, but may hold
+(with small probability) for the D2SQuery simulator.
+
+**Strengthening:** bidirectional. Case 1 (`j`-th entry `p`-forward) is Eq. 26; Case 2 (`j`-th entry
+`p⁻¹`) has no paper counterpart but is *required* by `not_collisionFwdBwd_of_not_combined`
+(Lemma 5.10, Item 3). The `≠`-output conditions are forced by base-trace non-redundancy.
+See `DSFS-archive/BadEvents_deep_analysis.md` §7. -/
+def E_func : Prop :=
```
alexanderlhicks, line 502: **[MED, new] `E_func` (hence `E`) is defined strictly larger than paper Eq 27.**
Paper Eq 27 anchors only on a forward `p` entry; Case 2 here (anchor `p⁻¹`, lines 511–516) has, per the docstring, *"no paper counterpart"* (added to prove `not_collisionFwdBwd_of_not_combined`). This is a defensible strengthening — it only *strengthens* the `¬E ⇒ …` lemmas (5.10/5.12/5.14/5.16). **But** `E = E_dup ∨ E_func` is now a superset of the paper's `E`, and the still-open `lemma_5_8` claims the paper's `Pr[E]` bound on the enlarged event, whereas CO25's Eq-34 counting covers only the forward direction. Before discharging `lemma_5_8`, confirm the same constant holds on the bidirectional event (plausible by permutation symmetry, but not derived in the paper) and document the deviation.

ArkLib/OracleReduction/FiatShamir/SingleSalt.lean
```diff
@@ -0,0 +1,303 @@
...
+
+If `saltedIPVerifier V` has state-restoration knowledge soundness (for `relInSalted relIn`)
+with error `ε`, then `Verifier.singleSaltFiatShamir V` has straightline KS in the ROM with
+error `ε`.
+
+**Proof sketch**: given the SR extractor `E_SR` from `h_sr_ks`, the FS straightline extractor
+(1) extracts `(τ, messages)` from the FS proof, (2) filters the FS query log to reconstruct
+the IP SR query log, and (3) calls `E_SR (stmtIn, τ) witOut ip_transcript ip_log`.
+**Open seam**: body is `sorry` — used as Seam #2 in `Soundness.lean` (Theorem 6.2). -/
+theorem theorem_3_19_straightline_ks
```
alexanderlhicks, line 277: **[MED] `theorem_3_19` drops the extraction-time guarantee; extractor oracle-access is over-general.**
CO25 Thm 3.19 asserts both a KS-error bound *and* `etNARG ≤ etsr + O(r_max·t)`. The ArkLib KS notion (`adaptiveNARGKnowledgeSoundnessWithCoins`) models only failure-probability, so the extraction-time half is silently dropped — soundness-safe, but the name/docstring shouldn't be read as delivering the efficiency claim. Separately, the SR-hypothesis extractor runs over live `oSpec` while the conclusion extractor runs over an arbitrary `auxSpecE` (possibly `[]ₒ`) from static logs; for nontrivial `oSpec` the conclusion looks stronger than the hypothesis supports, so a proof would need to lean on the DSFS-specific straightline instantiation.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Defs.lean
```diff
@@ -1,83 +1,494 @@
...
+        (gSpec (U := U) StmtIn pSpec δ))] :
+    OracleReduction.OracleDistribution
+      (gSpec (U := U) StmtIn pSpec δ) :=
+  OracleReduction.D_ROM _
+
+/-- Bridge: `SampleableType` for `gSpec` (Hyb₁ `g`) derived from
+granular `VCVCompatible` base-type hypotheses. Eliminates verbose `SampleableType (OracleFamily
+(gSpec …))` at call sites in §5.8 hybrids and in `BadEvents.lemma_5_8`'s
+eager `𝒟_Σ` sampling. -/
+instance instSampleableTypeEncodedChallengeOracle
```
alexanderlhicks, line 428: **[LOW] Sorried instance drops hypotheses its sibling carries (one of three such instances).**
This `instance … := by sorry` omits the `[VCVCompatible StmtIn] [VCVCompatible U] [∀ i, VCVCompatible (Challenge i)]` that `instSampleableTypeDecodedChallengeOracle` (@483) carries, so for infinite `StmtIn`/`U` the `SampleableType` genuinely doesn't exist and the `sorry` can't be honestly closed. There are **three** sorried `SampleableType` instances (here, @489, and `OracleDistribution.lean:331`); as global instances they propagate implicitly into `D_ROM`/`D_IP`/`𝒟_Σ`, so any downstream `#print axioms`-clean claim is false while they stand. Add the missing binders and discharge all three via the new `VCVCompatible → FinEnum → SampleableType` bridges in `Prelude.lean`.

ArkLib/OracleReduction/FiatShamir/Basic.lean
```diff
@@ -160,13 +160,14 @@ variable [∀ i, SampleableType (pSpec.Challenge i)]
 theorem fiatShamir_completeness (relIn : Set (StmtIn × WitIn)) (relOut : Set (StmtOut × WitOut))
     (completenessError : ℝ≥0) (R : Reduction oSpec StmtIn WitIn StmtOut WitOut pSpec) :
   R.completeness init impl relIn relOut completenessError →
-    R.fiatShamir.completeness (do
+    R.fiatShamir.completeness (init := do
```
alexanderlhicks, line 163: **[LOW, pre-existing — noting for completeness]** The FS challenge oracle is initialized to the constant `fun ⟨i,_⟩ => default`, so the honest run reproduces one execution at the fixed challenge vector `(default,…,default)`. `R.completeness` only bounds reject probability *averaged over uniform* challenges, so it can't imply FS completeness at that single point for general error `e` — unfaithful (the paper uses uniform-RO challenges) and unfillable as written (the `-- This is not a standard pre-sampling of` comment acknowledges it). `git diff` shows this PR only made a cosmetic named-argument change here. Needs a genuine pre-sampled random challenge init.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/BadEvents.lean
```diff
@@ -1,343 +1,1660 @@
...
+  (∃ j' < j, ∃ stateIn1 stateOut1, baseTrace[j'] = ⟨.inr <|.inl stateIn1, stateOut1⟩ ∧
+    stateOut1.capacitySegment = capSeg) ∨
+  (∃ j' < j, ∃ stateOut2 stateIn2, baseTrace[j'] = ⟨.inr <|.inr stateOut2, stateIn2⟩ ∧
+    stateIn2.capacitySegment = capSeg) ∨
+  (∃ j' ≤ j, ∃ stateIn3 stateOut3, baseTrace[j'] = ⟨.inr <|.inl stateIn3, stateOut3⟩ ∧
+    stateIn3.capacitySegment = capSeg) ∨
+  (∃ j' ≤ j, ∃ stateOut4 stateIn4, baseTrace[j'] = ⟨.inr <|.inr stateOut4, stateIn4⟩ ∧
+    stateOut4.capacitySegment = capSeg)
+
+/-- CO25 Definition 5.7 — Event `E_h(tr)` (Eq. 23).
```
alexanderlhicks, line 408: **[NIT] Equation citations are systematically off by −1 across this file.** `E_h` cites Eq 23 but is **Eq 24**; likewise `E_p` 24→**25**, `E_{p⁻¹}` 25→**26**, `E_func` 26→**27**, `E_inv` 35→**36**, `E_fork,{h,p,h·p}` 38/39/40→**39/40/41**, `E_time,{h,p}` 41/42→**42/43**. Root cause: Def 5.5's redundancy *triple* (Eqs 21/22/23) was counted as a pair. The Definition and Lemma numbers (5.5–5.18) are all correct. (The predicates themselves — `E`, `E_prp`, the `E_fork` boundaries, `E_time` — check out clause-by-clause against §5.6.)

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/BadEvents.lean
```diff
@@ -1,343 +1,1660 @@
...
+/-- CO25 Definition 5.7 — Event `E_p(tr)` (Eq. 24).
+An output capacity segment `s_C` of a `p`-entry in the base trace `tr̄` previously (or
+simultaneously for some branches) appears as an output or input capacity segment of `h`, `p`,
+or `p⁻¹`:
+
+```
+E_p(tr) := ∃ j > 0, s_C ∈ Σ^c :  tr̄_j = (p, ·, (·, s_C))  and
+  ∃ j' < j : tr̄_{j'} = (h, ·, s_C)  ∨  ∃ j' < j : tr̄_{j'} = (p, ·, (·, s_C))
+  ∨  ∃ j' < j : tr̄_{j'} = (p⁻¹, ·, (·, s_C))
+  ∨  ∃ j' ≤ j : tr̄_{j'} = (p, (·, s_C), ·)  ∨  ∃ j' < j : tr̄_{j'} = (p⁻¹, (·, s_C), ·)
```
alexanderlhicks, line 436: **[NIT]** This docstring writes arm 5 as `∃ j' < j`, but the def (`isDuplicatedPriorCapacity` arm 5) and paper Eq 25 both use `∃ j' ≤ j` — the def/paper are correct, the prose is a typo. Also the `E_p`/`E_{p⁻¹}` docstrings call the shared predicate "uniform `≤ j`", but it is asymmetric (`<j` on arms 1–3, `≤j` on arms 4–5) — which is exactly right vs the paper; only the description is off.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/Soundness.lean
chung-thai-nguyen: [DONE] Section 6 is now pushed into a later PR, the issue will be addressed there.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/Soundness.lean
chung-thai-nguyen: [DONE] Section 6 is now pushed into a later PR, the issue will be addressed there.

ArkLib/OracleReduction/Security/StateRestoration.lean
chung-thai-nguyen: [DONE] Section 6 is now pushed into a later PR, the issue will be addressed there.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/BadEvents.lean
chung-thai-nguyen: [DONE] E_func is intentionally defined symmetrically compared to the original asymmetric definition in the paper, downstream bad events lemmas (5.10 -> 5.16) got proven under this.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/AbortAnalysis.lean
chung-thai-nguyen: [DONE] Claim 5.19 and AbortAnalysis.lean changes have been **removed and deferred until the proof of Claim 5.21**, where the **statements can be specialized to its actual use**. The retained Backtrack construction now requires inputCapacities.Nodup and returns noResult for cyclic/self-loop candidates, eliminating the reported counterexample.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/AbortAnalysis.lean
chung-thai-nguyen: [DONE] Claim 5.20 has been **removed and deferred** with AbortAnalysis. Its later formulation will explicitly bind trΔ to the source trace before claiming that LookAhead cannot abort.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/AbortAnalysis.lean
chung-thai-nguyen: [DONE] Lemma 5.18 and the associated abort-analysis chain have been **removed from this PR**. (The later proof will use a **per-execution formulation** before deriving the distribution-level bad-event statement).

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/AbortAnalysis.lean
chung-thai-nguyen: [DONE] Lemma 5.17 has been **removed and deferred together with Claims 5.19–5.20**. This PR therefore no longer relies on the blocked proof route or overclaims a StdTrace/D2STrace result.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
chung-thai-nguyen: [DONE] I **fixed the bound** for Claim 5.24. It now uses the exact CO25 constant with (L+1) and +1.

ArkLib/OracleReduction/FiatShamir/SingleSalt.lean
chung-thai-nguyen: [DONE] Theorem 3.19 and the Section 6 extractor layer have been **removed from this PR**. Oracle access and any formal extraction-time guarantee will be handled explicitly in the later straightline-(K)S PR.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Defs.lean
chung-thai-nguyen: [DONE] All three relevant SampleableType instances are now **proved**: the generic FS challenge table, the encoded gSpec oracle family, and the decoded eSpec oracle family.

ArkLib/OracleReduction/FiatShamir/Basic.lean
chung-thai-nguyen: [DONE] The generic and single-salt FS completeness statements **now initialize one uniformly sampled complete challenge table**. The single-salt statement additionally requires a fresh uniform salt sampler independent of the statement, prover state, and FS table. The theorem proofs are intentionally out of scope.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/BadEvents.lean
chung-thai-nguyen: [DONE] The documentation **now reflects the actual asymmetric quantifiers**: < j for arms 1–3 and ≤ j for arms 4–5. The Lean predicate itself was already correct.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/BadEvents.lean
chung-thai-nguyen: [DONE] I shifted every CO25 equation citation from old Eq. 6 onward by +1 to match the latest version of the paper.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
```diff
@@ -1,123 +1,1080 @@
...
+        (oSpec := oSpec) (U := U) (challengeSpec := challengeSpec) oSpecImpl D_g)
+      (gImpl := gImpl) V maliciousProver
+      (hyb1Line4Trace
+        (δ := δ) (Salt := Salt)
+        (oSpec := oSpec) (StmtIn := StmtIn) (pSpec := pSpec) (U := U))
+
+/-- CO25 Claim 5.21.
+`Δ(Hyb_0, Hyb_1) ≤ (7·T² − 3·T) / (2·|Σ|^c)` with `Hyb_0 / Hyb_1` sampled eagerly via
+`hyb_0` / `hyb_1`. -/
+theorem claim_5_21
```
alexanderlhicks, line 687: [HIGH] **The statement is false as written: `(tₕ tₚ tₚᵢ : ℕ)` are free, so the claim asserts the
bound for *every* budget, including `t = 0`.**

In the paper, Claims 5.21–5.24 live *inside the proof of Lemma 5.1*, whose preamble fixes a
"(t_h, t_p, t_{p⁻¹})-query argument prover P̃"; Claim 5.21's proof concludes "by Lemma 5.8",
whose statement is "*For every (t_h, t_p, t_{p−1})-query algorithm P̃ …*". So the paper's
`t`'s are the query counts *of that prover*.

In ArkLib, `MaliciousProver` (Defs.lean:723) is a bare `OracleComp` with no intrinsic budget —
the budget is a *proposition about* the prover, not part of its type — so it must appear as a
hypothesis. This theorem has none: instantiate at `tₕ = tₚ = tₚᵢ = 0` and it asserts
`Δ(Hyb₀, Hyb₁) ≤ (7(L+1)² − 3(L+1)) / (2|Σ|^c)` — a fixed constant — for **every** prover.
A prover making `q ≫ |Σ|^c` permutation queries distinguishes the real permutation oracle
(Hyb₀) from the `D2SQuery` simulator (Hyb₁) with probability → 1 (birthday-hit an `E`-event:
real `p` never triggers `E_func`, the simulator does), so the ≤ fails.

The file itself shows the intended pattern: `lemma_5_1_inner` (:960) takes
`IsLemma5_1QueryBound maliciousProver tₕ tₚ tₚᵢ →`, and `BadEvents.lemma_5_8` (:806) takes
`hMaliciousBound : IsLemma5_8QueryBound …`. Since 5.21–5.24 are exactly the steps of
`lemma_5_1_inner`'s intended proof, the cleanest fix is to thread the *same* hypothesis:

```lean
theorem claim_5_21 … (tₕ tₚ tₚᵢ : ℕ)
    (hBound : IsLemma5_1QueryBound maliciousProver tₕ tₚ tₚᵢ) : …
```

Same omission in `claim_5_22` (:736) — at `t = 0` it claims `Δ(Hyb₁, Hyb₂) ≤ ∑ᵢ ε_cdc,i` for
unbounded provers, but the paper's `θ★·maxᵢ ε_cdc,i` term exists precisely because each of the
prover's `t_p` forward-p queries can trigger a ψ⁻¹ decode — and in `claim_5_24` (:865).
(`claim_5_23` is exempt: `Δ = 0` holds pointwise, no budget needed.)

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/Backtrack.lean
```diff
@@ -34,79 +49,1054 @@ subject to the following conditions:
...
+  -- Each entry `⟨msgIdx, v⟩` carries the encoded message `α̂_{msgIdx} ∈ Σ^{ℓ_P(msgIdx)}`.
+  let mut alphaHat_acc : List
+      (Sigma fun msgIdx : pSpec.MessageIdx => Vector U (messageSize msgIdx)) := []
+  -- Step 4(a): For every i ∈ [k] = {1, …, k}.
+  for i in challengeIdxList (pSpec := pSpec) do
+    let L_P_before := sumMessageBlocksBefore (pSpec := pSpec) i
+    let L_V_before := sumChallengeBlocksBefore (pSpec := pSpec) i
+    -- Step 4(a).i — block pointer at round `i`:
+    --   `L_ptr(i) := L_δ + Σ_{j<i} L_P(j) + Σ_{j<i} L_V(j)`  (CO25 Eq. 7/8).
+    let L_ptr := Lδ (δ := δ) + L_P_before + L_V_before
```
alexanderlhicks, line 964: [HIGH] **The block pointer is off by one round's message width: `L_ptr` here equals the paper's
`Lptr(i) + L_P(i)`, so the message window reads the challenge-squeeze blocks and the honest
exact-fit chain is always rejected.**

Paper Step 4(a)i: "*Let Lptr(i) := L_δ + Σ_{ι<i} (L_P(ι) + L_V(ι))*" — the sum ranges over
**rounds ι strictly before round i**, so `[Lptr(i), Lptr(i)+L_P(i))` is exactly α̂ᵢ's absorb
window (Eq. 12; this file's own docstring at :863 quotes it correctly:
`α̂_i := concat(s_{R,in,L_ptr(i)}, …, s_{R,in,L_ptr(i)+L_P(i)-1})`).

But `sumMessageBlocksBefore i` (:748) sums over `messageIdxListBefore i` = messages with
**flat step index `< i.1`** (:739-741). In ArkLib's `ProtocolSpec`, a paper round is two flat
steps — the round-i message sits at flat index `i.1 − 1 < i.1` — so the sum *includes the
current round's own message*, and `L_ptr = Lptr_paper(i) + L_P(i)`.

Worked example (single-round `pSpec = ⟨[.P_to_V, .V_to_P]⟩`, no salt, `L_P(1) = L_V(1) = 1`):
take the honest exact-fit candidate chain `inputState = [s_in,0]` (`m_k + 1 = 1`, the chain
ending right after α₁ is absorbed).
- Paper: `Lptr(1) = 0`; guard `0 + 1 > 1` is false; window `[0,1)` = α̂₁ ✓; Step D exact-fit
  `0 + 1 = 1` fires → tuple stored. ✓
- Lean: `L_P_before = L_P(1) = 1` (message at flat 0 < 1), so `L_ptr = 1`; the Step 4(a)ii
  guard (:972) reads `1 + 1 > 1` → **`return none`**. ✗

Since the guard reads `L_ptr + L_P_i > m_k+1` with `L_ptr` already past the message, *every*
honest boundary chain with `L_P_i > 0` is rejected. Downstream, `d2sHandleForwardPermQuery`
then always takes the lazy-sampling branch, `d2sAlgo` never queries `f`, and the sorried
`lemma_5_1_inner` / `lemma_5_1_inner_tvBound` — which pin these *concrete* witnesses — are
false as written: in Hyb₄ the honest verifier's challenges come from `f` while an `f`-blind
`d2sAlgo(P̃)`'s transcript can't track them, so `Δ(Hyb₀, Hyb₄) ≈ 1 ≫ η★`. CI stays green only
because no compiled theorem exercises `extractCandidate`.

Fix: make the sum range over strictly-earlier *rounds*, e.g. filter messages by
`j.1 < i.1 − 1` — or, more robustly for non-alternating specs, exclude
`lastMessageBefore? i` from the sum (it is exactly the round-i message this loop is trying to
assemble). `assembleEncodedMessage`, the Step-D exact-fit test, `msgEndBlock`, and
`checkSqueezeWindow` all consume `L_ptr`, so a single fix at :964 repairs all of Step 4(a).
A cheap sanity theorem would have caught this and is worth adding alongside the fix:
*for an honest single-round trace, `backTrack` returns `some` with `encodedMessages = α̂₁`.*

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/TraceDataStructures.lean
```diff
@@ -0,0 +1,883 @@
...
+The `inlu`/`outlu` laws state that a lookup succeeds iff the entry exists exactly once and is the
+unique value/key match in the multiset; duplicate-entry traces are treated as multiple matches. -/
+class LawfulTraceTable (T : Type) (K V : outParam Type) [DecidableEq K] [DecidableEq V]
+extends TraceTableOps T K V where
+  toMultiSet : T → Multiset (K × V)
+  toMultiSet_empty : toMultiSet TraceTableOps.empty = (0 : Multiset (K × V)) := by simp [empty]
+  toMultiSet_add : ∀ t k v, toMultiSet (add t k v) = (k, v) ::ₘ toMultiSet t
+  -- **inlu's query result MUST BE UNIQUE**, i.e. two copies
+    -- of `(k, v)` in the multiset trigger the "multiple" case
+  inlu_eq_some : ∀ t k v,
```
alexanderlhicks, line 124: [HIGH] **The `count (k,v) = 1` law makes lookups fail on *identical duplicate* entries — but the
combined trace `tr_P̃ ∥ tr_V` always contains identical duplicates, so `d2sTraceSalted`
aborts on essentially every real execution.**

Where the duplicates provably come from: the honest verifier re-runs the prover's sponge
chain (`deriveTranscriptDSFSSalted` replays the same absorb/squeeze `p`-queries), and
`D2SQuery` Item 4(f) appends *every* forward-p query to the trace
(ProverTransform.lean:435-448 — correctly, per the paper). `TraceNabla.ofQueryLogForwardOnly`
(TraceTransform.lean:347-352) then inserts the full log into `tr_∇`, again matching the
paper's StdTrace Step 3 ("*For each entry in tr of the form ('p', s_in, s_out), run
tr_∇.p.add(s_in, s_out)*"). So `tr_∇.p` holds two copies of each honest-chain pair, this law
forces `inlu = none`, `classifyLookup` reports `.conflict`, and BackTrack/LookAhead return
`forkErr` where the paper's StdTrace succeeds.

On the paper side, Def 5.2's "⊥ if there are zero or multiple matches" is ambiguous about
*identical* copies — but only the duplicate-tolerant reading is consistent with the rest of
§5: the normative objects are the **set-valued** families `S_BT(tr,s)` (Def 5.3) and `S_LA`
(Eq. 14), and the bad events are defined on the **base trace** (Defs. 5.5–5.7, redundant
entries removed) precisely so that benign re-queries don't count as collisions. This file's
own `E` (BadEvents.lean) is base-trace-faithful — so under the current law, "¬E ⇒ StdTrace
does not abort" (Lemma 5.17, deferred) is false for this implementation, and the sorried
`claim_5_21` (KeyLemma.lean:662), which pins Hyb₀'s line-4 map to the concrete
`d2sTraceSalted`, inherits the problem: Hyb₀'s trace collapses to `[]` on the abort path
while Hyb₁'s does not, inflating the distance past the bound.

Two equivalent fixes, either is fine:
1. **Dedupe on insert** in `ofQueryLogForwardOnly` (insert `(k,v)` only if not already
   present). Sound because `D2SQuery` answers repeated queries consistently (Items 3/4(c)i),
   so the multiset and its dedup carry the same functional content.
2. **Weaken the law** to key-uniqueness only: `inlu t k = some v ↔ (k, v) ∈ toMultiSet t ∧
   ∀ v', (k, v') ∈ toMultiSet t → v' = v` (drop the `count = 1` conjunct; conflicting-*value*
   matches still fail, which is the paper's real "multiple matches" case). Mirror for `outlu`.

Option 2 keeps the table implementation honest about what it stores; option 1 keeps the law
strict. Either way, a regression theorem "*`d2sTraceSalted` does not abort on the honest
trace*" (same shape as the sanity check suggested on `extractCandidate`) would pin the fix.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/Lookahead.lean
```diff
@@ -1,133 +1,477 @@
...
+  | 0 => .done none
+  | fuel' + 1 =>
+    -- Look up successor in `tr_∇.p` (CO25 §5.3)
+    let succs := successorCandidates (T_P := T_P) (U := U) trΔp current
+    match hSuccs : succs with
+    | [] => .done none -- No successor, sequence ends
+    | [next] =>
+        -- Found unique successor (CO25 §5.3 maximal sequence)
+        if hNoLoop : current.capacitySegment = next.capacitySegment then
+          .forkErr -- Self-loop → `E_inv`
```
alexanderlhicks: [MED-HIGH] **A self-loop successor is turned into `err`; in the paper it merely terminates the chain.**
This is the LookAhead half of the July self-loop finding — the "reject cyclic backtrack
candidates" commit fixed BackTrack (`.noResult`, Backtrack.lean:705-710) but not this branch.

Paper LookAhead Step 1(d): "*∀ ι ∈ [0, m_k−1], s_{C,in,ι} ≠ s_{C,out,ι} (no "loops" across
query and answer capacity segments)*" — this is a *membership condition on chains in `S_LA`*,
not an error trigger. A loop pair simply cannot extend any chain, so the maximal chain ends
just before it; Step 2 then returns `err` **only** when `|S_LA| > 1` (2(a)), and otherwise
`none` (2(b)) or the padded `ρ̂ᵢ` (2(c)). So on a trace whose only anomaly is one self-loop
p-entry, paper-LookAhead returns a value; Lean returns `.forkErr`.

Why it matters even with AbortAnalysis deferred: the paper's Claim 5.20 assumes only
`E_prp(tr) = 0`, and a single self-loop entry is *not* an `E_prp` collision (it's an `E_p`
event via the `j' ≤ j` arm, i.e. part of full `E`). When Claim 5.20 comes back in the
AbortAnalysis PR, it will be false for this implementation exactly the way the July
`claim_5_19` was — better to align the algorithm now than to re-discover it then.

Fix: on `hNoLoop`, end the scan (`.done` with the chain built so far — the loop pair fails
condition 1(d), so the chain-so-far is the maximal candidate) instead of `.forkErr`.
Also the inline comment `-- Self-loop → E_inv` mislabels: `E_inv` (Def 5.11 / Eq. 36) is
about p⁻¹-constructed sequences; a forward self-loop is an `E_p` case.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
```diff
@@ -1,123 +1,1080 @@
...
+/-- CO25 Claim 5.22. Statistical-distance bound for `Hyb_1` vs `Hyb_2` (Eq. 54):
+`θ★(t_h, t_p, t_{p⁻¹}) · max_i ε_{cdc,i} + ∑_i ε_{cdc,i}`. -/
+def claim5_22Bound
+    (tₕ tₚ tₚᵢ : ℕ) (εcodec : CodecBias (pSpec := pSpec)) : ℝ :=
+  (θStar tₕ tₚ tₚᵢ : ℝ) * iSup (fun i => (εcodec i : ℝ))
+    + ∑ i, (εcodec i : ℝ)
+
+/-- CO25 Claim 5.24. Statistical-distance bound for `Hyb_3` vs `Hyb_4` (Eq. 56):
+`(7·(L+1)·(2·t_h + 1 + 2·t_p + L + 2·t_{p⁻¹})) / (2·|Σ|^c) − 5·(L+1) / |Σ|^c`. -/
+def claim5_24Bound (U : Type) [SpongeUnit U] [Fintype U]
```
alexanderlhicks, line 564: [MED] The bound is negative when `pSpec.totalNumPermQueries = 0`: at `L = 0, t = 0` it evaluates to
`7/(2|Σ|^c) − 5/|Σ|^c = −3/(2|Σ|^c) < 0`, while `tvDist ≥ 0` — so `claim_5_24` is
unsatisfiable for any `pSpec` with all message/challenge sizes 0 (Lean permits these; the
paper implicitly has `k ≥ 1` and `ℓ_V(i) ≥ 1`, hence `L ≥ 1`, where the bound is
nonnegative for all `t`). Adding the query-bound hypothesis from the `claim_5_21` comment
does not rescue this (a 0-query prover always exists). Suggest a `[NeZero pSpec.totalNumPermQueries]`
style hypothesis (or `1 ≤ L`) on `claim_5_24` — or document that the degenerate case is
knowingly excluded when the claim is consumed. `ηStar` is unaffected (its numerator is
positive for all `L, t`), so only this one claim needs the guard.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/BadEvents.lean
```diff
@@ -1,343 +1,1668 @@
...
+```
+
+where `T = tₕ + 1 + tₚ + L + tₚᵢ`. The experiment distributions match CO25 Lemma 5.8, but
+`E` uses the documented bidirectional strengthening of the paper's one-sided `E_func`; this
+statement intentionally retains the same closed-form bound for that stronger event. The left-hand
+side samples `(h, p, p⁻¹) ← 𝒟_𝔖(λ, n)` once at the start of the experiment (eager sampling,
+CO25 Def. 4.2) and corresponds to `KeyLemma.dsfsGame` (`Hyb_0`); the right-hand side runs
+`g ← 𝒟_Σ(λ, n)` via the `D2SQuery` simulator and corresponds to `KeyLemma.hybridGame`
+instantiated as `Hyb_1`. -/
+theorem lemma_5_8
```
alexanderlhicks: [MED] Two gaps between this statement and paper Lemma 5.8, both on a sorried goal:

1. **The paper's side condition is dropped.** Paper: "*For every (t_h, t_p, t_{p−1})-query
   algorithm P̃ **such that t_p ≥ L***". The Lean statement has `IsLemma5_8QueryBound` but
   nothing relating `tₚ` to `L = pSpec.totalNumPermQueries`, making it strictly stronger than
   the paper's. (I could not find where the paper's proof consumes `t_p ≥ L` — the birthday
   counting doesn't seem to — but "stronger than the source, without a note" is exactly what
   the `E_func` docstring now does right; either add the hypothesis or add the same style of
   disclosure.)
2. **Salt absorption is unaccounted.** The experiment's verifier is `V.toDSFS δ`, whose
   transcript derivation absorbs the salt (`deriveTranscriptDSFSSalted`, Defs.lean:926-932)
   — that is `Lδ = ⌈δ/R⌉` extra forward-p queries when `δ > 0`. The bound's
   `T = tₕ + 1 + tₚ + L + tₚᵢ` credits the verifier with only `L = Lₚ + Lᵥ` (Eq. 8), which
   matches the paper's *unsalted* accounting ("tr_V contains exactly 1 query to h, L_P + L_V
   queries to p"); §5.6's salted verifier does more. So the base trace of this statement's own
   experiment can be `Lδ` entries longer than `T`, and the birthday bound over `T` is at risk
   for `δ > 0`. The formalization has made a paper-level ambiguity concrete — suggest either
   using `L + Lδ` in the bound here, adding a `δ = 0`/`Lδ = 0` hypothesis, or a docstring
   note justifying why the salt queries can't contribute to `E`.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
```diff
@@ -1,123 +1,1080 @@
...
+
+Sampling shape (CO25 Def. 4.2 / Eqs. 15/52/54/4): both sides draw their oracles
+from `OracleDistribution` carriers at game start; the ambient `oSpec` is answered by
+the caller-supplied handler `oSpecImpl`. Left: `oSpecImpl` plus
+`𝒟_𝔖(λ,n) = duplexSpongeOracleDistribution` for `(h, p, p⁻¹)`. Right: `oSpecImpl` plus
+salted `𝒟_IP(λ,n) = D_IP_salted` for `f`.
+
+Re-packages the concrete `lemma_5_1_inner` (the witnesses are `ProverTransform.d2sAlgo` and
+`d2sTraceSalted`) as the existential consumed by `theorem_6_1_soundness`. -/
+theorem lemma_5_1
```
alexanderlhicks, line 1048: [MED] Paper Lemma 5.1 carries the side condition Eq. (6):
`t_p ≥ max{ Σᵢ ⌈ℓ_P(i)/r⌉, Σᵢ ⌈ℓ_V(i)/r⌉ }` (i.e. `tₚ ≥ max Lₚ Lᵥ` in this file's
vocabulary), attached to the same sentence that defines `θ★ := t_p`. Neither `lemma_5_1` nor
`lemma_5_1_inner*` states it, so the Lean lemma is strictly stronger than the paper's.
Plausibly survivable — §5.4 uses the analogous condition only for the *time*-complexity
bound (p. 35), which ArkLib doesn't model — but that's an argument that belongs in a
docstring disclosure (the `E_func` precedent), or just add
`(hTp : max pSpec.Lₚ pSpec.Lᵥ ≤ tₚ)` (the aliases of `totalNumPermQueriesMessage` /
`totalNumPermQueriesChallenge`, Defs.lean:314/327) and stay paper-exact. Also worth a docstring note either way: the Lean quantifier order
(`tₕ tₚ tₚᵢ` bound before `∃ d2sAlgoTransform d2sTraceTransform`) permits budget-dependent
witnesses, weaker than the paper's outermost ∃ — harmless since the supplied witnesses are
budget-independent, but a reader comparing against the paper should be told.

ArkLib/OracleReduction/Prelude.lean
```diff
@@ -52,11 +50,68 @@ instance {α : Type*} {n : ℕ} [VCVCompatible α] : VCVCompatible (Fin n → α
...
+    {α : Type*} [FinEnum α] : FinEnum (Equiv.Perm α) where
+  card := Fintype.card (Equiv.Perm α)
+  equiv := Fintype.equivFin _
+  decEq := inferInstance
+
+/-- `VCVCompatible α` implies `SampleableType α`.
+
+Direct bridge: `VCVCompatible` → `FinEnum` → `SampleableType`, without relying on
+multi-step typeclass synthesis that may fail at priority 50. -/
+noncomputable instance VCVCompatible.instSampleableType
```
alexanderlhicks: [MED, library hygiene] `VCVCompatible.instSampleableType` is a global `instance` with maximally general head
`SampleableType ?α`. VCVio deliberately does *not* register this: its own
`SampleableType.ofFintype` is "*provided as a `def` rather than an `instance` to avoid overlap
with `FinEnum.SampleableType`*", and its `Fintype` bridge warns "*many types already have a
`Fintype` instance that would not be definitionally equal*". `SampleableType` carries **data**
(`selectElem`), so this instance creates non-defeq diamonds with VCVio's computable instances
(`Fin n`, `Prod`, `instSampleableTypeFunc`, `instSampleableTypePerm`, …) and with
`Sampleable.toSampleableType` — and since Prelude is imported repo-wide and later instances
win, it can silently switch which `selectElem` *existing* code elaborates against
(noncomputable `Fintype.equivFin`-based instead of the canonical one). Not a soundness issue
— the class laws force all instances to agree on every `Pr[…]` — but it's the exact hazard
VCVio documents against, now installed globally. Suggest: demote to a `def` (call sites in
this PR are few) or `scoped instance` in a DSFS namespace; keep the *specific-head* bridges
(`instSampleableTypeEncodedChallengeOracle` etc.), which are fine. Same pattern, lower risk:
`instSampleableTypeEquivPermVCV` (:110, overlaps VCVio's `instSampleableTypePerm`) and
`instSampleableTypePiVCV` (OracleDistribution.lean:69).

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Defs.lean
```diff
@@ -1,84 +1,534 @@
...
+    extends HasMessageSize pSpec, HasChallengeSize pSpec where
+  /-- `φᵢ : Message i → Σ^{ℓ_P(i)}` — message encoder (CO25 Def. 4.1). -/
+  encode : (i : pSpec.MessageIdx) → pSpec.Message i → Vector U (messageSize i)
+  encode_injective : ∀ i, Function.Injective (encode i) -- `φᵢ` is injective
+  /-- `ψᵢ : Σ^{ℓ_V(i)} → Challenge i` — challenge decoder (CO25 Def. 4.1). -/
+  decode : (i : pSpec.ChallengeIdx) → Vector U (challengeSize i) → pSpec.Challenge i
+  decodingBias : pSpec.ChallengeIdx → NNReal -- `ε_cdc`
+  /-- For every `i`, `decode i` is ε-biased: `dist (𝒰 Challenge_i) (decode_i <$> 𝒰 Domain_i)`
+    ≤ `decodingBias i`. Matches `Deserialize.CloseToUniform.ε_close` (CO25 Definition 4.1). -/
+  decode_isBiased : ∀ (i : pSpec.ChallengeIdx)
```
alexanderlhicks, line 185: [MED] The `dist` in this field elaborates against `Serde.lean`'s `Dist (PMF α)` instance
(`dist a b = ∑ x, |a x − b x|` — unhalved L1; the instance is itself marked "Local instance
for now, will need to develop statistical distance a lot more", Serde.lean:36), but the
docstring and the paper (Def 4.1:
"*ε_cdc,i-close in statistical distance*") mean total variation = L1/2 — and VCVio's
`MetricSpace (PMF α)` (imported via Preliminaries.lean) registers `dist = PMF.tvDist`, the
halved one. Two consequences: (a) `decodingBias` as constrained here is `2 × ε_cdc^paper`,
so the codec terms of `claim5_22Bound`/`ηStar` are a factor 2 off from Eq. 5 (conservative
direction — a Lean-lawful codec is paper-lawful at ε/2 — so nothing is *false*, but the
constants stop being paper-exact, which this PR otherwise achieves everywhere); (b) with two
`Dist (PMF _)` sources in scope, which one a given file means is an elaboration accident.
Suggest stating the field via `PMF.tvDist` explicitly (no ambient `Dist`), or documenting the
×2 convention at `decodingBias` and in `claim5_22Bound`/`ηStar`'s docstrings so the
sorry-filler knows which constant they owe.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Defs.lean
```diff
@@ -1,84 +1,534 @@
...
+      [Fintype (pSpec.Challenge i)] [Nonempty (pSpec.Challenge i)],
+      dist (PMF.uniformOfFintype (pSpec.Challenge i))
+        (decode i <$> PMF.uniformOfFintype (Vector U (challengeSize i))) ≤ decodingBias i
+  /-- For every `i`, `decode i` is surjective: every challenge has at least one encoded preimage.
+    Required for the `ψ⁻¹` sampler in the Section 5.8 reduction. -/
+  decode_surjective : ∀ i, Function.Surjective (decode i)
+  /-- `ψᵢ⁻¹ : Challenge i → ProbComp (Σ^{ℓ_V(i)})` — preimage sampler (CO25 Def. 4.1). -/
+  sampleChallengePreimage :
+    (i : pSpec.ChallengeIdx) → pSpec.Challenge i → ProbComp (Vector U (challengeSize i))
+
```
alexanderlhicks, line 193: [LOW] The field packages "ψᵢ⁻¹ (CO25 Def 4.1)" as a bare `ProbComp` with no law tying it to
`decode` — nothing says it lands in the fiber, let alone uniformly, which is what paper
Lemma 3.2 and the §5.8 hybrids need. Today it has zero consumers (the hybrids correctly use
`uniformDeserializePreimage`, and `Preliminaries.lean` has the lawful
`sampleUniformPreimage` + `lemma_3_2`), so nothing is wrong *yet* — but any future statement
that quantifies over `[Codec …]` and touches this field will be hollow. Suggest either adding
the law (`(sampleChallengePreimage i c).support ⊆ decode i ⁻¹' {c}` + uniformity, or defeq to
`sampleUniformPreimage`) or deleting the field until it has a consumer. Related one-liner:
`decode_surjective` strengthens paper Def 4.1 (the paper first needs surjectivity at
Lemma 3.2, not in the codec definition) — justified here, but worth a sentence in the class
docstring so codec instantiations know the deviation is deliberate.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
```diff
@@ -1,123 +1,1080 @@
...
+    exact OracleComp.IsQueryBoundP prover
+      (isD2SAlgoChallengeQuery (oSpec := oSpec) (StmtIn := StmtIn) (U := U)
+        (Salt := Salt) (pSpec := pSpec)) t
+
+set_option linter.unusedDecidableInType false in
+set_option linter.unusedFintypeInType false in
+/-- CO25 Theorem 5.1 (Main lemma §5.8) — **inner, concrete-transform form.**  Identical to
+`lemma_5_1` but with the two transforms *fixed* to the concrete `ProverTransform.d2sAlgo` and
+`d2sTraceSalted` (rather than hidden behind `∃`).  This is the form §6.2 consumes
+(`theorem_6_2_straightline`): Construction 6.3's extractor runs the concrete `d2sTraceSalted`, so
```
alexanderlhicks: [NITS, batched] Docstring/naming debris spotted while cross-checking — none blocking, batching them here:
- `KeyLemma.lean:946, 1046`: docstrings still reference `theorem_6_1_soundness` /
  `theorem_6_2_straightline`, which were removed with the §6 deferral.
- `KeyLemma.lean:518`: `ηStar` docstring's shorthand "numerator = 7(t+L)² + …" doesn't expand
  to the code's (correct, per Eq. 5) `7t² + 28(L+1)t + 14(L+1)² − 3t − 13(L+1)`
  (`7(t+L+1)²` would give `14(L+1)t`, not `28(L+1)t`). (`:504`: `θStar` cites Eq. (5); its
  defining clause is the sentence carrying Eq. (6) — marginal, take or leave.)
- `BadEvents.lean` `E_func` docstring cites `DSFS-archive/BadEvents_deep_analysis.md`, which
  isn't in the repo — inline the relevant argument or drop the pointer.
- `BadEvents.lean:1402`: `E_fork` is the one definition in the file without a docstring —
  worth noting it *is* the paper's official Def 5.13 (`|S_BT| > 1`) and that
  `E_fork_h/p/h_p` (Eqs. 39-41, currently unconsumed) are the paper's special cases whose
  exhaustiveness is not formalized.
- `OracleDistribution.lean:294-308, 380-395`: the §5.8 mapping names
  `section58EncodedChallengeOracle`/`…Decoded…` don't exist (the real decls are
  `gSpec`/`eSpec`/`D_Sigma`/`D_e` in `DuplexSponge/Defs.lean`); "Eq. 52" for the decoded
  oracle should be Eq. 53; the header says Level-3 lifts are "left as `sorry`" but the file
  contains none (say "left to downstream").
- `Backtrack.lean:729`: local `Lδ` re-definition shadows `Defs.lean:297`
  (`numSaltBlocks`, same body).
- `Backtrack.lean:1026-1028`: `Vector.replicate … 0` silently zero-fills a never-assembled
  message; `none` would be honest (reachable for non-alternating specs where only
  `lastMessageBefore?` is extracted per challenge).
- `Defs.lean:825-834` (`processRoundDSFS`): the prover absorbs the *final* message; paper
  Construction 4.3 guards steps (b)-(e) with "If i < k". Transcript-equivalent, but it shifts
  the prover's p-query count by `⌈ℓ_P(k)/r⌉` — flag for whoever proves
  `lemma_5_1_inner_queryBound`.
- `BadEvents.lean:548, 695, 719`: `traceDistOfConcreteExperiment`,
  `lemma5_8ProjectedTraceDistOfConcreteExperiment`, and `lemma5_8TraceExperiment` have no
  consumers (superseded by the abortable pipeline `lemma_5_8` actually uses); delete or mark,
  and update `Defs.lean:1047`, which cites the superseded `lemma5_8TraceExperiment`.

alexanderlhicks (changes requested):
Thanks — every item from the July review has been genuinely addressed: the §6/abort-chain
deferrals are clean (the false statements are gone from the tree, not just untouched), the
Claim 5.24 and Lemma 5.8 constants now match the paper exactly, the `E_func` strengthening is
properly disclosed, the `SampleableType` instances are proven, and the completeness statements
now sample uniform tables. Lemmas 5.10/5.12/5.14/5.16 check out as genuinely proven and
faithful, and the η★ = Claim 5.21 + 5.22 + 0 + 5.24 assembly reproduces the paper's p. 55 sum
term-for-term. Nice work.

A fresh pass over the new material found a small number of statement-level issues of the same
kind the July round targeted — sorried Props that are false or unprovable as written, all
cheap to repair *before* anyone attempts the proofs. Inline comments follow; the three HIGH
ones (missing query-bound hypotheses; a block-pointer off-by-one-round in `extractCandidate`;
duplicate-entry lookups aborting on real traces) are the ones I'd gate on.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
chung-thai-nguyen: [DONE - C19] Resolved via Nondegeneracy.lean - implicitly required by the paper proof strategy.

ArkLib/OracleReduction/Prelude.lean
chung-thai-nguyen: [DONE - C22] Demoted the general-head sampling bridges to explicit definitions to preserve/avoid conflicts with **VCVio’s canonical function/permutation instances**. The dependent FS/encoded/decoded oracle-family instances remain specific-head bridges (`instSampleableTypeEncodedChallengeOracle` etc), which Alex explicitly confirmed are fine.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
chung-thai-nguyen: [DONE - C25]
- `KeyLemma.lean:946,1046`: removed stale Section 6 theorem references.
- `KeyLemma.lean:504,518`: corrected `θStar` to Eq. (6) and expanded the exact `ηStar` numerator.
- `BadEvents.lean:E_func`: replaced the missing archive link with an inline explanation.
- `BadEvents.lean:1402`: documented `E_fork` and the status of Eqs. (39)–(41).
- `OracleDistribution.lean:294–308,380–395`: corrected the distribution names, locations, and Eq. (53) reference. The Level-3 wording was already current.
- `Backtrack.lean:729`: removed the duplicate local `Lδ`.
- `Backtrack.lean:1026–1028`: missing messages now reject the candidate instead of being zero-filled.
- `Defs.lean:825–834`: documented final-message absorption and its query-bound impact.
- `BadEvents.lean:548,695,719`: deleted the unused experiment helpers and updated the stale `Defs.lean:1047` reference.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Defs.lean
chung-thai-nguyen: [DONE - C24] Removed the unused, unconstrained Codec.sampleChallengePreimage field and its Codec.mk' argument. decode_surjective remains and is used by the lawful uniformDeserializePreimage construction, which is now the **sole** preimage-sampling path.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
chung-thai-nguyen: [DONE - C15] Claims 5.21, 5.22, and 5.24 now require IsLemma5_1QueryBound maliciousProver tₕ tₚ tₚᵢ, tying every numerical query budget to the actual malicious prover. The previously possible arbitrary/zero-budget instantiations are therefore excluded.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/Backtrack.lean
chung-thai-nguyen: [DONE - C16] L_ptr(i) now excludes the current round’s prover message via `sumMessageBlocksBeforeExclLast`, matching the paper’s strictly-earlier-round sum. The case `L_ptr = Lptr_paper(i) + L_P(i)` no more happens. Moreover, as required implicitly by the paper and the proof strategy, **NondegenerateRounds** is now added as a constraint for pSpec.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/Lookahead.lean
chung-thai-nguyen: [DONE - C18] A forward self-capacity loop now terminates the LookAhead scan with `.done`, rather than producing `forkErr`. The comment now correctly classifies this as an E_p case.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Defs.lean
chung-thai-nguyen: [DONE - C23] Removed ArkLib’s competing unhalved-L1 Dist (PMF α) instance. Both `Deserialize.CloseToUniform.ε_close` and `Codec.decode_isBiased` now explicitly use `PMF.tvDist`, so `decodingBias`, `claim5_22Bound`, and `ηStar` follow the paper’s total-variation convention exactly, without ambient-instance ambiguity or a factor-two workaround.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/TraceDataStructures.lean
chung-thai-nguyen: [DONE - C17] Both suggested changes are applied, similar holes are also updated to ensure consistency:
- Trace tables now treat repeated identical `(query, answer)` pairs `idempotently` while preserving genuine conflicts between distinct mappings.
- All trace-construction and live D2S update paths use set-semantic `TraceTableOps.insert`
- lookup laws and BackTrack/LookAhead candidate searches **ignore duplicate occurrences**.
- Regression lemmas cover both successful identical replay and rejection of distinct conflicts.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/BadEvents.lean
chung-thai-nguyen: [DONE - C20]
- `t_p ≥ L` is intentionally dropped because it's not necessary ƒor the probability argument of Lemma 5.8
- Added shared `L_totalRateBlocks δ pSpec = L_δ + L_P + L_V` and used it in the bounds of the soundness lemmas to account the salt absorption. It is used consistently by Lemma 5.8, Claims 5.21/5.24, Lemma 5.1, and `ηStar`. Claims 5.22/5.23 require no such parameter.

ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean
chung-thai-nguyen: [DONE - C21]:
- Documented that `t_p ≥ max {L_P, L_V}` belongs to the paper’s unformalized running-time analysis and is unnecessary for the query/TV-distance statements, so it's omitted at least for now.
- Also documented that although the existential theorem type permits budget-dependent witnesses, the supplied d2sAlgo and d2sTrace(Salted) witnesses are budget-independent.

chung-thai-nguyen:
Thanks for your reviews @alexanderlhicks. I have resolved them. Now the PR is ready for a new round of review.

PR #470, opened 2026-04-21
fix(BCIKS20): correct H_tilde' coefficient indexing

ainta:
## Summary

Fix `H_tilde'` in `ArkLib/Data/Polynomial/RationalFunctions.lean` so that it matches the monicization defined by `H_tilde`.

- correct the coefficient formula in `H_tilde'`
- fix the `natDegree = 0` case, which should return `Polynomial.C (H.coeff 0)`
- prove `H_tilde_equiv_H_tilde'` for the corrected definition

On the counterexample from #438, for `H = h₀ + h₁Y + h₂Y²`, the old `H_tilde'` gives constant term `h₂²`, while the corrected definition gives `h₀ * h₂`, matching `H_tilde`.

Closes #438.

### DIFF
--- a/ArkLib/Data/Polynomial/RationalFunctions.lean
+++ b/ArkLib/Data/Polynomial/RationalFunctions.lean
@@ -83,6 +83,19 @@ noncomputable def H_tilde' (H : F[X][Y]) : F[X][Y] :=
       ∑ i ∈ Finset.range d,
         Polynomial.C (hᵢ i * W ^ (d - 1 - i)) * Polynomial.X ^ i
 
+omit [IsDomain F] in
+/-- If `H` has positive degree in `Y`, then `H_tilde' H` is monic. -/
+lemma H_tilde'_monic (H : F[X][Y]) (hH : 0 < H.natDegree) :
+    (H_tilde' H).Monic := by
+  classical
+  have hdeg : H.natDegree ≠ 0 := Nat.ne_of_gt hH
+  rw [H_tilde', if_neg hdeg]
+  exact Polynomial.monic_X_pow_add <| (Polynomial.degree_sum_le _ _).trans_lt <| by
+    exact (Finset.sup_lt_iff (WithBot.bot_lt_coe H.natDegree)).2 <| by
+      intro i hi
+      exact (Polynomial.degree_C_mul_X_pow_le i _).trans_lt
+        (WithBot.coe_lt_coe.2 (Finset.mem_range.mp hi))
+
 private lemma monicize_term {K : Type} [Field K] (a b : K) (i d : ℕ)
     (ha : a ≠ 0) (hi : i < d) :
     (Polynomial.C a ^ (d - 1)) * (Polynomial.C b * (Polynomial.X / Polynomial.C a) ^ i) =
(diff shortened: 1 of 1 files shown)


quangvdao:

The core fix looks right to me: the corrected coefficient indexing matches the termwise monicization of `H_tilde`, and the `natDegree = 0` branch is the right mathematical value for constant-in-`Y` polynomials.

One follow-up design point before this is ready for approval/merge from my side: this PR makes explicit that `H_tilde' H` is not always monic. If `H.natDegree = 0`, then `H_tilde' H = Polynomial.C (H.coeff 0)`, which is generally not monic. That is fine for the equivalence theorem, but later API such as `canonicalRepOf𝒪` uses `Polynomial.modByMonic`, so the BCIKS development should make the positive `Y`-degree assumption explicit wherever quotient/remainder facts are intended.

The shape I’d expect is:

```lean
lemma H_tilde'_monic (H : F[X][Y]) (hH : 0 < H.natDegree) :
    (H_tilde' H).Monic := ...
```

Then thread this invariant through the parts of the API that rely on canonical representatives. Concretely:

```lean
noncomputable def canonicalRepOf𝒪 {H : F[X][Y]}
    (hH : 0 < H.natDegree) (β : 𝒪 H) : F[X][Y] :=
  Polynomial.modByMonic β.out (H_tilde' H)
```

and then update downstream users accordingly:

```lean
noncomputable def weight_Λ_over_𝒪 {H : F[X][Y]}
    (hH : 0 < H.natDegree) (f : 𝒪 H) (D : ℕ) : WithBot ℕ :=
  weight_Λ (canonicalRepOf𝒪 hH f) H D
```

The same `hH` should then appear in later BCIKS statements that use `weight_Λ_over_𝒪`, such as `Lemma_A_1`, and eventually in the Claim A.2 API if those bounds depend on canonical representatives. This avoids pretending the constant-in-`Y` case has a monic modulus, while preserving the corrected value of `H_tilde'`.

On the private lemmas: I think they are okay as private helpers for this PR. If we want reusable infrastructure, I would not expose `monicize_term` and `monicize_leading_term` exactly as written, since they are specialized to this proof’s `d - 1` setup. A cleaner reusable lemma would be a general polynomial rescaling identity in the `Polynomial` namespace for terms of the form

```lean
Polynomial.C a ^ n * (Polynomial.C b * (Polynomial.X / Polynomial.C a) ^ i)
```

with the BCIKS-specific `d - 1` and leading-term cases derived locally.

Please also update this branch onto current `main` before merge.

ainta:
Thanks, I addressed this by adding `H_tilde'_monic` under the hypothesis `hH : 0 < H.natDegree`, and threading that hypothesis through the modByMonic/canonical-representative path: `canonicalRepOf𝒪`, `weight_Λ_over_𝒪`, `Lemma_A_1`, `weight_ξ_bound`, and `β_regular`.

I kept the monicization helper lemmas private, as suggested.

I also rebased the branch onto current `main` and re-ran:

- `lake env lean ArkLib/Data/Polynomial/RationalFunctions.lean`
- `lake env lean ArkLib/Data/CodingTheory/ProximityGap/BCIKS20/ListDecoding/Agreement.lean`
- `git diff --check`

The branch is now pushed at `4abb2ba88863927d3041f351fd214203c8f1ce0f`.

One API choice: I kept the existing `β`/`α`/`γ` consumer API stable, and made the positive-degree hypothesis explicit on the statements that actually rely on canonical representatives. I’m happy to thread `hH` through those public constructors too if you prefer the stricter Claim A.2 API.

quangvdao (approved):

PR #429, opened 2026-03-23
δ must be positive in proximity gap theorems

ElijahVlasov:
δ must be positive in proximity gap theorems and strictly less than the Johnson bound.

PR #475, opened 2026-04-27
feat(proximity-gap): correlated agreement for affine spaces (Thm 1.7)

XC0R:
Fully prove `correlatedAgreement_affine_spaces`; the (δ,ε)-correlated agreement theorem for affine spaces over Reed-Solomon codes (BCIKS20 §6.3, Theorem 1.7).

Proof structure follows the paper:

1. All elements of U are δ-close to V (Lemma 6.3 + scaling invariance + span extension)
2. Divergence-attaining basepoint u* with δ* = δᵣ'(u*, V)
3. Bucketing via list-decoding: < |F| codewords per element, pigeonhole on |F|^k elements gives common codeword v₀ for all of U
4. jointAgreement from shared agreement set D' of size ≥ (1-δ)|ι|

Also adds:

- `prob_coeff_eq_prob_affine`: bridge between coefficient-parameterised and subtype-parameterised probability (fiber-constant surjection)
- `gs_existence`, `gs_divisibility`: Guruswami-Sudan with rate-corrected degree bound (ρ = k/n)
- `rs_listDecoding_card_lt_field`: RS list-decoding bound with full case analysis (deg = 0, 1, ≥ 2)
- `exists_gs_multiplicity`: internal GS multiplicity selection with m = ⌈√ρ/(2η)⌉ + 1

Breaks the import cycle `AffineSpaces → DivergenceOfSets → ReedSolomonGap → AffineSpaces` by parameterizing `proximity_gap_affineSubspace` and `concentration_bounds` with `hPG` hypothesis, removing the `ReedSolomonGap` import from `DivergenceOfSets.lean`.

### DIFF
--- a/ArkLib/Data/CodingTheory/DivergenceOfSets.lean
+++ b/ArkLib/Data/CodingTheory/DivergenceOfSets.lean
@@ -10,7 +10,6 @@ import ArkLib.Data.CodingTheory.Basic.LinearCode
 import ArkLib.Data.CodingTheory.Basic.RelativeDistance
 import ArkLib.Data.CodingTheory.ProximityGap.Basic
 import ArkLib.Data.CodingTheory.ProximityGap.BCIKS20.ErrorBound
-import ArkLib.Data.CodingTheory.ProximityGap.BCIKS20.ReedSolomonGap
 import ArkLib.Data.CodingTheory.ReedSolomon
 import ArkLib.Data.Probability.Notation
 import Mathlib.LinearAlgebra.AffineSpace.AffineSubspace.Defs
@@ -184,8 +183,11 @@ theorem proximity_gap_affineSubspace {ι : Type} [Fintype ι] [Nonempty ι] [Dec
     {F : Type} [Fintype F] [Field F] [DecidableEq F]
   {deg : ℕ} {domain : ι ↪ F}
   (U : AffineSubspace F (ι → F)) [Nonempty U] {δ : ℝ≥0}
-  (hδ : δ ≤ 1 - ReedSolomon.sqrtRate deg domain)
-  (hε : ProximityGap.errorBound δ deg domain < 1) :
+  (_hδ : δ < 1 - ReedSolomon.sqrtRate deg domain)
+  (hPG : ∀ {k t : ℕ} [NeZero k] [NeZero t] (C : Fin t → (Fin k → (ι → F))),
+    ProximityGap.δ_ε_proximityGap
+      (ReedSolomon.toFinset domain deg)
+      (Affine.AffSpanFinsetCollection C) δ (errorBound δ deg domain)) :
   Xor'
     (Pr_{let u ← $ᵖ U}[Code.relDistFromCode u (RScodeSet domain deg) ≤ δ] = 1)
     (Pr_{let u ← $ᵖ U}[Code.relDistFromCode u (RScodeSet domain deg) ≤ δ] ≤
@@ -199,13 +201,13 @@ theorem proximity_gap_affineSubspace {ι : Type} [Fintype ι] [Nonempty ι] [Dec
   let u : Fin k → (ι → F) := fun i => (e i : U)
   -- Turn this into a (trivial) collection of one affine span
   let C : Fin 1 → (Fin k → (ι → F)) := fun _ => u
-  -- Apply ProximityGap Theorem 1.2
+  -- Apply ProximityGap Theorem 1.2 via hypothesis
   have hpg : ProximityGap.δ_ε_proximityGap
       (ReedSolomon.toFinset domain deg)
       (Affine.AffSpanFinsetCollection C)
       δ
       (errorBound δ deg domain) :=
-    ProximityGap.proximity_gap_RSCodes (C := C) (deg := deg) (domain := domain) (δ := δ) hδ hε
+    hPG C
   -- Specialize to the unique element of the collection
   let S : Finset (ι → F) := Affine.AffSpanFinset (C 0)
   have hS_mem : S ∈ Affine.AffSpanFinsetCollection C := by
@@ -801,8 +803,11 @@ theorem concentration_bounds {ι : Type} [Fintype ι] [Nonempty ι] [DecidableEq
   (hdiv_pos : 0 < (divergence U (RScodeSet domain deg) : ℝ≥0))
   (hdiv_lt : (divergence U (RScodeSet domain deg) : ℝ≥0) <
     1 - ReedSolomon.sqrtRate deg domain)
-  (hε_cb : ∀ (δ' : ℝ≥0), δ' ≤ 1 - ReedSolomon.sqrtRate deg domain →
-    ProximityGap.errorBound δ' deg domain < 1) :
+  (hPG : ∀ {δ : ℝ≥0}, 0 < δ → δ < 1 - ReedSolomon.sqrtRate deg domain →
+    ∀ {k t : ℕ} [NeZero k] [NeZero t] (C : Fin t → (Fin k → (ι → F))),
+    ProximityGap.δ_ε_proximityGap
+      (ReedSolomon.toFinset domain deg)
+      (Affine.AffSpanFinsetCollection C) δ (errorBound δ deg domain)) :
     let δ' := divergence U (RScodeSet domain deg)
     Pr_{let u ← $ᵖ U}[Code.relDistFromCode u (RScodeSet domain deg) ≠ δ']
       ≤ errorBound δ' deg domain := by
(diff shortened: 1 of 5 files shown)


quangvdao:
Thanks for this PR. I think the main theorem is very close, but I do not think `hGS` should remain as a public hypothesis.

Currently `correlatedAgreement_affine_spaces` asks the caller to provide:

```lean
hGS : 1 < deg → ∃ m : ℕ, 1 ≤ m
  ∧ (δ : ℝ) < gs_johnson deg (Fintype.card ι) m
  ∧ gs_degree_bound deg (Fintype.card ι) m / (deg - 1) < Fintype.card F
```

This is exactly the Guruswami-Sudan multiplicity-selection arithmetic that the BCIKS20 proof performs internally. In the paper, the multiplicity is chosen from the slack

```lean
η = 1 - sqrtRate - δ
```

and the proof then verifies both:

```lean
δ < gs_johnson deg n m
gs_degree_bound deg n m / (deg - 1) < q
```

So exposing `hGS` makes this theorem a conditional variant of Theorem 1.7, rather than the paper theorem.

I think the right fix is:

1. Tighten the Johnson-bound hypothesis from

```lean
δ ≤ 1 - ReedSolomon.sqrtRate deg domain
```

to

```lean
0 < δ
δ < 1 - ReedSolomon.sqrtRate deg domain
```

This matches BCIKS20 and also matches the direction already proposed in #429. The closed endpoint is not actually usable here: if `δ = 1 - sqrtRate`, then for every finite `m`,

```lean
gs_johnson deg n m = 1 - sqrtRate - sqrtRate / (2 * m) < δ
```

so the strict `δ < gs_johnson ... m` part of `hGS` is impossible.

2. Prove a local/helper lemma that constructs the GS multiplicity.

Morally:

```lean
lemma exists_gs_multiplicity
    (hδPos : 0 < δ)
    (hδ : δ < 1 - ReedSolomon.sqrtRate deg domain)
    (hRS : deg + 1 ≤ Fintype.card ι)
    (hε : errorBound δ deg domain < 1) :
    ∃ m : ℕ, 1 ≤ m
      ∧ (δ : ℝ) < gs_johnson deg (Fintype.card ι) m
      ∧ gs_degree_bound deg (Fintype.card ι) m / (deg - 1) < Fintype.card F
```

The explicit multiplicity should be slightly stronger than the informal paper expression:

```lean
m = Nat.ceil (sqrtRate / (2 * η)) + 1
```

where `η = 1 - sqrtRate - δ`.

The `+ 1` is useful because `ceil (sqrtRate / (2η))` only gives a non-strict inequality in the exact-integral case, while the GS divisibility call needs strict `<`.

3. Use `errorBound δ deg domain < 1`, or an equivalent explicit field-size hypothesis, to prove the degree/list-size side.

The current `hF : 5 ≤ Fintype.card F` is enough only for the existing special `m = 1` shortcut. Once `m` is chosen as above, the proof needs the BCIKS20 field-size condition hidden in the Johnson-regime error bound. The cleanest Lean formulation seems to be to derive/use:

```lean
errorBound δ deg domain < 1
```

This is already the direction taken in #477 for `proximity_gap_RSCodes`.

4. Replace the `hGS hdeg` call with this helper.

The current use site is the multiplicity selection inside `rs_listDecoding_card_lt_field` / the affine-spaces proof. After the helper exists, the proof should manufacture `m` internally, and the public theorem no longer needs `hGS`.

A possible structure is:

```lean
obtain ⟨m, hm, hm_johnson, hm_degree⟩ :=
  exists_gs_multiplicity hδPos hδ hRS hε
```

or, for the later `δ_star` use, with the already-proved `hεδ_star_lt_one`.

One more small issue: the docstring says Theorem 1.6, but this affine-space statement is BCIKS20 Theorem 1.7. Theorem 1.6 is the low-degree parameterized curves theorem.

Interaction with existing PRs:

- #429 already changes these proximity-gap hypotheses to `0 < δ` and strict `< 1 - sqrtRate`. This PR should either merge/rebase after #429, or apply the same convention here.
- #477 adds `hε : errorBound δ deg domain < 1` to the RS proximity-gap theorem. That is aligned with what this PR needs for the GS multiplicity selection. I would suggest reusing that `errorBound < 1` formulation rather than adding an unrelated custom field-size hypothesis.

Requested change: please remove `hGS` from the public theorem signature by proving the GS multiplicity-selection arithmetic internally, under strict `δ < 1 - sqrtRate` and the appropriate `errorBound < 1` / field-size condition.

XC0R:
All review items addressed:

1. `hGS` removed from public signature. GS multiplicity proved internally via `exists_gs_multiplicity` with witness `m = ⌈√ρ/(2η)⌉ + 1`.
2. Strict δ: `hδ_pos : 0 < δ`, `hδ : δ < 1 - sqrtRate`.
3. `exists_gs_multiplicity` sorry-free. Johnson bound from `gs_sufficient_multiplicity_bound`, degree bound via `gs_degree_bound_div_lt` + `errorBound < 1` field-size extraction.
4. `hε : errorBound δ deg domain < 1` used for field-size condition, aligned with #477.
5. Docstring + title corrected to Theorem 1.7.

Additional hypotheses on `correlatedAgreement_affine_spaces`:
- `hdeg : 1 < deg` — required by multiplicity selection (false for deg ≤ 1)
- `hJohnson` — Johnson regime restriction
- `hε` — errorBound < 1

Import cycle (`DivergenceOfSets → ReedSolomonGap`) broken by abstracting `proximity_gap_affineSubspace` via `hPG` parameter. Subsumes #477.

quangvdao:
Thanks for the quick fixes. I think the branch is much closer now, and CI being green is good, but I still see a few substantive API/theorem-shape and performance issues that should be resolved before merge.

1. `proximity_gap_RSCodes` now appears to be Johnson-regime only, but is still documented and named as the full BCIKS20 Theorem 1.2.

The new signature requires:

```lean
(hJohnson : (1 - (LinearCode.rate (ReedSolomon.code domain deg) : ℝ≥0)) / 2 < δ)
```

so it excludes the unique-decoding branch of `errorBound`, even though `errorBound` explicitly has both the unique-decoding branch and the Johnson branch. This makes the public theorem weaker than its docstring/name suggest.

I think we should either:

- restore the unique-decoding branch and make `proximity_gap_RSCodes` cover the full stated range, or
- rename/redocument this as the Johnson-regime RS proximity gap theorem, so users do not mistake it for the full Theorem 1.2.

2. `concentration_bounds` now takes an abstract `hPG` that does not seem directly instantiable from the PR's own `proximity_gap_RSCodes`.

`concentration_bounds` asks for an `hPG` working for every:

```lean
δ ≤ 1 - ReedSolomon.sqrtRate deg domain
```

but `proximity_gap_RSCodes` now additionally needs strict positivity, strict `<`, `hε`, `hdeg`, `hRS`, `hF`, and the Johnson-regime hypothesis above. In particular it cannot supply the requested `hPG` at the boundary or in the unique-decoding regime.

So the abstraction in `DivergenceOfSets` seems to hide real obligations rather than preserving a usable RS corollary. Could you either adjust the abstract hypothesis to match the actual RS theorem range, or provide a wrapper/corollary showing how the new `proximity_gap_RSCodes` instantiates the needed `hPG` for the concentration theorem?

3. The new `maxHeartbeats 800000` on `exists_gs_multiplicity` is a performance concern.

Our Lean build times are already creeping up, so I do not want to merge a large heartbeat increase without trying to reduce it or isolate the expensive arithmetic proof. Could you please investigate whether this proof can be split into smaller helper lemmas, moved to a narrower file, or otherwise made less fragile before merge? If the high heartbeat is truly unavoidable, please document the measured impact and why this bound is needed.

There are also a few smaller cleanup items:

- `ReedSolomonGap.lean` still says `correlatedAgreement_affine_spaces` is Theorem 1.6 and currently `sorry`'d. It is now Theorem 1.7 and proved.
- The docstring for `correlatedAgreement_affine_spaces` says the close probability is “at most `ε`”, but the actual correlated-agreement definition triggers when it is `> ε`.
- There is a “sorry for now” comment inside completed code in `AffineSpaces.lean`; please rewrite it to describe the proof step without mentioning `sorry`.

I would like to see these addressed before approving, since otherwise we may merge a theorem/API surface whose name suggests a full RS proximity-gap result, while the implemented statement only covers the Johnson branch, does not compose cleanly with the downstream concentration theorem, and may add noticeable proof-performance cost.

XC0R:
Addressed all first-review items. Import cycle resolved.

- `hGS` removed; `exists_gs_multiplicity` proved internally with `m = ⌈√ρ/(2η)⌉ + 1`
- Strict δ: `_hδ_pos : 0 < δ`, `hδ : δ < 1 - sqrtRate`
- `_hε : errorBound δ deg domain < 1` for field-size condition
- Docstring and all internal references corrected to Theorem 1.7

Import cycle `AffineSpaces → DivergenceOfSets → ReedSolomonGap → AffineSpaces` broken by parameterizing `proximity_gap_affineSubspace` and `concentration_bounds` with `hPG` hypothesis, removing the `ReedSolomonGap` import from `DivergenceOfSets.lean`.

Additional hypotheses vs the original sorry stub:
- `hdeg : 0 < deg`; required by `errorBound_mono` for δ_star derivation (`errorBound` is not monotone when `deg = 0`)
- `hRS : deg + 1 ≤ Fintype.card ι`
- `_hδ_pos`, `_hε` on `correlatedAgreement_affine_spaces`
- Matching hypotheses on `proximity_gap_RSCodes` (forced by the call in the k ≥ 2 path)

The strict δ and internal GS multiplicity suggestions sharpened the theorem considerably.

quangvdao:
Thanks for the latest update. This addresses several of the earlier concerns: the stale Theorem 1.6 / `sorry` docs are fixed, the correlated-agreement docstring now has the right `> ε` direction, and the list-decoding proof now appears to split internally between the unique-decoding and Johnson regimes rather than exposing `hJohnson`.

I still see two things I would like resolved before approval.

1. `concentration_bounds` still does not seem directly instantiable from `proximity_gap_RSCodes`.

The abstract hypothesis in `DivergenceOfSets.lean` asks for an `hPG` for every:

```lean
δ < 1 - ReedSolomon.sqrtRate deg domain
```

but `proximity_gap_RSCodes` requires additional hypotheses:

```lean
0 < δ
errorBound δ deg domain < 1
hdeg : 0 < deg
hRS : deg + 1 ≤ Fintype.card ι
```

In particular, the `δ` chosen inside `concentration_bounds` can still be `0` when the divergence predecessor is zero, while `proximity_gap_RSCodes` requires `0 < δ`. So the abstraction still seems stronger than the concrete RS theorem it is meant to package.

Could you either adjust the `hPG` hypothesis so it matches the actual usable RS range, or add a wrapper/corollary showing that `proximity_gap_RSCodes` supplies the `hPG` needed by `concentration_bounds` under the intended assumptions?

2. The `exists_gs_multiplicity` proof is still a performance hotspot.

The heartbeat bound is reduced, which helps, but this is still present:

```lean
set_option maxHeartbeats 400000 in
```

Given build times are already creeping up, I would prefer not to merge a new 400k-heartbeat arithmetic proof without either factoring it further or documenting measured impact. Could you try splitting the arithmetic into smaller helper lemmas, or report the local/CI timing impact and why the remaining heartbeat bound is necessary?

Smaller cleanup: there are some stale/conflicting proof comments around the basepoint choice in `correlatedAgreement_affine_spaces`; please clean those up while touching the file.

Also, the branch is currently behind `main` by `adb6be12` (`docs: fix knowledge base audit references (#487)`), so it should be updated before merge.

XC0R:
Addressed review items:

- `concentration_bounds.hPG` now requires `0 < δ`, matching `proximity_gap_RSCodes`. The δ = 0 edge case is handled by applying hPG at δ'/2 with errorBound monotonicity.
- Extracted `gs_degree_bound_le_inv_mu` helper from `exists_gs_multiplicity`. Removed the `maxHeartbeats 400000` override (now under default 200k).
- Removed stale basepoint comments in `correlatedAgreement_affine_spaces`.
- Rebased onto current main (09074798).

quangvdao (approved):
Approved. The latest revision addresses the remaining review concerns: the `hPG` hypothesis now matches the positive-δ RS proximity gap shape, the δ = 0 concentration edge case is handled via a positive intermediate threshold, and the high-heartbeat GS arithmetic proof was factored to run under the default heartbeat budget.

PR #477, opened 2026-04-27
Break DivergenceOfSets → ReedSolomonGap import cycle + Thm 1.2 proof

XC0R:
## Summary

- Refactor `proximity_gap_affineSubspace` and `concentration_bounds` in `DivergenceOfSets.lean` to take the proximity gap result as a hypothesis parameter instead of calling `proximity_gap_RSCodes` directly. Remove the `import ReedSolomonGap` from `DivergenceOfSets.lean`.
- Port the `proximity_gap_RSCodes` (Theorem 1.2, BCIKS20) reduction into `ReedSolomonGap.lean`, now importing `AffineSpaces.lean`. The proof reduces to `correlatedAgreement_affine_spaces` (Theorem 1.6) via a `Xor'` case split: the right branch contradicts `hε < 1`, the left branch obtains `jointAgreement` for each word stack element.
- Add `hε : errorBound δ deg domain < 1` to the `proximity_gap_RSCodes` signature (required for `Xor'` exclusivity).

Breaks the latent import cycle `ReedSolomonGap → AffineSpaces → DivergenceOfSets → ReedSolomonGap` that would appear when both the Thm 1.6 proof (#475) and this Thm 1.2 proof are merged.

### DIFF
--- a/ArkLib/Data/CodingTheory/DivergenceOfSets.lean
+++ b/ArkLib/Data/CodingTheory/DivergenceOfSets.lean
@@ -10,7 +10,6 @@ import ArkLib.Data.CodingTheory.Basic.LinearCode
 import ArkLib.Data.CodingTheory.Basic.RelativeDistance
 import ArkLib.Data.CodingTheory.ProximityGap.Basic
 import ArkLib.Data.CodingTheory.ProximityGap.BCIKS20.ErrorBound
-import ArkLib.Data.CodingTheory.ProximityGap.BCIKS20.ReedSolomonGap
 import ArkLib.Data.CodingTheory.ReedSolomon
 import ArkLib.Data.Probability.Notation
 import Mathlib.LinearAlgebra.AffineSpace.AffineSubspace.Defs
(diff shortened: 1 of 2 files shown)


quangvdao:
Nice import-cycle break overall. I sanity-checked the dependency direction and the theorem shape, and the structural refactor looks sound: `DivergenceOfSets.lean` no longer imports `ReedSolomonGap.lean`, while `ReedSolomonGap.lean` can import `AffineSpaces.lean`.

One concern: the refactor seems to leave an instantiation gap for the concentration corollary. `proximity_gap_RSCodes` now requires

```lean
(hε : errorBound δ deg domain < 1)
```

but `DivergenceOfSets.concentration_bounds` now assumes a fully formed proximity-gap hypothesis for every `δ ≤ 1 - sqrtRate`:

```lean
(hPG : ∀ {δ : ℝ≥0}, δ ≤ 1 - ReedSolomon.sqrtRate deg domain →
  ∀ {k t : ℕ} [NeZero k] [NeZero t] (C : Fin t → (Fin k → (ι → F))),
  ProximityGap.δ_ε_proximityGap
    (ReedSolomon.toFinset domain deg)
    (Affine.AffSpanFinsetCollection C) δ (errorBound δ deg domain))
```

This is logically fine as an abstraction lemma, but it means the PR does not provide a direct usable specialization from `proximity_gap_RSCodes` to `concentration_bounds` unless callers can also supply `errorBound δ deg domain < 1` at each threshold. Could you either:

1. rename/document the `DivergenceOfSets` results as `_of_proximity_gap` style abstraction lemmas, or
2. add a downstream theorem importing both sides that recovers the concrete concentration bound from `proximity_gap_RSCodes`, with the needed `hε` assumption made explicit?

I also locally checked that both changed files compile:
`lake env lean ArkLib/Data/CodingTheory/DivergenceOfSets.lean`
`lake env lean ArkLib/Data/CodingTheory/ProximityGap/BCIKS20/ReedSolomonGap.lean`

XC0R:
Superseded by #475 which includes the import cycle break.

PR #430, opened 2026-03-26
Proximity Generators Definitions 

katyhr:
This PR achieves four main goals: 

1) formalises main concepts and results from Section `3` of `All Polynomial Generators Preserve Distance
with Mutual Correlated Agreement`. More precisely, 

- Def 3.5 MDS codes
- Lemma 3.6 (+proof)
- Def 3.7
- Def 3.10 generator
- Def 3.11 zero-evading generator
- Def 3.12 MDS generator
- Def 3.14 MCA generator
- Def 3.19 polynomial generator
- Remark 3.20 (plus proof) every polynomial generator is zero-evading

2) missing lemmas for generator matrices + theorem establishing existence of generator matrices
3) Reed-Solomon codes are MDS
4) a "counting" version of Schwartz-Zippel lemma for multivariate polynomials. Needed for the proof of Remark 3.20 but could also be of independent mathematical interest

PR #489, opened 2026-05-04
MCA generators lemmas

katyhr:
This PR formalises and proves Lemma 4.1, Lemma 4.2 and Definition 4.3 of [Bordage, S., Chiesa, A., Guan, Z., Manzur, I., *All Polynomial Generators Preserve Distance with Mutual Correlated Agreement*][BSGM25]

We add extra definitions and lemmas needed for that. More notably, 
- some definitions and a result on left pseudo inverses of matrices;
- a lemma on linear combinations of elements of a projected code


ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
```diff
@@ -0,0 +1,184 @@
...
+def IsMCAGenerator {S : Type} [Nonempty S] [Fintype S] (G : Generator S ℓ F) (ε_mca : I → I)
+  (LC : LinearCode ι F) : Prop :=
+  ∀ U : ℓ → (ι → F), ∀ γ : I,
+    Pr_{let x ←$ᵖ S}[(IsMCA G LC x U γ)] ≤ ENNReal.ofReal (ε_mca γ)
+
+/-- Let `G : S →F^ℓ` and `G′: S′→F^ℓ` be two generators. Their tensor product is the generator
+`G ⊗ G′: S × S′→ F^ℓ ⊗F^ℓ′` defined by `(x,x′) ↦ G(x) ⊗ G′(x′)`.
+Definition 4.3 [BCGM25]. -/
+def TensorGenerator {ℓ' : Type} [Fintype ℓ'] (S S' : Type)
+  (G : Generator S ℓ F) (G' : Generator S' ℓ' F) : (S × S') → TensorProduct F (ℓ → F) (ℓ' → F)
```
alexanderlhicks: **Correctness concern.** Paper Def 4.3 (p.22) says `G ⊗ G': S × S' → F^ℓ ⊗ F^ℓ' ≅ F^(ℓ×ℓ')` and uses the iso to treat the result as a `Generator (S × S') (ℓ × ℓ') F` — i.e., its output is a vector indexed by pairs. Downstream lemmas (Lemma 4.4 tensor MCA, p.23) need `(G ⊗ G')(x, x') · U` as a matrix-vector product on `ℓ × ℓ'`, which won't typecheck against an abstract `TensorProduct` element.

As written here, `TensorGenerator G G'` is **not** a `Generator (S × S') (ℓ × ℓ') F` and can't be plugged into `IsMCAGenerator` / `IsZeroEvadingGenerator` without first applying the iso.

Suggest defining the elementary-tensor form directly (this is mathematically equivalent and applies the iso by unfolding):

```lean
def TensorGenerator {ℓ' : Type} [Fintype ℓ'] {S S' : Type}
    (G : Generator S ℓ F) (G' : Generator S' ℓ' F) :
    Generator (S × S') (ℓ × ℓ') F
  | (x, x'), (i, j) => G x i * G' x' j
```

While you're there, two style notes:
- `(S S' : Type)` could be `{S S' : Type}` (implicit) — saves call sites from passing them.
- `Type` could be `Type*` to avoid universe constraints (matching the rest of the file).

ArkLib/Data/CodingTheory/Basic/LinearCode.lean
```diff
@@ -265,6 +253,156 @@ scoped syntax &"ρ" term : term
 scoped macro_rules
   | `(ρ $t:term) => `(LinearCode.rate $t)
 
+/-- Let `c` be a word of length `ι`. For every finite `ι`-subset `T` , we define the projection of a
+word `c` to `T` as the word obtained by restricting the indexing set of `c` to `T`.
+We denote this by `c|[T]`.
+Definition 3.6 [BCGM25]. -/
```
alexanderlhicks: **Regression from main's state after #430.** Paper p.17 has:
- *Definition 3.5* — MDS code
- *Lemma 3.6* — MDS iff k rows are linearly independent (a Lemma, not a Definition)
- *Definition 3.7* — projection of code `C|_T`

So `Definition 3.7` is the correct citation here (and on line 266 below for `projectedCode`). #430 set this to `3.7` after my earlier review; this PR regresses to `3.6`. Please revert both occurrences to `3.7`.

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
```diff
@@ -0,0 +1,184 @@
...
+| (x, x') => TensorProduct.tmul F (G x) (G' x')
+
+end CoreDefinitions
+
+namespace PolyGenIsZeroEvading
```
alexanderlhicks: **Namespace regression.** Previously `namespace PolynomialGenerator`, now `PolyGenIsZeroEvading`. The narrower name limits future extensibility — any subsequent polynomial-generator result (e.g., the MCA theorems from §6 of the paper) won't naturally fit here.

Suggest reverting to `namespace PolynomialGenerator`.

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
```diff
@@ -0,0 +1,184 @@
...
+  unfold minSeedCard
+  split_ifs
+  aesop
+
+noncomputable local instance {F : Type} [Fintype F] {S : Set F} : Fintype S := Fintype.ofFinite ↑S
+
+/-- If `G` is a polynomial generator, then `G` is zero-evading with error the maximum of the total
+degrees of the individual polynomials divided by the size of the smallest evaluation sets `S i`.
+Remark 3.20 [BCGM25]. -/
+theorem remark_3_20
```
alexanderlhicks: **Naming regression.** Previously `theorem poly_gen_is_zero_evading`, now `theorem remark_3_20`. The new name only cites the paper rather than describing content, which conflicts with the descriptive-naming convention documented in `docs/wiki/coding-theory-conventions.md` (e.g., `linear_lambda_ge_elias_volume_eli57`, `rs_epsCA_breakdown_cs25`).

Suggest reverting to `poly_gen_is_zero_evading` (or a similar descriptive form) and keeping the `Remark 3.20 [BCGM25]` citation in the docstring.

Also, the previous version had a `Note:` paragraph explaining the dual-form choice (individual degrees vs total degrees) — that was useful documentation and was removed in this PR. Worth restoring.

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
```diff
@@ -0,0 +1,117 @@
...
+    simp [zeroExtend, j.property]
+  rintro ⟨T, hT₁, hT₂, j, hT₃⟩
+  exact ⟨T, hT₁,
+    by convert hT₂ using 1; exact funext fun _ => by simp [vecMul_projectedGenerator],
+    ⟨j, by rw [zeroExtend_val] ; assumption⟩⟩
+
+/-- Let `G : S → 𝔽^ℓ` be an MCA generator with error `ε_mca`, and `κ` a
+subset of `ℓ`. Then the projected generator over `κ` is an MCA generator with the same error as `G`.
+Corollary 4.2 [BCGM25]. -/
+lemma generatorSubset [Nonempty S] (G : Generator S ℓ F) (ε_mca : I → I) (LC : LinearCode ι F)
```
alexanderlhicks, line 109: Optional: paper proves Cor 4.2 as a direct application of Lemma 4.1 (with `A` = selector matrix), but the Lean proof here takes a different route via `zeroExtend` + `isMCA_projectedGenerator_of_isMCA`. The Lean route is arguably cleaner (no selector matrix to construct) and mathematically correct — but readers comparing to the paper proof may be briefly confused.

Consider a one-line docstring note like: "*This direct proof via `zeroExtend` avoids constructing the selector matrix that the paper's proof uses to invoke Lemma 4.1.*"

alexanderlhicks (commented):
Thorough correctness review against [BCGM26](https://eprint.iacr.org/2025/2051) §4 (pp.22-23).

**Lemma 4.1 (`pseudoinverseGen`)** and **Corollary 4.2 (`generatorSubset`)** are mathematically correct and the proofs follow the paper faithfully. Nice work.

A few things to address before this can land — one correctness concern around `TensorGenerator`, a couple of regressions from #430's state on main, and some cosmetic items. Inline comments below; see also the file-level notes:

- **`TensorGenerator` return type** (correctness) — the def as written is not actually a `Generator`, which will block downstream use (e.g., Lemma 4.4 in the paper).
- **Citation regression** — `Basic/LinearCode.lean` had `Definition 3.7 [BCGM25]` after #430; this PR changes both to `3.6`. Paper p.17 has `3.7` for projection (`3.6` is a Lemma in the paper, not a Definition).
- **Namespace + theorem renames** — `PolynomialGenerator` → `PolyGenIsZeroEvading` and `poly_gen_is_zero_evading` → `remark_3_20` are regressions per ArkLib's descriptive-naming convention. Plus the dual-form `Remark 3.20` note was removed; worth keeping.
- **Lint** — 15 cosmetic warnings on touched files (continuation-indent, one `:` before line break, one space-before-semicolon). Surfaces as `ERR_IND` / `ERR_CLN` / `ERR_SEM` via `./scripts/lint-style.sh`.

L4.1 proof walkthrough notes (for posterity): the contrapositive step uses `projectedCode_linearCombination` with coefficients `B j` (where `B` is the left inverse), yielding `∑ i, B j i * (A·U)_{i,k} = ((B·A)·U)_{j,k} = U_{j,k}` since `B·A = 1`. Algebra matches the paper's `Mat(c'_1,...,c'_ℓ') := A_L · Mat(c_1,...,c_ℓ)` step.

alexanderlhicks:
experimented with posting comments straight from claude code so take with a grain of salt in cases where it might have come up with something that isn't accurate.

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
alexanderlhicks: This suggestion seems right

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
alexanderlhicks: that one is really up to you I think

alexanderlhicks (approved):

PR #493, opened 2026-05-06
fix(BCIKS20): correct modified GS rate bound

ainta:
## Summary
- correct `ModifiedGuruswami.Q_D_Y` to use `(k + 1) / n` instead of `k + 1 / n`
- align the Y-degree bound with the surrounding BCIKS20 Guruswami-Sudan rate bounds

## Non-goals
- this does not migrate the BCIKS20 list-decoding skeleton to the newer `gs_*` APIs using `k / n`; that would require a coordinated refactor of the related bounds

## Tests
- `git diff --check`
- `lake build ArkLib.Data.CodingTheory.ProximityGap.BCIKS20.ListDecoding.Guruswami`
- `lake build ArkLib.Data.CodingTheory.ProximityGap.BCIKS20.ListDecoding.Agreement`


quangvdao (approved):
nice, thanks for the catch!

PR #494, opened 2026-05-07
fix(BCIKS20): correct curve bad-coordinate set

ainta:
## Summary
- correct the bad-coordinate set in `large_agreement_set_on_curve_implies_correlated_agreement` to depend on the coordinate being counted
- count coordinates where some curve coefficient word differs, instead of comparing the global images of the two word families

## Non-goals
- this does not prove the surrounding `sorry` theorem or change the stronger list-decoding curve statement

## Tests
- `git diff --check origin/main...HEAD`
- `/Users/ainta/.elan/bin/lake build ArkLib.Data.CodingTheory.ProximityGap.BCIKS20.Curves`


alexanderlhicks:
Thanks!

alexanderlhicks (approved):

PR #500, opened 2026-05-12
doc(oracle-reduction): add conceptual documentation for Basic.lean

NhlanhlaHasane:
> Added detailed documentation for Oracle Reduction concepts, including Prover, Verifier, and OracleVerifier definitions, as well as interaction, execution, and security layers.


alexanderlhicks:
Thanks for the PR! I don't think these are necessary (there is the wiki and blueprint), inaccuracies and maintenance would also be annoying (the Oracle stuff still changes frequently enough). I'll close this for now.

PR #503, opened 2026-05-13
feat(LogUp): Implemented Logup protocol and its correctness proofs depending on propext, sorryAx, Classical.choice and Quot.sound

jCabala:
Hi,

I'm interested in implementing LogUp in ArkLib. I have been working on the protocol recently and now have an implementation, together with empty correctness proofs. I would be happy to continue working on this if you think it would be useful.

I'm sharing the current state of the work here in the hope of getting some feedback: whether the overall direction looks reasonable, and what the best next steps would be from the perspective of someone more experienced with the library.

Here is a short overview of the files:

- `Protocol.lean` - defines the prover, verifier, and the full LogUp protocol as an ArkLib reduction.
- `Sumcheck/SumcheckPolynomial.lean` - constructs the polynomial used by the prover in the sumcheck phase and proves that it has sufficiently low degree.
- `Sumcheck/SumcheckBridge.lean` - connects that polynomial to ArkLib's existing sumcheck infrastructure.
- `Security/Completeness.lean` and `Security/Soundness.lean` - state the completeness and soundness theorems. The proofs are currently left as sorry.

There is also one TODO about sampling a challenge, which I am slightly unsure how to handle, but I will try to work it out.

I would be very grateful for any feedback. Even a quick indication of whether this is something the project would currently be interested in would already be very helpful.


quangvdao:
Awesome! I will try to take a look soon

ArkLib/ProofSystem/Logup/Security/Completeness.lean
```diff
@@ -0,0 +1,41 @@
+import ArkLib.OracleReduction.Security.Basic
+import ArkLib.ProofSystem.Logup.Protocol
+
+/-!
+# LogUp Completeness
+
+Main completeness statement for Protocol 2 of `paper.txt`.
```
alexanderlhicks: This should be a proper reference (to https://eprint.iacr.org/2022/1530?) rather than pointing to paper.txt
Adding some context / outline of the protocol / highlighting main theorems would also be nice.

ArkLib/ProofSystem/Logup/Protocol.lean
```diff
@@ -0,0 +1,306 @@
...
+  ⟨2, rfl⟩
+
+/-- The verifier for the outer LogUp phase. -/
+noncomputable def outerVerifier :
+    OracleVerifier oSpec (StmtIn F n M) (OStmtIn F n M)
+      (StmtAfterOuter F n M params) (OStmtAfterOuter F n M params)
+      (outerPSpec F n params) where
+  verify := fun _ challenges => do
+    let x : F := challenges (outerChallengeXIdx F n M params)
+  -- TODO: Replace the current table-scan rejection check with a faithful sampler
```
alexanderlhicks: Perhaps this could be rewritten based on remark 3 from Ulrich's paper? It seems you are already modelling the ensuing completeness error in Completeness.lean?

alexanderlhicks:
Seems like Quang didn't end up finding time for this. Let me add some notes.

ArkLib/ProofSystem/Logup/Sumcheck/SumcheckBridge.lean
```diff
@@ -0,0 +1,278 @@
...
+
+variable (F : Type) [Field F] [Fintype F] [DecidableEq F] (n M : ℕ)
+variable (params : ProtocolParams M)
+
+/-- Individual-degree bound for LogUp's embedded sumcheck polynomial. -/
+def logupSumcheckDegree (_params : ProtocolParams M) : ℕ :=
+  M + 3
+
+/-- The concrete ArkLib Sumcheck transcript shape for LogUp's embedded sumcheck. -/
+abbrev logupSumcheckPSpec : ProtocolSpec (Fin.vsum (fun _ : Fin n => 2)) :=
```
alexanderlhicks: This isn't necessarily a blocker, but I'm not convinced these abbrevs relaly increase readability by hiding the underlying sumcheck operations.

ArkLib/ProofSystem/Logup/Sumcheck/SumcheckBridge.lean
```diff
@@ -0,0 +1,278 @@
...
+variable {ι : Type} (oSpec : OracleSpec ι)
+variable (F : Type) [Field F] [Fintype F] [DecidableEq F] [Fact ((-1 : F) ≠ 1)] (n M : ℕ)
+variable (params : ProtocolParams M)
+
+/-- Context lens from LogUp's retained outer state to ArkLib's generic Sumcheck state.
+
+The projection builds the generic zero-sum claim and its single polynomial oracle. The lift drops
+the generic Sumcheck output because the top-level LogUp protocol returns only success/failure.
+-/
+noncomputable def logupSumcheckContextLens :
```
alexanderlhicks: This doesn't seem right. I would expect this to carry the sumcheck output?

ArkLib/ProofSystem/Logup/Sumcheck/SumcheckBridge.lean
```diff
@@ -0,0 +1,278 @@
...
+  toFun := bitToSign F
+  inj' := by
+    intro a b h
+    fin_cases a <;> fin_cases b
+    · rfl
+    · exact absurd h hSigns
+    · exact absurd h.symm hSigns
+    · rfl
+
+private theorem sum_piFinset_map_univ_eq_sum_hypercube
```
alexanderlhicks: this could be reusable so it could be moved to somewhere in Data (or ToMathLib but it's likely arklib specific enough to go in Data).

ArkLib/ProofSystem/Logup/Common.lean
```diff
@@ -0,0 +1,421 @@
...
+@[reducible, simp]
+def OStmtOut : OutputOracleIdx → Type :=
+  fun i => Fin.elim0 i
+
+/-- The full LogUp protocol has no output oracle interfaces to provide. -/
+instance instOStmtOutOracleInterface :
+    ∀ i, OracleInterface (OStmtOut i) :=
+  fun i => Fin.elim0 i
+
+/-- The full protocol has a trivial final relation: successful verification returns `Unit`. -/
```
alexanderlhicks: this is only true if step 4 of construction 2 in Ulrich's paper has been done (see sumcheckbridge comment), and is crucial with respect to soundness and completeness where this appears.

ArkLib/ProofSystem/Logup/Common.lean
```diff
@@ -0,0 +1,421 @@
...
+/-- Public parameter assumptions for Protocol 2.
+
+The paper fixes a finite field with characteristic larger than `M * 2^n`; we also record that
+`-1` and `1` are distinct so that `{±1}^n` is genuinely a Boolean hypercube in the field.
+-/
+structure StmtIn (F : Type u) [Field F] [Fintype F] (n M : ℕ) : Type u where
+  /-- The paper's characteristic condition `char(F) > M * 2^n`. -/
+  charLarge : M * 2 ^ n < ringChar F
+  /-- The two signs used for the hypercube are distinct. -/
+  signsDistinct : (-1 : F) ≠ 1
```
alexanderlhicks: Throughout the other files this is stated as Fact ((-1 : F) ≠ 1, it would be good to be consistent i.e. use Fact throughout (and drop this here) or remove the explicit fact and derive it from the field.

ArkLib/ProofSystem/Logup/Common.lean
```diff
@@ -0,0 +1,421 @@
...
+/-- A Lagrange oracle is a function on `H`; queries ask for inner products with `L_H(., z)`. -/
+structure LagrangeOracle (F : Type u) (n : ℕ) where
+  values : Hypercube n → F
+
+instance {F : Type u} {n : ℕ} : CoeFun (LagrangeOracle F n) (fun _ => Hypercube n → F) where
+  coe oracle := oracle.values
+
+/-- A multilinear oracle over the row variables of `H = {±1}^n`, represented in Lagrange form. -/
+@[reducible]
+def MultilinearOracle (F : Type u) (n : ℕ) : Type u :=
```
alexanderlhicks: similar to abbrevs elsewhere, I don't think this increases readability and doesn't otherwise do anything.

ArkLib/ProofSystem/Logup/Common.lean
```diff
@@ -0,0 +1,421 @@
...
+  sumSize : ℕ
+  /-- Protocol 2 requires `1 ≤ ℓ`. -/
+  sumSize_pos : 0 < sumSize
+  /-- Protocol 2 requires `ℓ ≤ M + 1`. -/
+  sumSize_le : sumSize ≤ M + 1
+
+namespace ProtocolParams
+
+/-- The number of partial-sum groups `K = ceil((M + 1) / ℓ)`. -/
+def numGroups {M : ℕ} (params : ProtocolParams M) : ℕ :=
```
alexanderlhicks: is this always equal to ⌈(M+1)/ℓ⌉? (I'm just staring at it, maybe it is)

ArkLib/ProofSystem/Logup/Common.lean
```diff
@@ -0,0 +1,421 @@
...
+
+/-- Interpret term index `0` as the table and term indices `1, ..., M` as lookup columns. -/
+def termToInput {M : ℕ} (i : TermIdx M) : InputOracleIdx M :=
+  if h : i.val = 0 then
+    .table
+  else
+    .column ⟨i.val - 1, by omega⟩
+
+/-- Interpret input-oracle labels as term indices `0, ..., M`. -/
+def inputToTerm {M : ℕ} : InputOracleIdx M → TermIdx M
```
alexanderlhicks: could add roundtrip lemmas between inputToTerm and termToInput

ArkLib/ProofSystem/Logup/Common.lean
```diff
@@ -0,0 +1,421 @@
+import Mathlib.Algebra.BigOperators.Fin
+import Mathlib.Algebra.CharP.Defs
+import Mathlib.Data.Fin.Basic
+import Mathlib.Data.Finset.Basic
+import Mathlib.Data.Fintype.Basic
+import Mathlib.Tactic
```
alexanderlhicks: Please import only the required tactics (#min_imports should help).

ArkLib/ProofSystem/Logup/Protocol.lean
```diff
@@ -0,0 +1,306 @@
...
+  -- for x ∉ { -t(u) : u ∈ H }.
+    for u in (Finset.univ : Finset (Hypercube n)).toList do
```
alexanderlhicks: if this runs over the entire hypercube it will be in exponential time

ArkLib/ProofSystem/Logup/Common.lean
alexanderlhicks: Please import only the required tactics (#min_imports)

ArkLib/ProofSystem/Logup/Protocol.lean
```diff
@@ -0,0 +1,306 @@
...
+    OracleInterface (((outerPSpec F n params) ++ₚ logupSumcheckPSpec F n M params).Message i)
+  exact ProtocolSpec.instOracleInterfaceMessageAppend
+
+/-- The full LogUp verifier challenges are sampleable: outer LogUp challenges followed by the
+embedded sumcheck challenges. -/
+noncomputable instance instPSpecChallengeSampleable
+    {F : Type} [Field F] [Fintype F] [DecidableEq F] [SampleableType F] {n M : ℕ}
+    {params : ProtocolParams M} :
+    ∀ i, SampleableType ((pSpec F n M params).Challenge i) := by
+  letI : Inhabited F := ⟨0⟩
```
alexanderlhicks: surprising you would need this 🤔

ArkLib/ProofSystem/Logup/Sumcheck/SumcheckPolynomial.lean
```diff
@@ -0,0 +1,416 @@
...
+
+open scoped BigOperators
+
+section SumcheckPolynomial
+
+variable (F : Type) [Field F] (n M : ℕ)
+variable (params : ProtocolParams M)
+
+/-- The `{±1}` Lagrange basis polynomial for one hypercube row. -/
+private noncomputable def signedBasisPolynomial (u : Hypercube n) :
```
alexanderlhicks: seems to duplicate eqpolynomial from https://github.com/Verified-zkEVM/ArkLib/blob/main/ArkLib/Data/MvPolynomial/Multilinear.lean to some extent, it should be possible to integrate with this.

ArkLib/ProofSystem/Logup/Sumcheck/SumcheckPolynomial.lean
```diff
@@ -0,0 +1,416 @@
...
+    MvPolynomial (Fin n) F :=
+  MvPolynomial.C (((2 : F) ^ n)⁻¹) *
+    ∏ j : Fin n, (1 + MvPolynomial.C (bitToSign F (u j)) * MvPolynomial.X j)
+
+/-- Multilinear extension of a function on the signed hypercube `{±1}ⁿ`. -/
+private noncomputable def signedMLEPolynomial (values : Hypercube n → F) :
```
alexanderlhicks: seems to duplicate MLE from https://github.com/Verified-zkEVM/ArkLib/blob/main/ArkLib/Data/MvPolynomial/Multilinear.lean to some extent, it should be possible to integrate with this.

ArkLib/ProofSystem/Logup/Sumcheck/SumcheckPolynomial.lean
```diff
@@ -0,0 +1,416 @@
...
+            MvPolynomial.degreeOf i (MvPolynomial.X j) := by
+          exact MvPolynomial.degreeOf_mul_le i _ _
+        _ ≤ 0 + 0 := by
+          gcongr
+          · exact (MvPolynomial.degreeOf_C (R := F) a i).le
+          · exact (MvPolynomial.degreeOf_X_of_ne (R := F) i j hij).le
+        _ = 0 := by omega
+    exact (MvPolynomial.degreeOf_add_le i _ _).trans (max_le hone hmul)
+
+private theorem signedBasisPolynomial_degreeOf (u : Hypercube n) (i : Fin n) :
```
alexanderlhicks: the computation here seems to partly be a repeat of signedLinearFactor_degreeOf above, this isn't a blocker but perhaps a lemma that makes both easier can be extracted (and possibly reused elsewhere).

ArkLib/ProofSystem/Logup/Security/Soundness.lean
```diff
@@ -0,0 +1,40 @@
...
+
+variable {ι : Type} (oSpec : OracleSpec ι)
+variable (F : Type) [Field F] [Fintype F] [DecidableEq F] [Fact ((-1 : F) ≠ 1)]
+  [SampleableType F]
+variable (n M : ℕ)
+variable (params : ProtocolParams M)
+variable {σ : Type} (init : ProbComp σ) (impl : QueryImpl oSpec (StateT σ ProbComp))
+
+/-- Paper-shaped soundness error for the LogUp outer checks plus the embedded sumcheck error. -/
+noncomputable def logupSoundnessError (F : Type) [Fintype F] (n M : ℕ) (params : ProtocolParams M)
```
alexanderlhicks: It could be cleaner to prove each part of the soundness error separately.

ArkLib/ProofSystem/Logup/Security/Soundness.lean
```diff
@@ -0,0 +1,40 @@
...
+    (sumcheckSoundnessError : ℝ≥0) : ℝ≥0 :=
+  ((((M + 1) * Fintype.card (Hypercube n) - 1 : ℕ) : ℝ≥0) /
+      ((Fintype.card F - Fintype.card (Hypercube n) : ℕ) : ℝ≥0)) +
```
alexanderlhicks: careful about division by 0

ArkLib/ProofSystem/Logup/Security/Completeness.lean
```diff
@@ -0,0 +1,41 @@
...
+
+/-- Completeness error forced by the current rejection-based model of the `x` challenge.
+
+Protocol 2 samples `x` from the complement of the table poles. The current verifier samples from
+all of `F` and rejects poles, so the intended completeness statement carries this explicit bad-`x`
+probability. Once the challenge sampler is modeled as the complement distribution, this should
+collapse to perfect completeness.
+-/
+noncomputable def logupCompletenessError (F : Type) [Fintype F] (n : ℕ) : ℝ≥0 :=
+  (Fintype.card (Hypercube n) : ℝ≥0) / (Fintype.card F)
```
alexanderlhicks: just flagging this since there was some inconsistency elsewhere about whether you were going with perfect completeness or not -- completeness error is fine but should be consistent across the formalization.

alexanderlhicks:
Thanks for the PR and sorry for the wait.
I think there are some issues the need to be resolved before we can think of merging this. It would be nice to consider the following beyond the inline comments:
- Making this friendly to further logup related work (logup*, logup GKR). At the very lest clearer docstrings to allow others to correctly reuse the work here if it is merged would be good.
- Lemmas from the paper which are currently not in the PR are would be useful e.g. lemma 4/5 at the very least, but broadly speaking I don't see a reason to not provide coverage of the paper as a whole.
- Make sure you are using the pre-existing sumcheck reduction and oracle/lens abstractions as well as possible.

Maybe this can be improved as is, but it's possible that a fresh PR after taking another look at things might work out too, as well as making the spec computable and sorry-free.

jCabala:
Thanks for the very thorough review and feedback! I’ll work through all of it and ping you again once I have a version that I think is ready for another review.

I’ll keep working on this PR unless the changes end up diverging too much and it makes more sense to start from scratch. In that case, I’ll close this one and open a cleaner PR.

Thanks again!

jCabala:
Hi @alexanderlhicks ,

I’ve been working through your suggestions and pushing the PR forward more generally. The PR has changed quite a bit as a result, so here is a summary of where it has landed.

* **Inline comments.** I believe I have addressed all of them, apart from the remarks about the soundness proof. I have been resolving threads as I went.

* **Reusability and docstrings.** This was a major focus. I hope the docstrings are more helpful now. I also restructured the files into a coarser layout: `Protocol.lean` contains the protocol, `Algebra.lean` contains the underlying mathematics, and the security results live under `/Security`. I am trying to keep `Algebra.lean` decoupled from the protocol types so that the algebra can be reused independently. I also removed most of the abbreviations; the remaining ones should be definitions that genuinely help readability, but please let me know if you feel they still are not helpful.

* **Paper coverage (Lemmas 4 and 5).** Both are now formalized in a paper-shaped form. I chose to state them at the ring level over `F[X]` rather than using `RatFunc`. In particular, `clearedDecomp_injective` corresponds to Lemma 4, giving uniqueness of fractional decompositions, and `setInclusion_iff_cleared` corresponds to Lemma 5, giving the two-direction set-inclusion criterion. I also added `setInclusion_eval_forward`, which is the evaluated forward direction used by completeness. Lemma 4 and the converse direction of Lemma 5 are intended to support soundness.

* **Protocol logic.** I removed the redundant verifier-side checks. The verifier no longer scans the table for poles or ranges over the hypercube; instead, the formalization simply carries the resulting nonzero completeness error. I also added a final verification stage that queries the retained oracles at sumcheck’s final point and checks that the output relation holds.

* **Reusing the sumcheck / oracle / lens machinery.** The embedded sumcheck phase now uses the framework’s generic sumcheck reduction, connected to LogUp’s state through the standard lens. The full protocol is also composed using the standard sequential-composition operators. I additionally replaced my polynomial helpers with ArkLib’s existing multilinear-extension machinery.

* **Completeness proof.** Completeness is now proved.

I also briefly looked at the two related protocols to understand how much of this groundwork might carry over to the broader LogUp family.

* **LogUp*.** The protocol seems substantially different from this one, so beyond some shared algebra, mainly the fractional-decomposition / logarithmic-derivative machinery in `Algebra.lean` , I do not currently see much room for direct reuse.

* **LogUpGKR.** Reuse looks a little more plausible. The outer phase could share a few messages with this LogUp implementation. More similarities may become apparent during implementation, although having a GKR spec in place first would probably make that much clearer.

As part of making the code reusable, I am therefore trying to keep the algebra decoupled from the protocol types, since that seems like the layer most likely to be shared across the family.

One design question I would appreciate your thoughts on: I implemented the final check inside the reduction. An alternative would be to mirror generic sumcheck more closely: the protocol would output the final claim, and the final equality check would be pushed into the output relation rather than enforced inside the protocol. Do you think that would be the better design?

Please feel free to take another look whenever convenient. In the meantime, I think I will start looking into soundness. Any feedback would be very much appreciated.

alexanderlhicks:
Thanks! Sorry I missed the notification earlier, I'll get to this asap.

ArkLib/ProofSystem/Logup/Security/Soundness.lean
```diff
@@ -0,0 +1,2190 @@
...
+    (OracleVerifier.append_soundness
+      (lang₂ := (logupMidRelation F n M params).language) _ _ ?_ ?_) ?_
+  · exact logup_outer_soundness oSpec F n M params init impl hcard
+  · exact logup_sumcheck_soundness oSpec F n M params init impl sumcheckSoundnessError
+      hSumcheckSoundness
+  · exact logup_finalCheck_soundness oSpec F n M params init impl
+
+end Soundness
+
+end Logup
```
alexanderlhicks, line 511: 🤖 _AI-generated review comment (Claude), curated before posting — treat as a suggestion._

**CI blocker — file length.** `Soundness.lean` is 2190 lines; ArkLib's long-file linter caps at 1500, so `validate.sh --lint` / CI fails here:
```
Soundness.lean:2191: This file is 2191 lines long, but the limit is 1500.
```
Either split the `## Helper Lemmas` block into e.g. `Logup/Security/Soundness/Lemmas.lean` (preferred), or add a local opt-out at the top of the file:
```lean
set_option linter.style.longFile 2300
```

ArkLib/ProofSystem/Logup/Security/Soundness.lean
```diff
@@ -0,0 +1,2190 @@
...
+
+end
+
+section
+
+omit [DecidableEq F] [SampleableType F]
+/-- Prover messages preserve failure of the outer bad-event invariant.
+
+The round-by-round soundness state only changes on verifier challenge rounds.  When the next round
+is prover-to-verifier, adding the prover message cannot move a transcript back into the invariant.-/
```
alexanderlhicks: 🤖 _AI-generated review comment (Claude), curated before posting — treat as a suggestion._

**CI lint — doc-string terminator.** This docstring ends with `…invariant.-/`; the style linter requires a space or newline before the closing `-/`:
```
Soundness.lean:1120:97: doc-strings should end with a single space or newline
```
Fix: `…invariant. -/` (or put `-/` on its own line).

ArkLib/ProofSystem/Logup/Security/Soundness.lean
```diff
@@ -0,0 +1,2190 @@
...
+private theorem outerSoundnessState_full_prob_zero
+    (stmtPair : StmtIn F n M × (∀ i, OStmtIn F n M i))
+    (tr : (outerPSpec F n params).FullTranscript)
+    (hfalse :
+      ¬ outerSoundnessState (F := F) (n := n) (M := M) (params := params)
+        stmtPair (Fin.last 4) tr) :
+    Pr[(· ∈ (logupMidRelation F n M params).language) |
+      OptionT.mk do
+        (simulateQ impl
+          (((outerVerifier oSpec F n M params).toVerifier).run stmtPair tr)).run' (← init)] = 0 := by
```
alexanderlhicks: 🤖 _AI-generated review comment (Claude), curated before posting — treat as a suggestion._

**CI lint — line length.** This line exceeds the 100-character limit (`longLine`). Wrap it, e.g. break before `] = 0 := by`.

ArkLib/ProofSystem/Logup/Security/Soundness.lean
```diff
@@ -0,0 +1,2190 @@
...
+          have hq :
+              simulateQ
+                (QueryImpl.liftTarget (StateT σ ProbComp)
+                  (ProtocolSpec.challengeQueryImpl (pSpec := outerPSpec F n params)))
+                (liftM (OracleSpec.query qIn) :
+                  OracleComp ([(outerPSpec F n params).Challenge]ₒ'
+                    ProtocolSpec.challengeOracleInterface) F) =
+              (liftM ($ᵗ F) : StateT σ ProbComp F) := by
+            rw [simulateQ_query]
+            simp [qIn, ProtocolSpec.challengeQueryImpl, QueryImpl.liftTarget_apply,
```
alexanderlhicks: 🤖 _AI-generated review comment (Claude), curated before posting — treat as a suggestion._

**Lint (flexible `simp`).** This `simp [qIn, …]` is flagged as a *flexible* tactic modifying the goal; `simp?` suggests pinning it:
```lean
simp only [ProtocolSpec.ChallengeIdx, Nat.reduceAdd, Fin.vcons_fin_zero, MultiplicityMessage,
  HelperMessages, BatchingChallenge, ProtocolSpec.Challenge, OracleQuery.input_query,
  OracleQuery.cont_query, QueryImpl.liftTarget_apply]
```
The same pattern appears at a couple of other spots in this file — worth a sweep. Minor, but keeps CI green and the proof robust to `simp`-set drift.

ArkLib/ProofSystem/Logup/Security/Soundness.lean
```diff
@@ -0,0 +1,2190 @@
...
+# LogUp Soundness
+
+Soundness target for the LogUp lookup argument (Cryptology ePrint Archive, Paper 2022/1530,
+<https://eprint.iacr.org/2022/1530>).
+
+The protocol verifier is the sequential composition of three phases (outer LogUp, embedded
+sumcheck, final point check), so its soundness error decomposes as a sum of one error per phase.
+We bound each phase separately and combine them with `OracleVerifier.append_soundness`, which
+turns the soundness of a composed verifier into the sum of the parts' errors. This matches the
+paper's Theorem 4, where the total error is `ε₁ + ε₂ + ε₃ + εsumcheck`.
```
alexanderlhicks: 🤖 _AI-generated review comment (Claude), curated before posting — treat as a suggestion._

**Docstring accuracy (soundness-safe).** "This matches the paper's Theorem 4" slightly overstates it. The outer error (lines 52–55) splits the Schwartz–Zippel term into *two* full `(M+1)·2ⁿ/|F|` summands (poles + roots) where Theorem 4 has a single `(|H|(M+1)−1)/|F|`, and `x` is sampled from all of `F` rather than `F∖(−t(H))` (paper Remark 3). So the formal bound has the *same per-phase shape* but is deliberately *coarser/larger* than Theorem 4. Suggest wording like: "same structure as Theorem 4; the formal bound is a (looser) upper bound."

ArkLib/ProofSystem/Logup/Security/Soundness.lean
```diff
@@ -0,0 +1,2190 @@
...
+noncomputable def logupOuterSoundnessError (F : Type) [Fintype F] (n M : ℕ)
+    (params : ProtocolParams M) : ℝ≥0 :=
+  ((((M + 1) * Fintype.card (Fin n → Fin 2) : ℕ) : ℝ≥0) /
+      (Fintype.card F : ℝ≥0)) +
+    ((((M + 1) * Fintype.card (Fin n → Fin 2) - 1 : ℕ) : ℝ≥0) /
+      (Fintype.card F : ℝ≥0)) +
+    (((params.numGroups * n : ℕ) : ℝ≥0) / (Fintype.card F : ℝ≥0)) +
+    ((1 : ℕ) : ℝ≥0) / (Fintype.card F : ℝ≥0)
+
+/-- Soundness error of the final LogUp point check (paper's `ε₂ = K/|F|`).
```
alexanderlhicks: 🤖 _AI-generated review comment (Claude), curated before posting — treat as a suggestion._

**Label / attribution (soundness-safe).** Two things about `logupFinalCheckSoundnessError = K/|F|`:
1. The final check is deterministic — `finalCheckPSpec : ProtocolSpec 0` (no messages, no challenges), as your own comment at line 1965 notes — so its *true* soundness error is 0. Charging `K/|F|` is harmless slack, but it is slack.
2. The docstring calls this "paper's ε₂ (reducing the domain identities to the Lagrange-kernel point evaluation)", but that reduction is already realized in the **outer** error as the `K·n/|F|` term (line 56, `params.numGroups * n`). So ε₂ is effectively attributed twice.

Suggest: relabel this as conservative slack for the deterministic check (or set it to 0), and move the "ε₂ / Lagrange-kernel reduction" wording to the line-56 term where it actually lives.

ArkLib/ProofSystem/Logup/Algebra.lean
```diff
@@ -0,0 +1,1964 @@
...
+/-- The multiplicity of a value `z` in a finite sequence `a : ι → F`. -/
+def seqMultiplicity {ι : Type*} [Fintype ι] (a : ι → F) (z : F) : ℕ :=
+  (Finset.univ.filter fun i => a i = z).card
+
+/-- **Lemma 5** (set inclusion), cleared-denominator form. When the field characteristic exceeds the
+sizes of the index types, the set underlying `a` is contained in that of `b` iff there exist
+multiplicities `m` making the cleared logarithmic-derivative identity hold in `F[X]` (the cleared
+form of `∑_i 1/(X + a i) = ∑_j m j/(X + b j)`). The honest witness is the normalized multiplicity
+`m j = ma(b j)/mb(b j)`. Generalizes the paper's equal-length sequences to arbitrary index types. -/
+theorem setInclusion_iff_cleared {ι κ : Type*} [Fintype ι] [Fintype κ]
```
alexanderlhicks: 🤖 _AI-generated review comment (Claude), curated before posting — treat as a suggestion._

**Dead lemma.** `setInclusion_iff_cleared` has no consumers — the only mention is prose in a docstring (line 368); the security proofs use `setInclusion_eval_forward` and `clearedOccurrences_ne_zero_of_fiber_sum_ne_zero` instead. If it's kept as a faithful rendering of the paper's Lemma 5 (a fine reason), add a one-line docstring saying so ("kept as a faithful statement of Lemma 5; not load-bearing"); otherwise consider removing it to cut surface area.

ArkLib/OracleReduction/Composition/Sequential/General.lean
```diff
@@ -24,6 +25,14 @@ variable {ι : Type} {oSpec : OracleSpec ι}
 
 section Composition
 
+theorem mem_support_liftM_oracleComp {ι τ : Type} {spec : OracleSpec ι}
```
alexanderlhicks, line 32: 🤖 _AI-generated review comment (Claude), curated before posting — treat as a suggestion._

**API placement (shared file).** `mem_support_liftM_oracleComp` is declared in the root namespace with no docstring, but it's a thin wrapper around `OracleComp.mem_support_of_mem_support_liftComp` and is conceptually an `OracleComp` lemma. Since this lands in the shared composition file, consider moving it under `namespace OracleComp` (beside the lemma it wraps) and adding a short docstring — better discoverability, and it keeps the top-level namespace uncluttered.

ArkLib/OracleReduction/Composition/Sequential/General.lean
```diff
@@ -60,6 +69,78 @@ lemma seqCompose_succ {m : ℕ}
     seqCompose Stmt Wit P =
       append (P 0) (seqCompose (Stmt ∘ Fin.succ) (Wit ∘ Fin.succ) (fun i => P (Fin.succ i))) := rfl
 
+/-- If every prover in a sequential composition preserves a projection of the statement, then the
+whole composed prover preserves that projection. -/
+theorem seqCompose_preserves {m : ℕ} :
```
alexanderlhicks, line 80: 🤖 _AI-generated review comment (Claude), curated before posting — treat as a suggestion._

**Optional generality (shared file).** `Prover.seqCompose_preserves` hardcodes the witness family to `fun _ => Unit`. That's exactly what the LogUp consumer needs, so this is fine to merge as-is — but since it's landing in the general composition library, generalizing over an arbitrary witness family would make it reusable beyond this protocol. Non-blocking; a maintainer call on intended library scope.

alexanderlhicks (commented):
## 🤖 AI-generated review — LogUp / Haböck "Protocol 2" (ePrint 2022/1530)

_This review was generated by an AI assistant (Claude) and curated before posting. Please treat the points below as suggestions, not an authoritative verdict._

Thanks for the substantial rework — I reviewed this adversarially (per-file + cross-checked against the paper) and also did a full build. Headline:

**The mathematics is real, faithful, and non-vacuous.** I specifically hunted the failure modes that tend to bite security formalizations and none of them stick:
- `inputRelation` (Protocol.lean:244) is the genuine set-inclusion lookup relation and is actually consumed.
- `outputRelation = Set.univ` (Protocol.lean:634) is **not** a vacuity bug — soundness is only measured over accepting (`some`) runs, so the bad event is exactly "verifier accepts a false statement".
- The outer round-by-round bad-event machine `outerSoundnessState` has no tautological (`else True`) branch, and its `rbrSoundness` is proved without `sorry`.
- Faithful to Protocol 2: Lemmas 4/5 carry the char `p > M·|H|` hypothesis, the batched `Q` polynomial and degree `M+3` match Theorem 4, and the `{±1}ⁿ → hypercube` change is handled honestly.
- Every file under `ArkLib/ProofSystem/Logup/**` is `sorry`-free (build-confirmed — no "declaration uses sorry" warnings for any Logup file).

**On the axiom status — this is _not_ on you.** A full build (3595/3595, toolchain v4.30.0) plus `#print axioms` reports:
```
'Logup.logup_soundness'    depends on axioms: [propext, sorryAx, Classical.choice, Quot.sound]
'Logup.logup_completeness' depends on axioms: [propext, sorryAx, Classical.choice, Quot.sound]
```
The `sorryAx` is **entirely inherited** from pre-existing framework lemmas this PR does not touch: `Verifier.append_soundness` / `Reduction.append_completeness` (`Composition/Sequential/Append.lean`), `rbrSoundness_implies_soundness` + `rbrKnowledgeSoundness_implies_rbrSoundness` (`Security/Implications.lean`), the generic sumcheck `oracleVerifier_rbrKnowledgeSoundness` (`Sumcheck/Spec/SingleRound.lean`), and the `liftContext_*` family (`Security/Reduction.lean`). Your own additions — including the three new sumcheck lemmas — are `sorry`-free. So `logup_soundness`/`logup_completeness` will become axiom-clean **automatically** once those upstream lemmas are closed; nothing to fix here on the math.

The only ask on this point is **framing**: the PR currently reads as a finished end-to-end proof. A short note (module docstring and/or PR description) stating that the composed theorems are *contingent* on the named framework `sorry`s — ideally pasting the `#print axioms` output — keeps future readers from being misled.

The inline comments are the concrete, in-scope items: a few CI-lint blockers, and some documentation-accuracy nits (all soundness-safe). Nothing math-breaking. Nice work.

ArkLib/OracleReduction/Composition/Sequential/General.lean
jCabala: Added to namespace.

ArkLib/ProofSystem/Logup/Security/Soundness.lean
jCabala: Fixed

ArkLib/ProofSystem/Logup/Security/Soundness.lean
jCabala: Fixed

ArkLib/ProofSystem/Logup/Security/Soundness.lean
jCabala: fixed

ArkLib/ProofSystem/Logup/Security/Soundness.lean
jCabala: Set logupFinalCheckSoundnessError to 0 and moved the paper’s ε₂ / Lagrange-kernel explanation to the outer-phase error, where that random point is actually sampled.

ArkLib/ProofSystem/Logup/Algebra.lean
jCabala: Kept and added a comment

ArkLib/OracleReduction/Composition/Sequential/General.lean
jCabala: Done: generalized Prover.seqCompose_preserves over an arbitrary witness family and updated the Sumcheck/LogUp consumer to instantiate it with fun _ => Unit.

ArkLib/ProofSystem/Logup/Security/Soundness.lean
jCabala: Fixed

ArkLib/ProofSystem/Logup/Security/Soundness.lean
jCabala: I splitted it into 3 files.
Did a similar split for Algebra as it was also too big.

jCabala:
Thanks, @alexanderlhicks, for the additional feedback! I’ve implemented all of your suggestions and replied to each of the conversations you opened before resolving them.

The CI is currently failing because of unrelated documentation and blueprint issues: a missing BibTeX key (`NOZ26`) and stale non-LogUp declarations in `blueprint/lean_decls`. The failures are not related to the LogUp code.

ArkLib/OracleReduction/Composition/Sequential/General.lean
```diff
@@ -91,6 +191,58 @@ lemma seqCompose_succ {m : ℕ} (Stmt : Fin (m + 2) → Type)
...
+      have h' : out ∈ support (((do
+          let stmt₂ ← (V 0).run stmt trApp.fst
+          let stmt₃ ← tail.run stmt₂ trApp.snd
+          return stmt₃) : OptionT (OracleComp oSpec) (Stmt (Fin.last (m + 1))))) := by
+        change out ∈ support ((Verifier.append (V 0) tail).run stmt trApp) at h
+        rw [Verifier.append_run] at h
+        exact h
+      rw [mem_support_bind_iff] at h'
+      rcases h' with ⟨stmt₂, h₁, hrest⟩
+      rw [mem_support_bind_iff] at hrest
```
alexanderlhicks, line 232: This rewrite fails on the submitted head. The preceding `rw [mem_support_bind_iff] at hrest` has already simplified the trailing `tail.run … >>= pure`, leaving `hrest` directly as membership in `support (tail.run …)`, so there is no remaining bind to rewrite here. Removing this second decomposition and using the existing `hrest` makes this theorem compile under the PR's pinned Lean version. Please also run the full build afterward, since this failure currently masks several later errors.

ArkLib/ProofSystem/Sumcheck/Spec/SingleRound.lean
```diff
@@ -1045,6 +1085,53 @@ lemma oracleReduction_verifier_eq_verifier {i : Fin n} :
     (oracleReduction R n deg D oSpec i).verifier = oracleVerifier R n deg D oSpec i := by
   rfl
 
+omit [SampleableType R] in
```
alexanderlhicks, line 1038: `SampleableType R` cannot be omitted from this theorem: the type of `h` refers to `oracleReduction`, whose construction requires that instance. After fixing the earlier sequential-composition error, Lean reports `cannot omit referenced section variable inst` here. Please remove this `omit` wrapper. The verifier-only preservation theorem below can remain independent of `SampleableType`.

ArkLib/ProofSystem/Sumcheck/Spec/General.lean
```diff
@@ -199,6 +199,86 @@ lemma oracleReduction_verifier_eq_oracleVerifier :
     (oracleReduction R deg D n oSpec).verifier = oracleVerifier R deg D n oSpec := by
   rfl
 
+omit [SampleableType R] in
```
alexanderlhicks, line 202: This has the same issue as the single-round prover theorem: the theorem statement contains `oracleReduction`, so `SampleableType R` is genuinely referenced and cannot be omitted. Removing this wrapper, while retaining the wrapper on the verifier theorem below, allows this module to elaborate.

ArkLib/ProofSystem/Logup/Protocol.lean
```diff
@@ -0,0 +1,816 @@
...
+  verify := fun _ challenges => do
+    let x : F := challenges (outerChallengeXIdx F n M params)
+    -- Following Remark 3 of the LogUp paper, the verifier samples `x` uniformly and does not
+    -- scan the table to reject poles. Pole challenges are treated as bad/failing inputs for the
+    -- honest handoff, and `Completeness.lean` accounts for that event.
+    let batch : BatchingChallenge F n params.numGroups :=
+      challenges (outerChallengeBatchIdx F n M params)
+    pure { xChallenge := x, zChallenge := batch.1, batchingScalars := batch.2 }
+
+  embed :=
```
alexanderlhicks, line 389: This construction was not migrated to the `OracleVerifier` API brought in by the recent merge from `main`. `embed` and `hEq` are no longer top-level verifier fields; the verifier now requires an `outputOracle`, normally an `.inl` `OracleOutputEmbedding` here, including its `outputInterface_heq` proof. Once the earlier masked errors are fixed, this currently produces “`embed` is not a field” and “Fields missing: `outputOracle`”. The final-check verifier has the same problem.

ArkLib/ProofSystem/Logup/Protocol.lean
```diff
@@ -0,0 +1,816 @@
...
+variable (F : Type) [Field F] [Fintype F] [DecidableEq F] (n M : ℕ)
+variable (params : ProtocolParams M)
+
+/-- Context lens from LogUp's retained outer state to ArkLib's generic Sumcheck state.
+
+The projection builds the generic zero-sum claim and its single polynomial oracle. The lift keeps
+the outer LogUp data and retained oracles together with Sumcheck's final point claim, so the next
+LogUp phase can perform the paper's final oracle-query check.
+-/
+noncomputable def logupSumcheckContextLens :
```
alexanderlhicks, line 533: The current `liftContext` APIs require an `OracleContext.ExecutableLens`/`OracleStatement.ExecutableLens`, but this remains an ordinary `OracleContext.Lens`. Consequently, all three uses below—prover, verifier, and reduction lifting—fail after the first CI error is repaired. Please provide executable input-oracle simulation and the required output-materialization contract rather than relying on the pre-merge lens API.

ArkLib/ProofSystem/Logup/Protocol.lean
```diff
@@ -0,0 +1,816 @@
...
+    let columnValues ← (Vector.finRange M).mapM
+      (fun i => finalCheckQuery oSpec F n M params (.input (.column i)) r)
+    let helperValues ← (Vector.finRange params.numGroups).mapM
+      (fun k => finalCheckQuery oSpec F n M params .helpers ⟨k, r⟩)
+    guard (qAtPoint (params.group) stmt.outer.xChallenge stmt.outer.zChallenge r
+      stmt.outer.batchingScalars multiplicity table (fun i => columnValues[i])
+      (fun k => helperValues[k]) = expectedValue)
+    pure ()
+
+  embed :=
```
alexanderlhicks, line 695: As above, `embed` and `hEq` are no longer fields of `OracleVerifier`. Even though the output oracle family is empty here, this still needs an `outputOracle := .inl { … }` value with the required coherence proof. This is another masked compilation error introduced by the incomplete merge/API migration.

ArkLib/ProofSystem/Logup/Protocol.lean
```diff
@@ -0,0 +1,816 @@
...
+bridge packaging `Q` for generic sumcheck, the context lens, the lifted prover/verifier, and
+the assembled oracle reduction. -/
+
+section Phase2
+
+variable (F : Type) [Field F] [Fintype F] [DecidableEq F] (n M : ℕ)
+variable (params : ProtocolParams M)
+
+/-- Individual-degree bound for LogUp's embedded sumcheck polynomial. -/
+def logupSumcheckDegree (_params : ProtocolParams M) : ℕ :=
```
alexanderlhicks, line 433: Why is the grouping parameter intentionally ignored here? Haböck §3.3 gives degree `ℓ + 2`, where `ℓ` is the group size; that is the commitment-size versus Sumcheck-cost tradeoff represented by `sumSize`. `M + 3` is a sound conservative bound, but it makes the protocol pay the worst-case Sumcheck degree even when small groups were requested. Please either derive the bound from the maximum group cardinality/`sumSize`, or document explicitly that this formalization deliberately uses the paper's coarse worst-case bound.

ArkLib/ProofSystem/Logup/Security/Soundness/Lemmas.lean
```diff
@@ -0,0 +1,1411 @@
...
+    · exact Finset.mem_insert.mpr
+        (Or.inr (Finset.mem_singleton.mpr (Subtype.ext rfl)))
+  · intro _
+    exact Finset.mem_univ i
+
+end
+
+/-- Protocol-level bridge for the outer phase: the local algebraic ingredients above imply the
+conservative scan-free outer soundness bound used by `logupOuterSoundnessError`. -/
+theorem logup_outer_soundness_from_local_algebra
```
alexanderlhicks, line 953: `_hcard`, `_hClearedDegree`, and `_hBadRoots` are not used by this bridge. In particular, the proof later calls the concrete `clearedLookupIdentity_bad_x_prob_le` theorem directly instead of deriving the bound from the supplied degree/root hypotheses. This makes the abstraction misleading, and it also leaves the public theorem's paper-shaped `hcard` premise decorative. Please either make the proof consume these hypotheses or remove them and state clearly that this is a concrete, unconditional union-bound theorem.

ArkLib/ProofSystem/Logup/Algebra/Lemmas.lean
```diff
@@ -0,0 +1,1419 @@
...
+              ∏ b ∈ (Finset.univ.erase a), (Polynomial.X + Polynomial.C (value b)))).coeff
+            (fiber.card - 1) =
+          if value a = z then coeff a * ∏ b ∈ rest, shifted b else 0 := by
+    intro a
+    have htaylor :
+        Polynomial.taylor (-z)
+            (Polynomial.C (coeff a) *
+              ∏ b ∈ (Finset.univ.erase a), (Polynomial.X + Polynomial.C (value b))) =
+          Polynomial.C (coeff a) *
+            ∏ b ∈ (Finset.univ.erase a), (Polynomial.X + Polynomial.C (shifted b)) := by
```
alexanderlhicks, line 544: This `simp` proof no longer closes after rebasing onto current `main`/Lean 4.33. The merge is textually clean but leaves an unsolved Taylor-algebra goal here. Please reprove this with the current Taylor API—preferably using a more explicit `simp only`/calculation—so the result is not dependent on the older simplifier behavior.

alexanderlhicks (changes requested):
The LogUp mathematics looks substantially sound against Haböck's paper, and the earlier review comments were mostly addressed. However, the submitted head does not build: the current CI failure masks additional Sumcheck and oracle-API migration errors. Please rebase/migrate the branch, restore a clean build, and run the full validation and axiom gates before requesting another review.

alexanderlhicks:
We are preparing a current-main sequential-composition consolidation of #635/#636/#643/#846, followed by remaining-caller migration and retirement of the false completeness contracts.

A concrete compatibility point in this PR is Prover.seqCompose_preserves: its explicit application of the old unrestricted Prover.append_run needs adaptation. The proved execution APIs use pure left output or an empty/message-opening seam; we will preserve a more general simulated-factorization completeness entry point as well. A support-level argument may preserve the projection theorem's broader contract, but support equality must not be used to assert raw program equality. Verifier.seqCompose_preserves should not acquire the prover purity restriction.

The two quantitative LogUp oracle-completeness calls remain part of the downstream migration inventory if this PR lands before that work. The consolidation does not by itself discharge LogUp's independent probability/security dependencies. I will post the finalized API/PR links after independent review and integrated validation.

alexanderlhicks:
The validated composition consolidation is https://github.com/Verified-zkEVM/ArkLib/pull/885, with caller migration and retirement in https://github.com/Verified-zkEVM/ArkLib/pull/887.

The old unrestricted completeness contracts are removed in #887. Guarded binary and finite-chain oracle APIs now make the verifier form, prover seam/output purity, and state-uniform suffix obligations explicit; all current-main callers are migrated without changing their maintained public theorem types.

This PR's projection-preservation compatibility repair remains separate: its prover proof cannot use the old unrestricted `Prover.append_run`. It needs a justified purity/seam hypothesis at that use, or an independently proved support-level factorization; raw program equality does not follow merely from equal supports. The verifier-only projection result can stay unrestricted. LogUp's independent completeness/security/lifting obligations are not automatically discharged by the new APIs.

PR #504, opened 2026-05-14
Life improvement membership lemmas for RS codes

ElijahVlasov:
Life gets much easier when one has nice lemmas to prove RS-membership without tears


alexanderlhicks (approved):

alexanderlhicks:
Thanks!

PR #505, opened 2026-05-15
Feat/abf26 plan

alexanderlhicks:
Got around to trying to provide full coverage of https://eprint.iacr.org/2026/680 here. Some slop involved and unreviewed beyond reference based guardrails. Ongoing (but progressing) effort to force claude to make use of ArkLib definitions/notations/conventions so that the final PR should integrate things nicely without causing friction or duplicated material -- partially an experiment to see if it is possible to get a standard LLM to do this well enough really. Some sorrys remaining.


quangvdao:
AI-authored comment from Codex (GPT-5), posted by the assistant on behalf of Quang Dao.

I think there is a small theorem/docstring mismatch in `ArkLib/Data/CodingTheory/ProximityGap/CapacityBounds.lean` around `rs_epsCA_johnson_jump_bchks25`.

The ABF26 T4.18 docstring says the claim is uniform: “for all `F` of characteristic 2 there exists a Reed-Solomon code...”. But the Lean statement currently existentially quantifies the field:

```lean
∃ ... (FC : Type) (_ : Field FC) (_ : Fintype FC) (_ : DecidableEq FC),
  CharP FC 2 ∧ ...
```

That proves existence of some characteristic-2 field, which is weaker than the stated theorem. Minimal fix would be to move `FC` into the theorem parameters, e.g.

```lean
theorem rs_epsCA_johnson_jump_bchks25
    {FC : Type} [Field FC] [Fintype FC] [DecidableEq FC] [CharP FC 2]
    (ε : ℝ≥0) (_hε : 0 < ε) :
    ∃ (ιC : Type) ...
```

and then existentially quantify only the code index/domain/rate data.

alexanderlhicks:
Thanks @quangvdao (and Codex)! Fixed in b50d9490 — `FC` and its `[Field FC] [Fintype FC] [DecidableEq FC] [CharP FC 2]` instances moved to theorem parameters (∀-quantified), keeping only the code-index/domain/rate data existentially quantified. Build clean.

While I was there I noticed T4.16 (`rs_epsCA_lower_capacity_bchks25_kk25`, two theorems up) has the opposite shape — its docstring genuinely is ∃F (an existence claim about codes), so the quantifier structure is fine, but Lean doesn't currently encode the "prime field F" / "|F| = poly(n)" / "smooth domain L" docstring claims as constraints inside the existential. Left for a separate fix since it has more moving parts than your flagged one.

quangvdao:
Thanks, this is looking useful. I think there are two statement-level issues to tighten before merge, plus one generated-file hygiene fix.

1. In `ArkLib/ProofSystem/ToyProblem/SoundnessBounds.lean`, the L6.12/L6.13 docstrings say the witnesses `(f₁, f₂)` lie outside `R̃²_{C,δ}`, but the theorem conclusions only assert existence of witnesses with a large `winningSet`. Minimal fix: add the missing outside-relation conjunct, roughly
   `¬ ToyProblem.relaxedRelation (k := k) (ℓ := 2) C δ v ![μ₁, μ₂] ![f₁, f₂]`,
   to both existential conclusions, or weaken the docstrings if that property is intentionally deferred.

2. `ToyProblem.relation` existentially chooses an encoder, while `accepts` / `verifier` use an externally supplied encoder. That lets the relation and protocol semantics talk about different encoders. Minimal fix: parameterize the relation/input/output relations by the fixed encoder, or require `C = Set.range encode`.

Also, `git diff --check origin/main..origin/feat/abf26-plan` reports trailing whitespace in generated `docs/kb/_generated/dedup-report.md`. This looks generated from unstripped doc snippets in `scripts/kb/find_dedup_candidates.py`; `.rstrip()` those snippets and regenerate should be enough.

Disclosure: posted by the assistant on behalf of Quang Dao. AI-authored using GPT-5.

alexanderlhicks:
Yeah the toy problem is currently very much WIP. It took some manual intervention for it to start using the oracle stuff in a reasonable way.

quangvdao:
AI-authored review from Claude (Fable 5), posted by the assistant on behalf of Quang Dao.

Full faithfulness pass against ePrint 2026/680 (all 54 pages) **and** the current `ef-millenium.tex`, plus a local build of the branch (clean, 4166 jobs). Overall: the coverage effort and audit discipline are genuinely good — the proven material I checked (T3.2, L3.7, App C except KKH26 Lem 1, L6.8/L6.12/L6.13, Fact 4.5, Claim B.1) is faithful, and both items from the earlier review rounds are verified fixed. But four **admitted (`sorry`'d) statements are provably false as stated**, which is the failure mode external admits most need protection from, plus two CI blockers and some stale source-tracking.

## Critical — false-as-stated admits

1. **`frs_list_decoding_capacity_cz25`** (`ListDecoding/Bounds.lean:522`, C3.5) takes arbitrary `ω : F` with no admissibility hypothesis (the file never mentions `Admissible`). Paper Def 2.15 bakes (L,s)-admissibility into FRS, and `Folded.lean:84`'s own docstring says admissibility "is left as a side condition for downstream statements about list decoding" — this downstream statement omits it. Counterexample: `ω = 1` collapses every fold, so `frsCode ≅ RS[F, L, k]` with each symbol repeated `s` times; with `F = GF(11), n = 8, k = 3, s = 4, η = 1/2` the claimed bound is `1.75` but a word agreeing with two distinct deg-≤2 polynomials on 5 points each gives `Λ ≥ 2`. Fix: add the in-tree strengthened `Admissible` hypothesis.

2. **`linear_C_le_generalized_singleton_st20`** (`ListDecoding/Bounds.lean:279`, T3.9). The exponent is real subtraction, so `C = ⊥ ⊆ GF(2)^10, ℓ = 1, δ = 0.9` satisfies all hypotheses (`Λ(⊥, 0.9) = 1 ≤ 1`) yet concludes `1 ≤ 2^(10−18) = 2^(−8)`. ST20's proof punctures `ℓ+1` distinct codewords, so it implicitly needs `|C| > ℓ`; add that hypothesis (the paper statement has the same gap — the admit shouldn't inherit it).

3. **`linear_lambda_ge_entropy_volume`** (`ListDecoding/Bounds.lean:259`, C3.8). Continuous `δ` in the entropy vs `Λ` quantizing at `⌊δn⌋`: for `C = ⊤ ⊆ GF(2)², δ = 0.49` the RHS is ≈ `2.0` but `Λ = 1` (radius `⌊0.98⌋ = 0`). Needs the integrality guard from the precise MS77 form (or entropy evaluated at `⌊δn⌋/n`).

4. **`listDecodingUpperBoundAttack` (`bits := 117`)** (`ToyProblem/Leaderboard.lean:980`; `sorry` at `:1039`). The anchor `koalaIRS` is `RS[n=4, k=2]` over `|F| ≈ 2^186`. At `n = 4` the MDS structure caps `|Λ(C^{≡2}, δ)|` at ~36 for all `δ < δ_min = 3/4`, so the attack term is ≈ `2^(−181)` and the spot-check term at `δ = 0.6` is `(0.4)^128 ≈ 2^(−169)` — the claimed `2^(−117)` floor on `bestProvableError` is unachievable; this `sorry` cannot be discharged. (The `:1036` comment's "`|Λ| → ∞` as `δ → δ_min`" is false at `n = 4`; the paper's Table 4 value `δ* = 0.468` is a paper-scale `n = 2^18..2^21` phenomenon.) `securityGap_koalaIRS_anchors = 53.01` inherits the false anchor. The FRS anchors in `Impl/FRS.lean` use genuine paper-scale parameters and are sound. Fix: paper-scale IRS anchor, or an honest (~256-bit) ceiling at this anchor.

## High — CI blockers & source tracking

5. **Generated-file reject**: the PR still touches `docs/kb/_generated/references.json`; CI's "Reject generated KB updates in PRs" step hard-fails on it (`b151ff43e` dropped the other generated files but missed this one). Revert to main.

6. **Docs-integrity failure**: `docs/wiki/proximity-prize-leaderboard.md:194` links `../../../proximity-prize` (external repo, outside the tree). De-link to plain `` `proximity-prize` `` code text, as lines 199/254/285/312 already do.

7. **Stale attribution to a since-removed theorem**: `rs_epsCA_breakdown_cs25` (`CapacityBounds.lean:411`) is tagged "ABF26-T4.17 [CS25 Cor 1]", but in the current `.tex` CS25 Cor 1 sits inside an `\ignore[old]` block (≈L2031), and the T4.17 slot is now CS25 Thm 3 (which this PR separately and correctly formalizes as `rs_epsCA_subfield_lower_cs25_thm3`). Unlike `mcaConjecture` (whose `\ignore` status *is* disclosed in the docstring and audit — nice), this removal is undisclosed. Re-tag to cite CS25 directly and note the removal in the audit row.

8. **New in the current `.tex`, uncovered**: `prop:mca-information-set-lower-bound` ("MCA error is linear": `ε_mca(C,δ) ≥ min{⌊δn⌋/|F|, 1}` for linear codes, `δ ∈ (0, δ_min)`) ships with a complete elementary proof in the source — an ideal fully-provable in-tree target, currently absent from both the Lean tree and the audit.

## Medium

9. **Admits stronger than their citations**: (a) `linear_epsMCA_1_5_johnson_gkl24` / `linear_epsCA_1_5_johnson_bgks20` (T4.11) generalize from linear codes over `F` to arbitrary F-additive alphabets; the paper licenses that only for L4.6 (footnote 6). Restrict to `LinearCode ι F` or add a justification. (b) `frs_is_subspaceDesign_gk16` (T2.18, `SubspaceDesign.lean:140`) drops the paper's `|F| > n` hypothesis and states τ as `k/(s·n)/(s−r+1)` with no `k ≤ s·n`/injectivity guard.

10. **§4 omissions not listed**: T4.8 (AHIV17) and T4.9 Item 1 (BCIKS20 Thm 1.4) have no ε-form statements, and none of the paper's `ε_mca = ε_ca` *equalities* is captured; the `CapacityBounds.lean` header lists only T4.15 as deferred. Similarly, the PDF's T3.15 (KKH26 asymptotic bound `≥ n^c` at `δ_min − O_{ρ,c}(1/log n)`) is tracked nowhere — the audit's "T3.15" row is actually CW07; Cor C.6 (its main ingredient) is proved, but the asymptotic corollary is neither stated nor listed as deferred.

11. **Citation contract**: ~32 net-new citation keys used in docstrings (ABF26, KKH26, GG25, CS25, BCHKS25, GCXK25, Jo26, …) have no entries in `blueprint/src/references.bib` (the PR's only bib change fixes the pre-existing `codingtheory` entry), contra `docs/wiki/blueprint-and-citations.md`.

12. **Same-PR wiki rule**: `docs/wiki/repo-map.md` doesn't mention the new `ProofSystem/ToyProblem/` family or the new `CodingTheory/{ListDecoding,Connections,ReedSolomon/*}` subdirectories. Also `Stir/ProximityGap.lean` references `docs/kb/ABF26_INTEGRATION_PLAN.md`, which doesn't exist in the tree.

## Low

- `Jqℓ` docstring (`JohnsonBound/Family.lean:64`) claims the printed paper has `ℓ/(ℓ−1)`; the current PDF prints `(ℓ−1)/ℓ` — the Lean formula is correct, the "correction" note is stale.
- `additive_code_supports_erasure_correction_grs12` proves neither additivity nor GRS12 complexity content (docstring is honest; name is not).
- `ProbabilityTheory.Pr_eq_tsum_indicator` (Notation.lean) duplicates `Probability.prob_tsum_form_singleton` (Instances.lean); `MvPolynomial.totalDegree_le_of_degreeOf_lt` belongs in `ToMathlib/`.

## Verified fine (majority of the PR)

Definitions layer (D2.2–D2.20, Λ, ε_pg/ε_ca/ε_mca incl. strictness and F-additive generality, line-decoding, grand-challenge boxes, toy relation/protocol/winning set, IOR appendix defs), T3.2 + C3.3 (modulo the honest Plotkin-corner sorry), L3.7, T3.12/T3.13 constants and log bases, all of Appendix C (C.1, C.3–C.7 proven; C.2's admit faithful), T4.9.2/T4.12/T4.13/T4.14/T4.16–T4.18/L4.19/T4.21/T5.1–T5.4 statement forms, the L6.6/L6.8 error terms (convex form is strictly stronger than the paper's sum — good), L6.12's corrected `N/(|F|+2N)` denominator, Claim B.1, and the FRS leaderboard anchors. Earlier-round fixes confirmed: T4.18 field quantifier, KKH26 prime/poly/smooth encoding, L6.12/L6.13 violation conjunct, encoder pinning. Deliberate deviations (strengthened `Admissible`, L2.17's `r ∈ [1,s]`, T4.18's multiplicative window, T3.14's dimension pin, R4.10's no-crossing guard, T4.7's `t > 0`) are all documented and are repairs of genuine paper edge-case bugs. Sorry census matches +37, all tagged; no `axiom`/`native_decide`; `ArkLib.lean` regenerated byte-identically; Binius/Stir/Whir touches are additive only.

One process suggestion: since the `.tex` numbering keeps drifting (T4.17→T4.19 already patched once), consider standardizing the `ABF26-*` sorry tags and audit rows on **tex labels** (`thm:ca-lower-bound`, `cor:kikh-vanilla`, …) instead of theorem numbers — half the attribution issues above are numbering drift.

Disclosure: posted by the assistant on behalf of Quang Dao. AI-authored using Claude Fable 5 (Claude Code).

dhsorens:
# Correctness review: Lean formalization vs. `papers/ef-millenium` (ABF26)

Methodology: 8 parallel paper-vs-Lean audits by area (core definitions; Johnson/MDS; KKH26 appendix; Elias/Singleton/capacity; CA/MCA + CS25/DG25; subspace-design/folded RS; toy IOP soundness chain incl. the RbrGame framework; SoundnessBounds attacks), plus kernel-level `#print axioms` checks on the headline theorems and compiled counterexample checks against this branch's build.

## 🔴 Critical: `dim_frsCode` is vacuous — its hypothesis is unsatisfiable

**`ArkLib/Data/CodingTheory/ReedSolomon/Folded.lean:125`**

```lean
lemma dim_frsCode ... (domain : ι ↪ F) (k s : ℕ) (ω : F)
    (h_encoder_inj : Function.Injective (frsEvalOnPoints domain s ω)) :
    Module.finrank F (frsCode domain k s ω) = k
```

`frsEvalOnPoints domain s ω : Polynomial F →ₗ[F] (ι → Fin s → F)` has the **whole polynomial ring** as its domain (`Folded.lean:70-73`). An injective linear map from the infinite-dimensional `Polynomial F` into the `s·|ι|`-dimensional codomain cannot exist — the nonzero polynomial `∏_{(x,j)} (X − domain x · ω^j)` is in the kernel. So `h_encoder_inj` is **false for every instantiation** and the lemma is vacuously true.

Machine-checked against this branch (both examples compile clean):

```lean
import ArkLib.Data.CodingTheory.ReedSolomon.Folded
open ReedSolomon.Folded

-- The hypothesis implies False, for any field/domain/s/ω:
example {ι : Type} [Fintype ι] {F : Type} [Field F]
    (domain : ι ↪ F) (s : ℕ) (ω : F)
    (h : Function.Injective (frsEvalOnPoints domain s ω)) : False := by
  have : Module.Finite F (Polynomial F) :=
    Module.Finite.of_injective (frsEvalOnPoints domain s ω) h
  exact Polynomial.not_finite this

-- Hence the lemma "proves" absurdities: rank 1,000,000 in a 1-dimensional ambient space
example {F : Type} [Field F] [DecidableEq F]
    (domain : Fin 1 ↪ F) (ω : F)
    (h : Function.Injective (frsEvalOnPoints domain 1 ω)) :
    Module.finrank F (frsCode domain 1000000 1 ω) = 1000000 :=
  dim_frsCode domain 1000000 1 ω h
```

Three compounding problems:

1. **The docstring advertises a discharge path that cannot exist.** `Folded.lean:122-123` suggests `h_encoder_inj := Polynomial.degreeLT_eval_inj`, and `Folded.lean:182` claims `frsEvalOnPoints_domRestrict_injective` is "the in-tree bridge that `dim_frsCode`'s `h_encoder_inj` hypothesis was waiting for." It isn't and can't be — that lemma proves injectivity of the **`domRestrict` to `degreeLT F k`**, a different proposition of a different type.
2. **The naive fix is false.** Restating with the satisfiable `domRestrict` hypothesis, the conclusion `finrank = k` fails whenever `k > s·|ι|` (the code lives in an `s·|ι|`-dimensional space). A correct restatement needs `Admissible`, `ω ≠ 0`, **and `k ≤ s·|ι|`**.
3. **This is the load-bearing dimension fact** behind the FRS rate `ρ = k/(s·n)` used by the subspace-design chain (T2.18, C2.21, prize parametrizations) and it is advertised in the file header (`Folded.lean:25`). The Koala instantiation (`Impl/FRS.lean:259`) happens to dodge it by using the `domRestrict` bridge directly for encoder injectivity — which is why nothing downstream currently fails — but the first consumer of `dim_frsCode` will find their goal unreachable (or convert it into an unguarded admit of a false statement).

**Suggested fix:**

```lean
lemma dim_frsCode ... (hadm : Admissible (Finset.univ.map domain) s ω) (hω : ω ≠ 0)
    [NeZero k] (hk : k ≤ s * Fintype.card ι) :
    Module.finrank F (frsCode domain k s ω) = k
```

deriving injectivity-on-`degreeLT` from `frsEvalOnPoints_domRestrict_injective` and computing the rank of the image of the *restricted* map (note `frsCode = (degreeLT F k).map (frsEvalOnPoints …)` equals the range of the `domRestrict`, so `Submodule.finrank_map_le`-style reasoning applies after the injectivity transfer).

## 🟡 Secondary findings

- **`Admissible` is unsatisfiable when `0 ∈ L`** (`Folded.lean:63-66`): the intra-orbit clause requires `0·ω^i ≠ 0`. `koalaFRSDomain` documents dodging this by excluding 0, but any statement hypothesizing `Admissible` on a zero-containing domain is silently vacuous. Worth a `0 ∉ L` guard or a warning in the definition docstring. (Related: ABF26 Definition 2.14's literal text also doesn't forbid intra-orbit collisions — the PR's strengthening is correct, but this corner is the residue.)
- **Knowledge-soundness content is degenerate relative to the paper** (documented in docstrings; worth restating in the PR description): `protocol62_knowledgeSound` uses the classical-choice extractor `extractZero`, so in ArkLib's cost-free model it certifies "Pr[accept ∧ no relaxed witness exists] ≤ ε" — the paper's Lemma 6.6 *efficient* extraction (O(enc + erasure-decode)) is not formalized. Similarly `protocol62_rbrKnowledgeSound` is the challenge-averaged rbr form, not Def A.5's worst-case-per-prefix form.
- **Stale comment**: `ProofSystem/ToyProblem/Spec/General.lean:378-380` still says the KS results "may be closed on their mathematical merits" — both are closed.
- **C3.8 docstring** (`Data/CodingTheory/ListDecoding/Bounds.lean:254`) calls `q^{n(ρ−1+H_q(δ))}/√(8nδ(1−δ))` the "MS77 volume estimate"; that expression is the list-size bound (the volume estimate lacks the `ρ−1`). Statement itself is correct.
- **Coverage gap**: the paper's MCA information-set lower bound (tex ~1867, `ε_mca ≥ min{⌊δn⌋/|F|, 1}`) has no in-tree counterpart. Fine if intentional; worth a row in the coverage matrix.

## ✅ Verified faithful

- **Core definitions** — `Lambda`/`closeCodewordsRel`/balls all use `≤` per Def 2.8; `minDist` excludes self-pairs; rate `k/n`; MDS `d = n−k+1`; RS `deg < k`; `restrictedRelHammingDist` = `Δ_T`; the `minDist/n ↔ δᵣ` bridge incl. empty-set corners.
- **Johnson family (T3.2/C3.3)** — `Jqℓ`/`Jcap` exact; the corrected `(ℓ-1)/ℓ` factor is mathematically forced (printed `ℓ/(ℓ-1)` would make the Johnson denominator negative); the single Plotkin-corner `sorry` is honestly disclosed and the statement survives a repetition-code sanity test.
- **KKH26 appendix** — all message-size windows, binomials (vanilla + antipodal even/odd), radius bookkeeping `δ_min − (Kd−m+1)/n = 1−K/h`, and useful-family definitions exact; the one `sorry` ([KKH26] Lemma 1) is a faithful external admit.
- **Elias/Singleton/capacity** — L3.7 proven in-tree with correct averaging; C3.8, T3.9–T3.14, the §4 capacity suite, CS25 Thm 3 (incl. the `𝔤` piecewise thresholds), DG25 2.5: all directions, exponents, floors, and constants match.
- **CA/MCA** — `epsCA`/`epsMCA`/`mcaEvent` match Defs 4.1/4.3 clause-for-clause; `ε_pg ≤ ε_ca ≤ ε_mca` proven in the right order; L4.6 at `δ < δ_min/2`; interleaving-MCA equality correctly guarded.
- **Toy IOP soundness chain** — `outputRelationFor` is a genuine AND with a *common* agreement set (interleaved metric); `accepts` checks `⟨ĝ,v⟩ = μ₁+γμ₂` and all `t` spot-checks; the convex error `(1−δ)^t + ε₀(1−(1−δ)^t)` is ≤ the paper's sum; L6.12's corrected `N·F/(F+2N)` denominator matches the tex; the RbrGame master lemmas are sound probability algebra.
- **Kernel axiom checks** — `protocol62_knowledgeSound`, both `simplified_iop_soundness_*`, `johnson_bound_lambda_le_ell`, `linear_lambda_ge_elias_volume_eli57` are all `[propext, Classical.choice, Quot.sound]`; no `sorryAx` leaks in from the IOR framework's pre-existing sorries. `mds_johnson_lambda_le` carries `sorryAx` exactly as disclosed.

The tree's documented paper repairs all check out as corrections *toward* soundness: the Johnson `(ℓ-1)/ℓ` list factor, the toy-soundness `max → sum` error term, the L6.12 winning-set denominator `F+N−1 → F+2N`, the KKH26 `w_S` sign inconsistency (tex 3429 vs 3446), and the rbr final-round state.

---

ArkLib/ProofSystem/ToyProblem/Leaderboard.lean
```diff
@@ -0,0 +1,1056 @@
...
+   on `[δ*, δ_cross ≈ 0.4695)` the Elias/[KKH26] table
+   (`tab:elias-lowerbound-thresholds`, `N ≈ 2^{186-116.49}`); on the short-length
+   band `[δ_cross, 3/4)` — where the spot-check has collapsed and the table is out
+   of range — the near-`δ_min` list-size blow-up (`|Λ| → ∞` as `δ → δ_min`,
+   cf. 2025/2197 Lemma 4.4). No proven `Lambda` lower bound exists in ArkLib
+   (`ListDecodability.lean` has only `Lambda_le_*` upper bounds), so this is
+   irreducibly external — exactly the status of the lower anchor's `ε_mca`
+   ceiling. **Axiom-clean is infeasible by design** (it is the prize's own
+   coding-theory content); the reduction is full down to these named admits. -/
+noncomputable def listDecodingUpperBoundAttack : SecurityUpperBound koalaIRS where
```
alexanderlhicks: 🔴 **BLOCKER (class-c false anchor).** `listDecodingUpperBoundAttack` with `bits := 117` is unsatisfiable at this toy point. `koalaIRS.code = RS[4,2]` is length-4 MDS, so the interleaved list size `N ≤ C(4,2) = 6` for every word; hence `bestProvableError ≤ 2^-128` and `2^-117 ≤ bestProvableError` (what `bits=117` asserts) is false. The §6.3 `N≈2^69`/116-bit figures are asymptotic in the real block length (~2^21) and don't transfer to n=4. Fix: larger-block IRS point, or lower `bits` to the true ~128, or withdraw this attack anchor. This propagates to `securityGap_koalaIRS_anchors = 53.01` (semantically void).

ArkLib/ProofSystem/ToyProblem/Leaderboard.lean
```diff
@@ -0,0 +1,1056 @@
...
+      -- The PROVEN L6.12 hook floors `winningSetSoundness` at `N/(|F|+2N)`.
+      refine le_trans ?_ (listDecoding_le_winningSetSoundness (C := koalaIRS.code) δ hδpos hδlt1
+        koalaIRS.enc koalaIRS.enc_injective rfl hF)
+      -- ★ Owed external bound (ii): the interleaved list size lower bound
+      -- `N/(|F|+2N) ≥ 2^(-117)` (`N ≳ 2^69`). On `[δ*, δ_cross)` this is the
+      -- Elias/[KKH26] table (`≈ 2^{186-116.49}`); on the short-length band
+      -- `[δ_cross, 3/4)` it is the near-`δ_min` list-size blow-up (2025/2197 L4.4).
+      -- No proven `Lambda` lower bound exists in ArkLib — irreducibly external,
+      -- exactly as the lower anchor's `ε_mca` ceiling. Phase-5/external-owed.
+      sorry
```
alexanderlhicks: Owed bound (ii) here reduces to `2^-117 ≤ N/(|F|+2N)` with `N = Lambda(interleaved RS[4,2])`, `|F|≈2^185.93`. Since `N ≤ 6` (MDS, length 4), RHS `≤ 6/2^186 ≈ 2^-183` — **false by ~66 bits**, not a hard-but-true external. This is the concrete site of the anchor defect above.

ArkLib/ProofSystem/ToyProblem/Leaderboard.lean
```diff
@@ -0,0 +1,1056 @@
...
+          _ = (1 - δ) ^ (128 : ℕ)
+                + winningSetSoundness koalaIRS.enc δ * (1 - (1 - δ) ^ (128 : ℕ)) := add_comm _ _
+      refine le_trans ?_ hconvex
+      have hδlt1 : δ < 1 := lt_trans hδ34 (by norm_num)
+      -- ★ Owed external bound (i): the interleaved list size is below the field
+      -- size `|F| = q^6 ≈ 2^186` (true in regime; no proven `Lambda` upper bound
+      -- in ArkLib bridges to the `q^6` numeric — owed external coding-theory).
+      have hF : ((Lambda (interleavedCodeSet (κ := Fin 2) koalaIRS.code) (δ : ℝ)).toNat : ℝ)
+          < Fintype.card koalaIRS.F := by
+        sorry
```
alexanderlhicks: Owed bound (i) `N < |F|` is *true but trivial* (`6 < 2^186`) and in-tree provable once the `N ≤ 6` MDS bound is stated — not "irreducibly external" as the comment says. Moot if the anchor is reworked.

ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean
```diff
@@ -0,0 +1,538 @@
...
+  sorry -- ABF26-C3.8; external admit, uses MS77 volume estimate.
+
+/-- **ABF26 Theorem 3.9 [ST20 Thm 1.2].** Generalized Singleton bound for list decoding.
+Let `F` be a finite field, `0 < ℓ < |F|`, `δ ∈ (0, 1)`, and let `C ⊆ F^n` be a linear
+error-correcting code of rate `ρ` with `|Λ(C, δ)| ≤ ℓ`. Then:
+
+  `|C| ≤ |F|^{n - ⌊(ℓ+1)/ℓ · δ · n⌋}`
+
+Equivalently, `δ ≤ ℓ/(ℓ+1) · (1-ρ)`. Admitted as an external result. -/
+theorem linear_C_le_generalized_singleton_st20
```
alexanderlhicks, line 301: 🟠 **class-c false as stated (0 consumers → faithfulness bug).** `linear_C_le_generalized_singleton_st20` is false: at `C=⊥, ℓ=1, δ=9/10, n=10` all hypotheses hold but the conclusion is `1 ≤ 2^(10-18)` (compile-confirmed `False`). The docstring itself notes the equivalent constraint `δ ≤ ℓ/(ℓ+1)·(1-ρ)`, but it is **absent from the Lean hypotheses**. Fix: add that hypothesis (or the exponent-nonneg guard) so the statement matches ST20 Thm 1.2.

ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean
```diff
@@ -0,0 +1,538 @@
...
+    _ = ((cnt f₀ : ℕ∞) : ENNReal) := by rw [ENNReal.ofReal_natCast, ENat.toENNReal_coe]
+    _ ≤ (Lambda ((C : Set (ι → F))) δ : ENNReal) := hLam
+
+/-- **ABF26 Corollary 3.8.** Volume-based lower bound on list size, using the MS77
+volume estimate `Vol_q(δ, n) ≥ q^{n·(ρ-1+H_q(δ))} / √(8·n·δ·(1-δ))`. With `ρ := k/n`:
+
+  `|Λ(C, δ)| ≥ q^{n·(ρ - 1 + H_q(δ))} / √(8·n·δ·(1-δ))`
+
+Uses `qEntropy` (ABF26 D2.2). Admitted as an external result. -/
+theorem linear_lambda_ge_entropy_volume
```
alexanderlhicks, line 266: 🟠 **class-c false as stated (0 consumers).** `linear_lambda_ge_entropy_volume` (C3.8) is false at small δ: e.g. `C=⊤, n=49, δ=0.02` gives `⌊δn⌋=0` so `Lambda=1`, while the entropy-volume RHS ≈ 625. Missing a regime hypothesis (`δ·n ≥ 1`, i.e. the radius where the volume estimate is valid). Fix: add it.

ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean
```diff
@@ -0,0 +1,538 @@
...
+When `η ≥ √(3/s)`, the bound simplifies to `|Λ(C, 1 - ρ - η)| ≤ 1/η`. Derives from
+T3.4 + T2.18 (FRS is τ-subspace-design).
+
+**Rate convention.** The FRS code `FRS[F, L, k, s, ω] ⊆ (F^s)^n` has rate
+`ρ = k / (s·n)` per ABF26 Definition 2.5 (the alphabet is `F^s`), **not** `k/n`.
+With this `ρ` both the radius and the list bound are the paper's expressions
+verbatim; e.g. the radius numerator `ρ·s = k/n`.
+
+Admitted as an external result. -/
+theorem frs_list_decoding_capacity_cz25
```
alexanderlhicks, line 862: 🟠 **class-c false as stated (0 consumers).** `frs_list_decoding_capacity_cz25` (C3.5) omits the admissibility hypothesis on `ω` (and the underlying `|F|>n`), so `ω=0` is a legal instantiation that degenerates `frsCode` and falsifies the bound. Fix: carry the `Admissible ω` (and `|F|>n`) hypotheses that CZ25 Cor 2.21 / GK16 require.

ArkLib/Data/CodingTheory/ExtensionCodes.lean
```diff
@@ -0,0 +1,320 @@
...
+
+/-- **ABF26 Lemma 2.21 [BCFW25 Lemma D.3].** List size of an extension code equals the
+list size of the corresponding interleaved base code. Let `C_B : B^k → B^n` be a
+linear code and `P` be an extension-field presentation. For every `δ ∈ (0, 1)`:
+
+  `|Λ(C_F, δ)| = |Λ(C_B^≡e, δ)|`
+
+where `C_F` is the extension code (D2.20) and `C_B^≡e` is the `e`-fold interleaved
+base code (D2.9). Admitted as an external result. -/
+theorem lambda_extensionCode_eq_lambda_interleaved
```
alexanderlhicks, line 311: 🟡 **class-b (fillable in-tree, currently admitted).** `lambda_extensionCode_eq_lambda_interleaved` (L2.21) is **true and provable in-tree**, axiom-clean, in ~70 lines (transport through the per-coordinate presentation equiv; a compiling probe was verified `[propext, Classical.choice, Quot.sound]`). Recommend proving it rather than admitting.

ArkLib/Data/CodingTheory/ExtensionCodes.lean
```diff
@@ -0,0 +1,320 @@
...
+  full `F`-`Submodule` (B-linear closure was always present; the F-scalar
+  closure is what the structural refactor delivers).
+- `lambda_extensionCode_eq_lambda_interleaved` (L2.21, [BCFW25 Lem D.3]):
+  `|Λ(C_F, δ)| = |Λ(C_B^≡e, δ)|`. Tagged sorry.
+
+## References
+
+- [ABF26] Arnon-Boneh-Fenzi. *Open Problems in List Decoding and Correlated Agreement*.
+  2026.
+- [BCFW25] Bordage-Chiesa-Fenzi-Wahby. Lemma D.3.
```
alexanderlhicks: 🟡 Citation misattribution. `[BCFW25]` bibkey is `BuenzCFW25` = **Bünz–Chiesa–Fenzi–Wang, *Linear-Time Accumulation Schemes*** — not "Bordage-Chiesa-Fenzi-Wahby". (Confirmed against tex L1320 and the reference PDF.) Please correct the author expansion.

ArkLib/Data/CodingTheory/ExtensionCodes.lean
```diff
@@ -0,0 +1,320 @@
...
+
+(The distance equality `δ_min(C_F) = δ_min(C_B)`, referenced in the L2.21 paragraph
+context, is from Dao-Petrov 2025 (Theorem 3.2 in their paper). The knowledge-base
```
alexanderlhicks: 🟡 Citation misattribution. The distance-equality result cited here as "Dao-Petrov 2025" is `DiamondP23` = **Diamond–Posen, *Succinct Arguments over Towers of Binary Fields*** (Thm 3.2). Please correct.

ArkLib/Data/CodingTheory/SubspaceDesign.lean
```diff
@@ -0,0 +1,153 @@
...
+
+Note: `[s]` in the paper denotes `{1, …, s}` (one-based), which we encode in Lean as
+`Finset.Icc 1 s`. With this convention `τ(1) = s·ρ/s = ρ` and `τ(s) = s·ρ`, matching
+the paper's boundary values.
+
+The FRS case requires `(L, s)`-admissibility of `ω`; the multiplicity case requires
+`|F| > n` and `char(F) > ρ·s·n > s`. We state only the FRS half here; the multiplicity
+half is gated on `D2.19 / DA.7` (univariate-multiplicity definition), which is tracked
+separately. Admitted as an external result. -/
+theorem frs_is_subspaceDesign_gk16
```
alexanderlhicks, line 166: 🟠 **Missing hypothesis (medium).** `frs_is_subspaceDesign_gk16` omits the `|F| > n` hypothesis that the pinned tex's `thm:folded-rs-are-subspace-design` now requires (introduced in the 32ffa05→120358f delta); the docstring also still carries the pre-delta multiplicity condition. Please add `|F|>n` and refresh the docstring.

docs/kb/audits/open-problems-list-decoding-and-correlated-agreement.md
```diff
@@ -1,279 +1,334 @@
...
+| `T3.13` | RS large list over prime fields | stated (external admit) | [ListDecoding/Bounds.lean](../../../ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean) | `CodingTheory.rs_lambda_large_prime_ghsz02` | GHSZ02 Cor 20; tagged sorry. |
+| `T3.14` | Large-rate RS lower bound | stated (external admit) | [ListDecoding/Bounds.lean](../../../ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean) | `CodingTheory.rs_lambda_high_rate_jh01` | JH01 Thm 2; tagged sorry. Statement corrected 2026-06-10: dropped the unsatisfiable `\|C\| = j+1` conjunct (misread of the paper's block length `\|L\| = j+1`, already encoded) and pinned the dimension `k := j−1` (rate `(j−1)/(j+1)`). **Re-review fix (2026-06-10b):** dimension pin corrected `k = j-1 → j` (deg-`<k` convention; `j-1` forced min-dist 3 vs 1-error radius — unsatisfiable). |
+| `T3.15` | CW07 hardness barrier | out of scope | none | `CodingTheory.rs_dlog_barrier` (external; not stated) | Algorithmic hardness (discrete-log reduction); we formalise combinatorial statements only. |
```
alexanderlhicks, line 171: 🟠 **Coverage gap (medium).** Paper §3 `thm:kikh-asymptotic-list-lower-bound` (RS list size ≥ n^c near δ_min — the *asymptotic* form, and ironically the true version of what the IRS attack anchor needs) has no ArkLib decl and no coverage-matrix row (the §3 table jumps T3.14→T3.15). Add a documented-deferred row.

docs/kb/audits/open-problems-list-decoding-and-correlated-agreement.md
```diff
@@ -1,279 +1,334 @@
...
+| `T4.18` | CA jump at Johnson bound | stated (external admit) | [CapacityBounds.lean](../../../ArkLib/Data/CodingTheory/ProximityGap/CapacityBounds.lean) | `CodingTheory.rs_epsCA_johnson_jump_bchks25` | BCHKS25 Cor 1.7; tagged sorry. Johnson radius `J(δ) := 1 - √(1-δ)` inlined. (Sits at position T4.19 after the 2026-06 `.tex` renumbering.) Statement corrected 2026-06-10: additive window `\|F\|^{(1+ε)/2} ± 1` was unsatisfiable together with `16 ∣ n` for almost all char-2 fields; replaced by the multiplicative window `x/2 ≤ n ≤ 2x` plus guards `ε < 1`, `\|F\| ≥ 1024`. |
+| `L4.19` | CA bounded below by sampling probability | stated (external admit) | [CapacityBounds.lean](../../../ArkLib/Data/CodingTheory/ProximityGap/CapacityBounds.lean), related DG25 work in [DG25/MainResults.lean](../../../ArkLib/Data/CodingTheory/ProximityGap/DG25/MainResults.lean) (contains 2 sorries) | `CodingTheory.linear_epsCA_ge_sampling_dg25` | DG25 Thm 2.5; tagged sorry. |
```
alexanderlhicks: 🟠 **Coverage gap (medium).** Paper §4 `prop:mca-information-set-lower-bound` (`ε_mca(C,δ) ≥ min{⌊δn⌋/|F|,1}` for δ<δ_min) is silently missing — no decl, no matrix row. Add a documented-deferred row.

docs/kb/audits/open-problems-list-decoding-and-correlated-agreement.md
```diff
@@ -1,279 +1,334 @@
...
+| `L6.5` | Every additive code supports erasure correction | present | `CodingTheory.additive_code_supports_erasure_correction_grs25`, `CodingTheory.eq_of_consistent_with_erased` in [Erasure.lean](../../../ArkLib/Data/CodingTheory/Erasure.lean) | same | **PROVEN sorry-free + axiom-clean 2026-06-10** (moved next to D6.4 — zero toy-problem content). The corrector is built classically: below `minDist C` erasures the consistent codeword is unique (`eq_of_consistent_with_erased`, Hamming-distance pigeonhole), so `Classical` choice yields the decoder; the failure clause is by construction. The paper's `O((s·n)³)` field-operation bound is out of ArkLib's cost-free model (extractors are unclocked library-wide); only existence is formalized. |
```
alexanderlhicks: 🟡 Matrix drift. The L6.5 row names `additive_code_supports_erasure_correction_grs25`, but the in-tree decl is `..._grs12`. Align the row to the actual symbol.

docs/kb/audits/open-problems-list-decoding-and-correlated-agreement.md
```diff
@@ -1,279 +1,334 @@
...
...apse from `koalaFRS`'s `98.91`. **Mechanism (not a sharper citation):** the `ε_mca` admit is the *same* τ-subspace-design family; folding closes the gap because larger `s` lets the operating point climb to high δ (`r` up to `108` keeps `τ(r+1) = s·ρ/(s−r)` near `ρ`), which small `s` cannot (`s ≤ 2^4` proves no soundness). The GG25 capacity corollary `frs_epsMCA_capacity_gg25` (`s > 16/η²`) does *not* apply at the `r = 108` point (`η = 1969/71784 ≈ 0.0274` needs `s > 16/η² ≈ 21266 > 4096`). Construction scaling validated against the paper's argument-size column (`s·|L| = 2^21` fixed ⇒ `3.91 MiB`). Owed externals mirror `koalaFRS` post-Track-B: `koalaFRS12Enc_injective` and `koalaFRS12_minRelDist` (`= 257/512`) are full `sorry`-free derivations through the same two `Folded.lean` bridges on the coset domain `koalaFRS12Domain j = γ^(2^12·j)` with the **same** `γ` (each row needs only `orderOf γ ≥ 2^21`), so both rows share the now-**proven** `koalaFRSγ_exists`; `koalaFRS12Enc_injective`, `koalaFRS12_minRelDist`, and `frsUpperBound_attack12` are fully axiom-clean. The only by-design admit left is the lower-bound τ-subspace-design `ε_mca` (≈2^(−142.7) actual, capped ≤ 2^(−118)·3/1000). |
```
alexanderlhicks: 🟡 Matrix drift. The D6.11 row misattributes external proximity-prize-repo entries as in-tree ArkLib theorems with renamed-away symbol names, contradicting the code and the sibling wiki leaderboard. Please reconcile.

alexanderlhicks (commented):
## Extreme adversarial review — 2026-07-02/03 (fresh, whole-PR)

Full re-review of this PR against ground truth (`ef-millenium.tex` @ `120358f`, reference PDFs, and first-hand `#print axioms`), independent of the 2026-06-29 review. Method: deterministic Phase-0 ledgers over all 413 PR declarations + 13 adversarial review surfaces + 37 per-`sorry` adjudicators + skeptic panels (130+ agents). Full artifacts + `VERDICT.md`: `docs/kb/queries/abf26-extreme-review-2026-07-02/`.

**This overturns the prior "GO-WITH-FIXES, 0 soundness blockers" verdict.**

### Verdict
- **Goal (i) faithful formalization — CHANGES REQUIRED.** Crown jewels + bulk of the library are sound and axiom-clean, but the PR admits **4 false statements** via `sorry` and needlessly admits **1 true, in-tree-provable** result.
- **Goal (ii) prize uncheatable — BLOCKED.** The IRS attack anchor is backed by a false lemma (below), and the prize repo's Verified-tier gate is bypassable (a compiled exploit lands a false `bits=256` entry in the winning tier — see prize-repo note).

### 🔴 Headline (cross-validated by hand + two independent agents)
**`listDecodingUpperBoundAttack` (`bits := 117`) is false and unsatisfiable.** `koalaIRS.code` is `RS[4,2]` — a **length-4 MDS** code, so its interleaved list size is `≤ C(4,2)=6` for every word. The owed bound `2^-117 ≤ N/(|F|+2N)` (`N = Lambda(interleaved RS[4,2])`, `|F|≈2^185.93`) is therefore `≤ 6/2^186 ≈ 2^-183` — **false by ~66 bits** — and the full anchor is unsatisfiable (`bestProvableError ≤ 2^-128`). Root cause: the paper's §6.3 attack numbers (`N≈2^69`, ~116-bit ceiling) are **asymptotic in the real block length (~2^21)** and cannot transfer to `n=4`. Consequently `securityGap_koalaIRS_anchors = 53.01` is arithmetically correct but semantically void. The IRS *lower* anchor (63.99) and all FRS anchors are **sound and not infected** (FRS survives because δ_min sits just above ½). Fix options: instantiate a larger-block IRS point, lower `bits` to the true ~128, or withdraw the IRS attack anchor.

### Sorry / axiom ledger (first-hand, 37 PR-owned)
`32×(a)` legitimate cited externals · `1×(b)` true-but-fillable · `4×(c)` false-as-stated · `0` `axiom` declarations · all 4 crown jewels + T3.2 + L6.12/13 axiom-clean.

- `(c)` **idx36** `listDecodingUpperBoundAttack` — false, prize-reachable (above).
- `(c)` **idx8** `linear_C_le_generalized_singleton_st20` — **compile-confirmed false** at `C=⊥, ℓ=1, δ=9/10` (`1 ≤ 2^(10-18)`); missing hyp `δ ≤ ℓ/(ℓ+1)(1-ρ)`. Zero consumers.
- `(c)` **idx7** `linear_lambda_ge_entropy_volume` — false at small δ (floor collapses the list); missing `δn ≥ 1`. Zero consumers.
- `(c)` **idx15** `frs_list_decoding_capacity_cz25` — false at `ω=0`; missing admissibility + `|F|>n`. Zero consumers.
- `(b)` **idx4** `lambda_extensionCode_eq_lambda_interleaved` — **compile-confirmed fillable** in-tree, axiom-clean (~70-line transport). Should be proven.

The three zero-consumer `(c)` sites are faithfulness bugs (add the missing regime hypotheses); only idx36 is prize-reachable.

### Prize repo (`~/proximity-prize`, not in this diff — for bookkeeping)
Compiled + `#print axioms`-verified exploit: a false `bits=256` entry reaches the **Verified** (winning) tier. `scripts/Leaderboard.lean:167` gates Verified by `!ax.contains sorryAx` (accepts any custom `axiom`/`native_decide`), and `scripts/check-hardening.sh:15` is a line-anchored grep evaded by `@[simp] axiom …` / `open X in axiom …`. Fix: intersect against the `{propext, Classical.choice, Quot.sound}` whitelist + replace the grep with a `collectAxioms`/AST gate + check `mkEntry`'s head constructor. Harder attacks (contract shadowing, bits-spoofing) verified **safe**.

### Sound (verified, not assumed)
Crown jewels (≥ paper strength, convex error ≤ paper's SUM), spec definitions, ConstrainedCode, the VCV-io staging mirror, `#597` namespace move, JohnsonBound de-Field-ing (conservative generalization), mergeability (clean vs `main`), and **all three PAPER_REVS.md paper-error claims independently confirmed**.

### Other (medium, non-blocking) — see inline comments
Two paper results silently missing from code + coverage matrix (`thm:kikh-asymptotic-list-lower-bound` §3, `prop:mca-information-set-lower-bound` §4); coverage-matrix `_grs25`/D6.11 drift; two citation misattributions; `SubspaceDesign` missing `|F|>n`; stale tex line refs; `validate.sh --docs` exits 1 on two broken links in the 2026-06-24 audit doc.

Prioritized fix list in `VERDICT.md §7`. Inline comments below mark the actionable in-tree sites.

alexanderlhicks:
starting to feel more confident about merging this :+1:

alexanderlhicks:
Tweaks to GrandMCAResolution in https://github.com/Verified-zkEVM/ArkLib/pull/505/commits/cc19bc17eadfd9953202b127bee2f56284b844e5 following the valid point surfaced by https://github.com/Verified-zkEVM/ArkLib/pull/617/

ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean
```diff
@@ -0,0 +1,878 @@
...
+
+Equivalently, `δ ≤ ℓ/(ℓ+1) · (1-ρ)`. Admitted as an external result.
+
+The hypothesis `_hC_card` (`|C| > ℓ`) is implicit in ST20's conventions (an `[n, k]`
+linear code with `k ≥ 1` and `|F| > ℓ` has `|C| = |F|^k > ℓ`): their pigeonhole proof
+produces `ℓ + 1` distinct codewords agreeing on a coordinate prefix, which requires
+`|C| ≥ ℓ + 1` to exist. Without it the statement is false at degenerate subspaces
+(e.g. `C = ⊥`, `ℓ = 1`, `δ` close to `1`: the floor exceeds `n` and the RHS drops
+below `1 = |C|`). -/
+theorem linear_C_le_generalized_singleton_st20
```
lalalune, line 301: **Blocking: this theorem is false as stated.** Take the ternary length-3 repetition code `C = {(0,0,0),(1,1,1),(2,2,2)}`, with `ℓ = 1` and `δ = 1/2`. Distinct codewords have Hamming distance 3, while `⌊δ·n⌋ = 1`; therefore radius-1 balls are disjoint and `Lambda C δ ≤ 1`. The new guard also holds because `|C| = 3 > 1 = ℓ`. But the claimed conclusion specializes to `3 ≤ 3^(3 - ⌊2·(1/2)·3⌋) = 3^0 = 1`. The ABF/ST20 argument requires an integer-lattice condition on `δ·n` (or an equivalent integral radius formulation) and a meaningful-radius inequality ensuring the subtraction exponent is justified. Please strengthen the declaration with those hypotheses and replace the `sorry` with a proof of the corrected statement.

ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean
```diff
@@ -0,0 +1,878 @@
...
+  `|Λ(C, δ)| ≥ q^{n·(ρ - 1 + H_q(δ))} / √(8·n·δ·(1-δ))`.
+
+Uses `qEntropy` (ABF26 D2.2). Admitted as an external result.
+
+The hypothesis `_hδn_int` (the radius `δ·n` is an integer) is the regime in which the
+MS77 single-term Stirling estimate is stated; the paper's corollary inherits it
+implicitly. Without it the bound is false at small `δ`: for `0 < δ·n < 1` the relative
+ball collapses to Hamming radius `0`, so the list is `{f} ∩ C` while the entropy-volume
+RHS can exceed `1`. -/
+theorem linear_lambda_ge_entropy_volume
```
lalalune, line 266: **This admitted declaration should be proved in this PR.** The existential hypothesis already provides the integer-lattice witness needed by the McEliece-Sarwate reduction. A direct proof should: unpack `hInt` to identify `δ * n` with an integer radius; rewrite the real-valued `Lambda` assumption at that radius; apply the lattice-point MS77 bound for a linear code; then discharge the cardinality/rank conversion using `linearCode_card`. Please keep the integer guard in the public statement and remove this new `sorry`.

ArkLib/Data/CodingTheory/SubspaceDesign.lean
```diff
@@ -0,0 +1,181 @@
...
+`∑ᵢ dim(A ⊓ ker projᵢ) = n = 7 > 6 = finrank A · τ(2) · n`. The load-bearing source fact
+is [GK16 Lemma 12]'s folded-Wronskian criterion, stated for **`γ` a generator of `F×`**
+(`W_γ(1, X^d) = X^d(γ^d − 1)` vanishes when `γ^d = 1`, `d < k`). The pinned tex
+(`thm:folded-rs-are-subspace-design`, L1263–1277) omits any order condition, and GG25's
+own restatement (Def 2.18 / Thm 2.19, `q > sn` only) is falsified by the same
+counterexample — recorded upstream as `PAPER_REVS.md` finding #13. We carry GK16's own
+generator hypothesis `_hω_gen` (not a weaker `ord(ω) ≥ k` guard, which blocks the known
+counterexample but is not licensed by the cited source — cf. the PAPER_REVS #12 lesson
+on unlicensed hybrid strengthenings). -/
+theorem frs_is_subspaceDesign_gk16
```
lalalune, line 166: **Please narrow or prove this admitted theorem before merge.** The required proof obligations are not just the displayed order guard: the folded-Wronskian argument needs an explicit message-degree budget (equivalently a condition such as `k ≤ s * n`, depending on this file’s indexing), the evaluation points must remain distinct through every fold, and the final subspace-design dimension bound must be connected to this declaration’s exact `m * (s - 1)` profile. If those hypotheses are already derivable from `h_rate` and the surrounding definitions, expose the lemmas and close the proof; otherwise add them to the theorem instead of leaving a stronger admitted interface.

lalalune (changes requested):
## Review of current head `213000c9f`

The July remediation fixes several earlier defects: the interleaved Reed-Solomon anchor now uses the intended `RS[2^21,2^20]` parameters; the extension-code dimension equality is proved; the entropy and folded-RS statements now include necessary regime/order guards; and the citation and theorem-matrix drift is substantially improved.

Two items still block merge.

### `linear_C_le_generalized_singleton_st20` is false

In `ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean:293`, take the ternary length-3 repetition code `C = {(0,0,0),(1,1,1),(2,2,2)}`, with `ℓ = 1` and `δ = 1/2`.

Every two distinct codewords have Hamming distance 3. Since `⌊δ n⌋ = 1`, radius-1 balls around distinct codewords are disjoint, hence `Lambda C δ ≤ 1`. The added cardinality guard holds: `|C| = 3 > 1 = ℓ`.

The conclusion nevertheless becomes `3 ≤ 3^(3 - ⌊2(1/2)3⌋) = 3^0 = 1`, a contradiction. The theorem needs an integer-lattice hypothesis on `δ n` (or an integral-radius formulation) together with the meaningful-radius inequality needed by the ST20 exponent argument.

### The branch conflicts with current `main`

A fresh merge simulation reports content conflicts in `ArkLib.lean`, `ArkLib/ToVCVio/OracleComp/Coercions/SubSpec.lean`, `ArkLib/ToVCVio/OracleComp/SimSemantics/SimulateQ.lean`, and the ABF26 audit ledger.

It also reports modify/delete conflicts for `ArkLib/ProofSystem/Whir/BlockRelDistance.lean`, `Folding.lean`, `MutualCorrAgreement.lean`, and `RBRSoundness.lean`. Those modules were removed from current `main`. Please rebase and deliberately migrate or drop the affected changes; restoring the deleted files would reintroduce retired modules.

### Admitted declarations to tighten

- `linear_C_le_johnson_mceliece_sarwate` at `Bounds.lean:266`: the statement now has the integer witness needed by the source theorem, but the proof remains admitted. Unpack the witness, reduce to the integral-radius MS77 bound, and close the linear-code cardinality/rank conversion.
- `foldedRS_is_subspaceDesign_gk16` at `SubspaceDesign.lean:166`: verify that the exact statement derives or exposes the message-degree budget, distinct folded evaluation points, and the claimed `m * (s - 1)` dimension profile. If these do not follow from `h_rate`, add the missing hypotheses rather than preserving a stronger admitted interface.

I am running repository validation at this exact PR head and will post the command result separately. The changed coding-theory modules have compiled successfully so far.

ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean
alexanderlhicks: Confirmed and fixed in `ffaa62ab`. The counterexample is exactly right — the `ℓ < |C|` guard I added on 2026-07-03 does not exclude it (1 < 3 holds), and the cardinality bound collapses to `3 ≤ 3^0 = 1`. Root cause is the floor/integer-radius quantisation: `_hΛ` alone doesn't pin `δ·n` to the lattice.

Rather than carry an integer-radius guard, I replaced `_hC_card` with the equivalent rate–radius hypothesis `δ ≤ ℓ/(ℓ+1)·(1 − dim C/n)` (the docstring's own stated equivalent). It rules out your counterexample (`1/2 ≤ 1/3` is false) and makes the cardinality bound a direct consequence of `|C| = |F|^{dim C}` together with `⌊(ℓ+1)/ℓ·δ·n⌋ ≤ n − dim C`. **The `sorry` is now removed** — the leaf is proven in-tree and axiom-clean `[propext, Classical.choice, Quot.sound]`. `_hΛ` is retained to document the ST20 list-decoding context it arises from.

ArkLib/Data/CodingTheory/ListDecoding/Bounds.lean
alexanderlhicks: Agreed the statement is sound with the `∃ d, d = δ·n` guard, but I'm keeping this as a tagged external admit for now. The reduction you sketch bottoms out on the MS77 entropy-volume estimate `Vol_q(δ,n) ≥ q^{n·H_q(δ)}/√(8·n·δ·(1−δ))`, which is **not** formalized in ArkLib — it only exists in the docstrings here. The neighbouring `linear_lambda_ge_elias_volume_eli57` gives the clean counting half (`Vol/q^{n−k} ≤ Λ`, proven), so C3.8 is exactly that plus the MS77 lower bound on `Vol` — but supplying MS77 is a substantial standalone analytic lemma (Stirling bounds on the Hamming-ball volume) that warrants its own effort rather than an inline fill. It stays consistent with the other six §3 external admits in this file. Tracked for a follow-up.

ArkLib/Data/CodingTheory/SubspaceDesign.lean
alexanderlhicks: Keeping this as a tagged external admit for [GK16], and I don't think the suggested extra hypotheses are the right move here. Two points:

1. **`IsSubspaceDesign` is an upper bound** on `(∑ᵢ dim(A ⊓ ker projᵢ))/n ≤ dim A · τ(r)`, so a larger message-degree budget (larger `k`, hence larger `τ`) makes the constraint *weaker*, not false — it can't be the source of an unsound instantiation. The evaluation points are also already forced distinct by `domain : ι ↪ F` plus the `Admissible` cross-fold condition.

2. The genuinely load-bearing source fact is [GK16 Lemma 12]'s folded-Wronskian criterion, which requires **`ω` a generator of `F×`** — that hypothesis (`_hω_gen : orderOf ω = |F| − 1`), together with `|F| > n` and `Admissible`, is already carried (restored in `25045d51`). Adding a `k ≤ s·n` guard that isn't needed for truth and isn't licensed by the cited source would be an unlicensed strengthening of the interface (the pattern PAPER_REVS #12 warns against).

So the statement is defensible as it stands; a real proof needs the folded-Wronskian machinery of GK16 Lemma 12, which is out of scope for this PR. Left as a faithful admit.

PR #511, opened 2026-05-19
Multilinear definition of RS-codes

ElijahVlasov:
The WHIR paper introduces an equivalent characterisation of the RS-codes in terms of multilinear polynomials. This PR introduces that reformulation and proves it.


alexanderlhicks:
Thanks!

alexanderlhicks (approved):

PR #512, opened 2026-05-20
FftDomain improvements

ElijahVlasov:
1. Make simp lemmas related to domain membership better. Different membership statements (`x \in domain.toFinset`, `x \in domain.toSubgroup`) simp down to ones of the form `x \in domain`.
2. Added `neg_mem_domain_iff_mem`, `neg_mem_domain_of_mem` and related.
3. `CosetFftDomain.injOn` is a simp lemma now.
4. `CosetFftDomain.ext` is now an ext-theorem (it wasn't before because of a typo).

### DIFF
--- a/ArkLib/Data/CodingTheory/ReedSolomon/FftDomain.lean
+++ b/ArkLib/Data/CodingTheory/ReedSolomon/FftDomain.lean
@@ -407,8 +407,9 @@ abbrev SmoothFftDomain (n : ℕ) (F : Type) [Field F] : Type := FftDomain (Fin (
 namespace FftDomain
 
 omit [DecidableEq F] in
-private lemma neg_one_in_domain' {n} [nz : NeZero n] {ω : SmoothFftDomain n F} :
-  ∃ k : Fin (2 ^ n), (ω k : F) = -1 := by
+@[simp]
+lemma neg_one_mem_domain {n} [nz : NeZero n] {ω : SmoothFftDomain n F} :
+  -1 ∈ ω := by
   have hn : n ≠ 0 := NeZero.ne _
   -- Let's denote this element as `k = 2^(i-1) : Fin (2^i)`.
   set k : Fin (2 ^ n) := ⟨2 ^ (n - 1), by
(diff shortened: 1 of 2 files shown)


quangvdao (commented):
Non-blocking informational notes from review. I did not find a correctness blocker here; these are API/docs polish suggestions only.

1. The module doc still lists `CosetFftDomain.neg_mem_dom_of_mem_dom`, but the PR renames the theorem to `CosetFftDomain.neg_mem_subdomain_of_mem_subdomain`. It would be good to update that doc entry so the public API list stays discoverable.

2. `neg_one_in_domain'` proves a useful witness form, roughly `∃ k, ω k = -1`, but it is private and primed while only the membership corollary is public. Future proofs may want the index witness directly, so consider exposing it under a descriptive public name, then deriving `neg_one_mem_domain` from it.

4. `CosetFftDomain.neg_mem_subdomain_of_mem_subdomain` now seems like it can be a small wrapper around the new general `CosetFftDomain.neg_mem_domain_of_mem` applied to `ω.subdomain i`, with `0 < i` supplying the `NeZero i.val` instance. That would avoid duplicating the witness proof and make the relationship between the APIs clearer.

Again, all of this is informational and non-blocking.

Posted by Cursor assistant (model: GPT-5) on behalf of the user (Quang Dao) with approval.

ArkLib/Data/CodingTheory/ReedSolomon/FftDomain.lean
```diff
@@ -396,6 +406,63 @@ abbrev SmoothFftDomain (n : ℕ) (F : Type) [Field F] : Type := FftDomain (Fin (
 
 namespace FftDomain
 
+omit [DecidableEq F] in
+private lemma neg_one_in_domain' {n} [nz : NeZero n] {ω : SmoothFftDomain n F} :
```
alexanderlhicks: I agree with @quangvdao's comment on this

> neg_one_in_domain' proves a useful witness form, roughly ∃ k, ω k = -1, but it is private and primed while only the membership corollary is public. Future proofs may want the index witness directly, so consider exposing it under a descriptive public name, then deriving neg_one_mem_domain from it.

ElijahVlasov: There's no need to make it public as the definition of domain membership is exactly that. So one can do 
```lean
example {n} [NeZero n] (ω : SmoothFftDomain n F) : True := by
  obtain ⟨i, hminusone⟩ : ∃ i, ω i = -1 := neg_one_mem_domain (ω := ω)
  tauto
```
image

ElijahVlasov: Anyway removed `neg_one_in_domain'`

ElijahVlasov:
> CosetFftDomain.neg_mem_subdomain_of_mem_subdomain now seems like it can be a small wrapper around the new general CosetFftDomain.neg_mem_domain_of_mem applied to ω.subdomain i, with 0 < i supplying the NeZero i.val instance. That would avoid duplicating the witness proof and make the relationship between the APIs clearer.

Removed `CosetFftDomain.neg_mem_subdomain_of_mem_subdomain` completely as it's not needed/used.

ElijahVlasov:
> The module doc still lists CosetFftDomain.neg_mem_dom_of_mem_dom, but the PR renames the theorem to CosetFftDomain.neg_mem_subdomain_of_mem_subdomain. It would be good to update that doc entry so the public API list stays discoverable.

Fixed.

alexanderlhicks (approved):

PR #516, opened 2026-05-26
feat/chore: generic ring switch

alexanderlhicks:
Pulling out the ring switching stuff from binius to make it more generic and reusable for other efforts (e.g. hachi).
WIP, will complete this before going back to #505.


alexanderlhicks:
I think there's enough here to make this a useful enough PR to merge and build upon, still needs a bit of clean up.

PR #517, opened 2026-05-27
Domains refactored one more time

ElijahVlasov:
Large quality-of-life refactor of the FFT domain library. Main changes:
1. Split `FftDomain.lean` into several files so no issues with the length-of-file lint anymore :rofl: . 
2. Now the `CosetFftDomain` is the ground layer definition for domains not `FftDomain`. `FftDomain` is defined as a `CosetFftDomain` whose coset shift is `1`.
3. Both `FftDomain` and `CosetFftDomain` implement `CosetFftDomainClass` that allows to reduce code duplication between the two :man_technologist: . `FftDomainClass` is for facts specific to non-coset domains.
4. Now there is only one version of `subdomain` — parameterised by natural numbers such that `subdomain \omega 0 = \omega`. This was called `subdomainNatReversed` previously and since this is the only flavour ever used in practice it has been made the definition.


ArkLib/Data/Domain/CosetFftDomain/ToFftDomain.lean
```diff
@@ -0,0 +1,137 @@
...
+lemma mul_mem_of_mem_toFftDomain_of_mem {y : F}
+  (hx : x ∈ ω.toFftDomain)
+  (hy : y ∈ ω) :
+  x * y ∈ ω := CosetFftDomainClass.mul_mem_of_mem_toFftDomain_of_mem hx hy
+
+end CosetFftDomain
+
+namespace FftDomain
+
+lemma toFffDomain_eq_self {ω : FftDomain ι F} :
```
alexanderlhicks: maybe toFftDomain_eq_self

alexanderlhicks (approved):

PR #433, opened 2026-03-28
WIP: rebuild Interaction and oracle protocol layers

quangvdao:
## Summary

This draft PR is the current working branch for the Interaction rebuild. The latest pass pins ArkLib to the slimmed VCVio strategy hierarchy and migrates the Interaction/oracle call sites onto the new `Spec.TwoParty` API.

## Latest Strategy Slimming Pass

- Pins VCVio to `06a5ab1956e82d433aceeb428b13615e43669edc`.
- Migrates ArkLib away from the older `Strategy.runWithRoles`, role-decoration, counterpart, public-coin counterpart, and monadic strategy helper names.
- Uses the new VCVio layout around `Spec.TwoParty.run`, `Spec.TwoParty.Focal`, `Spec.TwoParty.Counterpart`, `RoleDecoration`, `pairedSyntax`, and `LawfulCommMonad`.
- Adds VCVio `StrategyOver.Hom` / `StrategyOver.map` at the generic `PFunctor.FreeM` and `Spec` layers, plus `PublicCoinCounterpart.toCounterpartHom` for the public-coin-to-counterpart forgetting map.
- Updates oracle protocol/spec, Fiat-Shamir, BCS, Sumcheck, Spartan, FRI, and boundary/reification call sites to compile against the slim API.

## Broader Branch Shape

- Adds the `ArkLib/Interaction/` stack for interactive protocol semantics, including reduction/security surfaces, oracle execution/composition, boundary transport, reification, Fiat-Shamir, and BCS-related scaffolding.
- Adds interaction-native protocol frontends under `ArkLib/ProofSystem/Sumcheck/Interaction/` and `ArkLib/ProofSystem/Fri/Interaction/`.
- Adds computable polynomial support under `ArkLib/Data/CompPoly/` needed by the Sumcheck/FRI work.
- Starts blueprint coverage for the Interaction framework, including foundations, composition, security, oracle protocols, Fiat-Shamir, BCS, and boundary material.

## Validation

- Local: `./scripts/validate.sh` passes after the VCVio strategy API migration and strategy-hom follow-up.

## Notes For Reviewers

Please treat this as an evolving architecture branch rather than a final review request. High-level feedback on API shape, module ownership, naming, and whether the Interaction/oracle split is heading in the right direction is more useful at this stage than line-level polish.

---

Posted by Cursor assistant (model: GPT-5) on behalf of the user (Quang Dao) with approval.

PR #530, opened 2026-06-03
feat: coordinate-wise special soundness (CWSS)

tobias-rothmann:
## Coordinate-wise Special Soundness (CWSS)

The PR defines transcript trees over interactive protocols, abstracts their allowed branching patterns via `ChallengeTreeShape`, and uses that abstraction to state CWSS as extraction from accepting trees with coordinate-wise challenge branching. The same tree infrastructure also supports ordinary special soundness, which is technically a specialization of CWSS.

## Design

CWSS is layered on top of generic transcript-tree structure:

- `ChallengeTree` represents families of transcripts sharing prefixes and branching at verifier challenges.
- `ChallengeTreeShape` describes valid branching independently of any particular soundness notion.
- `CWSSStructure` supplies coordinate decompositions, alphabets, and soundness parameters, then induces a `ChallengeTreeShape`.
- CWSS soundness is stated for verifiers and oracle verifiers as extraction from accepting trees of that induced shape.

Sequential composition is handled generically at the transcript-tree level: a tree over `pSpec₁ ++ₚ pSpec₂` is split into a first-stage tree over `pSpec₁` and second-stage trees over `pSpec₂` below each first-stage leaf. CWSS composition then reuses this split.

## Tradeoffs

The transcript-tree composition code contains some dependent-index bookkeeping because appended protocol rounds do not line up definitionally with their left/right components in Lean. The upside is a cleaner public API: CWSS and special soundness can share reusable tree-shape composition rather than duplicating tree arguments.

- Seems like the `TranscriptTree` is also implemented in @quangvdao 's refactor #433 as `ClaimTree`.


tobias-rothmann:
Seems like the `TranscriptTree` is already implemented in @quangvdao 's refactor #433 as `ClaimTree`.

quangvdao:
I don't mind some duplication, can always be cleaned up

tobias-rothmann:
> I don't mind some duplication, can always be cleaned up

Do you plan to do major changes to the `ClaimTree` definition @quangvdao? If not, I would refactor `TranscriptTree ` to more closely mirror your definition so the transfer gets easier.

quangvdao:
Perhaps more changes will be needed to `ClaimTree` in my PR branch. I will take a look later

alexanderlhicks:
@quangvdao the current transcript tree here seems good, at least as far as hachi is concerned. I'll merge it now, and we can see about reconciling with your ClaimTree after.

ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/Composition.lean
```diff
@@ -0,0 +1,469 @@
...
+    exact hout
+
+omit [∀ i, SampleableType (pSpec₂.Challenge i)] in
+/-- Generic preservation of tree-special soundness under binary verifier append. -/
+theorem append_treeSpecialSound
+    (V₁ : Verifier oSpec Stmt₁ Stmt₂ pSpec₁)
+    (V₂ : Verifier oSpec Stmt₂ Stmt₃ pSpec₂)
+    (S₁ : ChallengeTreeShape pSpec₁) (S₂ : ChallengeTreeShape pSpec₂)
+    (verify₁ : Stmt₁ → pSpec₁.FullTranscript → Stmt₂)
+    (hV₁ : ∀ stmt tr, V₁.verify stmt tr = pure (verify₁ stmt tr))
```
alexanderlhicks, line 371: you may be able to reuse Verifier.IsPure from https://github.com/Verified-zkEVM/ArkLib/blob/main/ArkLib/OracleReduction/Basic.lean here

ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/Composition.lean
```diff
@@ -0,0 +1,469 @@
...
+/-- The appended `decompose` at a right index is the right component's `decompose`, up to cast. -/
+theorem append_decompose_inr (i₂ : pSpec₂.ChallengeIdx) :
+    HEq ((append D₁ D₂).decompose (ChallengeIdx.inr i₂)) (D₂.decompose i₂) :=
+  append_decompose_heqR D₁ D₂ (ChallengeIdx.sumEquiv_symm_inr i₂)
+
+end AppendChar
+
+variable {r : ℕ} {len : Fin r → ℕ} {pSpec : ∀ i, ProtocolSpec (len i)}
+
+/-- Sequential composition of a finite family of CWSS structures. -/
```
alexanderlhicks, line 190: You could probably do induction over this and generalize append_treeSpecialSound beyond the binary append case.

alexanderlhicks (dismissed):
Looks good! Left two comments but they can be addressed via small follow up PRs if applicable rather than holding this up any longer.

alexanderlhicks (approved):

PR #531, opened 2026-06-04
Add `evalOnPoints` simp lemmas and prove it preserves multiplication

ElijahVlasov:
Add `evalOnPoints` simp lemmas and prove it preserves multiplication.


alexanderlhicks (approved):

alexanderlhicks:
hmm maybe the declarations tracker I added is a bit too annoying right now.

PR #501, opened 2026-05-12
feat(polynomial): add evaluation-level lemmas for splitNth and foldNth

NhlanhlaHasane:
Add evaluation-level lemmas for splitNth and foldNth to enhance polynomial evaluation capabilities. These lemmas address gaps in existing definitions and support the verification of Plonky3's FRI folding operation.

PR #537, opened 2026-06-06
x: Sequential composition General.lean

NhlanhlaHasane:
Updates to OracleReduction/Composition/Sequential/General.lean

## Related PR
Depends on / related to #501


quangvdao:
Closing. Despite the title, this changes only `OracleReduction/Basic.lean` and `Data/Polynomial/SplitFold.lean`, with the same content as #500 and #501, which were closed. Both files have since moved to the module system (#897), so the diff no longer applies. Thanks for the contribution.

Issue #450, opened 2026-04-08
Add evaluation-level lemmas for `splitNth` and `foldNth` (even/odd decomposition)


klausnat:
I was verifying Plonky3's FRI folding operation (extracted from Rust to Lean via Aeneas) against ArkLib specifications, and I noticed that some evaluation-level lemmas are missing.

I would be happy to submit several PRs addressing these:

`Data/Polynomial/SplitFold.lean` defines `splitNth` and `foldNth` with
coefficient-level identities (`splitNth_def`) and degree bounds
(`foldNth_degree_le`), but has no evaluation-level lemmas.

In particular, the following are missing:

```lean
lemma splitNth_eval_comp_pow (f : 𝔽[X]) (s : 𝔽) (i : Fin n) :
    (eval₂ C (X ^ n) (splitNth f n i)).eval s = (splitNth f n i).eval (s ^ n)

-- Even/odd evaluation identities (n = 2 specialization)
lemma splitNth_two_eval_add (f : 𝔽[X]) (x : 𝔽) :
    f.eval x + f.eval (-x) = 2 * (splitNth f 2 0).eval (x ^ 2)

lemma splitNth_two_eval_sub (f : 𝔽[X]) (x : 𝔽) :
    f.eval x - f.eval (-x) = 2 * x * (splitNth f 2 1).eval (x ^ 2)

-- Main result: evaluation of foldNth in terms of f(x) and f(-x)
lemma foldNth_two_eval (f : 𝔽[X]) (x β : 𝔽) (hx : x ≠ 0) (h2 : (2 : 𝔽) ≠ 0) :
    (foldNth 2 f β).eval (x ^ 2) =
    (f.eval x + f.eval (-x) + β * (f.eval x - f.eval (-x)) * x⁻¹) * (2 : 𝔽)⁻¹
```

Thanks, 
Nat

alexanderlhicks:
https://github.com/Verified-zkEVM/ArkLib/pull/452 includes splitNth_eval_comp_pow  but not the other three. Are you planning to add them, would like us to add them, or did they just end up not being needed for the rust-lean examples in the end?

lalalune:
I'm working on this issue.

lalalune:
[...]

PR #548, opened 2026-06-08
feat(SplitFold): add even/odd evaluation lemmas for splitNth

klausnat:
## Motivation

In this PR we add evaluation-level lemmas relating `splitNth`, to connect the spec to FRI-style folding, where the even/odd split shows up as f(x) vs f(-x).

## Changes

Adds three lemmas (`n = 2` specialization):

- `splitNth_two_eval` is reusable pointwise even/odd decomposition.
- `splitNth_two_eval_add`
- `splitNth_two_eval_sub`

Built on the existing `splitNth_eval_comp_pow`.

## Notes

As for fourth lemma from the issue #450, `foldNth_two_eval`, we should first define `foldNth`
(mentioned in the module docstring, but not defined).

Solves partially #450

### DIFF
--- a/ArkLib/Data/Polynomial/SplitFold.lean
+++ b/ArkLib/Data/Polynomial/SplitFold.lean
@@ -25,7 +25,7 @@ This file defines n-way splitting and folding operations on polynomials.
 
 When `n = 2`, this recovers the even/odd splitting: `splitNth f 2 0` gives the even
 coefficients and `splitNth f 2 1` gives the odd coefficients (after appropriate
-reindexing). 
+reindexing).
 
 -/
 
@@ -254,7 +254,7 @@ lemma splitNth_degree_le {n : ℕ} {f : 𝔽[X]} [inst : NeZero n] :
 lemma folding_polynomial_eq_sum_splitNth {𝔽 : Type} [Field 𝔽]
   {f : Polynomial 𝔽} {n : ℕ}
   [inst : NeZero n] :
-  FoldingPolynomial.foldingPolynomial (X ^ n) f = 
+  FoldingPolynomial.foldingPolynomial (X ^ n) f =
     ∑ i, C (splitNth f n i) * (X ^ i.val) := by
   symm
   apply FoldingPolynomial.folding_polynomial_is_unique'
@@ -263,8 +263,8 @@ lemma folding_polynomial_eq_sum_splitNth {𝔽 : Type} [Field 𝔽]
       rw [splitNth_def (f := f) (inst := inst)]
     rw [
       Polynomial.map_sum,
-      Polynomial.eval_finset_sum] 
-    simp only [Polynomial.map_mul, map_C, coe_compRingHom, Polynomial.map_pow, map_X, 
+      Polynomial.eval_finset_sum]
+    simp only [Polynomial.map_mul, map_C, coe_compRingHom, Polynomial.map_pow, map_X,
     eval_mul, eval_C, eval_pow, eval_X]
     simp only [comp]
     conv =>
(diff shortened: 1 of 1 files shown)


ElijahVlasov:
Is this in any way better than what was done here long ago (and was removed): https://github.com/Verified-zkEVM/ArkLib/blob/181ae87d8357219a42b86f9d3ef245726ce64b4a/ArkLib/Data/Polynomial/EvenAndOdd.lean#L451?

ElijahVlasov:
> Is this in any way better than what was done here long ago (and was removed):
> 
> https://github.com/Verified-zkEVM/ArkLib/blob/181ae87d8357219a42b86f9d3ef245726ce64b4a/ArkLib/Data/Polynomial/EvenAndOdd.lean#L451
> 
> ?

And here: https://github.com/Verified-zkEVM/ArkLib/blob/181ae87d8357219a42b86f9d3ef245726ce64b4a/ArkLib/Data/Polynomial/EvenAndOdd.lean#L407

klausnat:
Thanks @ElijahVlasov  

But file `EvenAndOdd` was removed because `SplitFold` replaced it.

In this PR I am adding eval-lemmas in `SplitFold` (filling the gap).

also it was built on a different abstraction (`evenPart`/`oddPart`/`evenize`) rather than `splitNth`, which is what `SplitFold.lean` uses today.

Also there were no `f(x) ± f(-x)` evaluation identities, it had only lower-level pieces (`even_eval`, `evenPart_x_eval_eq`). 

in my PR they are in the FRI-ready form on the current `splitNth` API, and `SplitFold.lean` had no
evaluation-level lemmas at all before.

aryaethn:
These three even/odd lemmas are exactly what #450 needs. Two things that might help move it along:

**1. Looks like it needs a rebase.** The diff predates a recent change to this file — `main` now uses `Polynomial.eval_finsetSum` (the `eval_finset_sum` spelling in the diff context no longer exists), so it likely won't merge cleanly as-is. The few whitespace-only reversions on otherwise untouched lines could also be dropped to keep the diff focused on the new lemmas.

**2. The remaining #450 item (`foldNth_two_eval`) follows directly from these.** The `foldNth` named in the module docstring is realized by `FoldingPolynomial.polyFold` — `polyFold_eq_sum_of_splitNth` already gives `polyFold f n r = ∑ i, C (rⁱ) * splitNth f n i` — so the main result composes with your `splitNth_two_eval_add` / `splitNth_two_eval_sub`. I checked this compiles on top of your branch (current `main`, `[Field 𝔽]`):

```lean
lemma foldNth_two_eval (f : 𝔽[X]) (x β : 𝔽) (hx : x ≠ 0) (h2 : (2 : 𝔽) ≠ 0) :
    (FoldingPolynomial.polyFold f 2 β).eval (x ^ 2) =
      (f.eval x + f.eval (-x) + β * (f.eval x - f.eval (-x)) * x⁻¹) * (2 : 𝔽)⁻¹ := by
  rw [polyFold_eq_sum_of_splitNth, eval_finsetSum, Fin.sum_univ_two]
  simp only [Fin.val_zero, Fin.val_one, pow_zero, pow_one, eval_mul, eval_C, one_mul]
  rw [splitNth_two_eval_add f x, splitNth_two_eval_sub f x]
  field_simp
```

Happy to send this as a follow-up PR once yours lands, or you're welcome to fold it in here — whichever you prefer.

(Minor, optional: the module docstring lists `foldNth n f α` under "Main definitions", but there's no such `def` yet — only `polyFold`. Might be worth adding a `foldNth` alias or adjusting the docstring so the name resolves.)

quangvdao:
I merged current `main` into this branch so it could land after the module-system migration (#897). `SplitFold.lean` now takes main's version and re-adds the three lemmas against the renamed `eq_sum_splitNth` and `eval_finsetSum`. I also added @aryaethn's `foldNth_two_eval` from the review above as `polyFold_two_eval`, since `polyFold` is the existing name for the fold. `./scripts/validate.sh` passes locally. Thanks @klausnat, I'll merge this once CI is green.

PR #562, opened 2026-06-08
Remove `codewordToPoly` and change `decode` to `toPolynomial`

ElijahVlasov:
I'm accountable for introducing `codewordToPoly` which turns out to be a duplicate of the `decode` function from this very file. So I'm taking courage to remove it for good since `decode` has more facts proven about it.

I also believe that `decode` is a morally misleading term here as it might be associated with RS decoding procedures such as Gursuwami-Sudan or Berlekamp-Welch whilst what we have here is something rather tautological because we always know there is always such a polynomial. That's why `ReedSolomon.toPolynomial` reads slightly better IMO.


alexanderlhicks:
yep I think this terminology is clearer, thanks!

alexanderlhicks (dismissed):

alexanderlhicks (approved):

PR #569, opened 2026-06-10
fix(oracle-reduction): close vacuity in straightline knowledge soundness

alexanderlhicks:
`Verifier.knowledgeSoundness` was vacuously satisfiable: the straightline extractor returns `OptionT (OracleComp oSpec) WitIn`, and the knowledge game bound it inside the surrounding `OptionT` execution. In VCV-io semantics, `OptionT` failure contributes to the failure mass of the computation, which `probEvent` excludes (it only measures `some '' {x | p x}`). Hence the always-failing extractor `fun _ _ _ _ _ => failure` drove the game's bad-event probability to exactly 0, discharging knowledge soundness at error 0 for ANY verifier and ANY relations. (Verified by a Lean proof of `verifier.knowledgeSoundness init impl relIn relOut 0` for arbitrary `verifier`, `relIn`, `relOut`, depending only on the standard axioms; reproduced in the PR description.)

Fix: run the extractor's `OptionT` layer explicitly (`.run`) and keep the resulting `Option WitIn` in the game's output, changing the bad event to

  (∀ extractedWitIn ∈ extractedWitIn?, (stmtIn, extractedWitIn) ∉ relIn) ∧
    (stmtOut, witOut) ∈ relOut

so extractor failure now counts as an adversary win, matching the cryptographic definition (the extractor must succeed whenever the prover convinces the verifier). This keeps `Extractor.Straightline`'s type, the definition's signature, and the error type unchanged, so no downstream statement changes; failures of the protocol execution itself (verifier aborts) remain excluded, consistent with plain soundness. The rbr variant (`rbrKnowledgeSoundness`) uses a pure-function extractor and is unaffected.

Also prove the previously-sorried `Verifier.id_knowledgeSoundness` (via `Extractor.Straightline.id`), which both removes a sorry and witnesses that the strengthened definition remains satisfiable; as a result `OracleVerifier.id_knowledgeSoundness` is now sorry-free as well (both depend only on [propext, Classical.choice, Quot.sound]).

Blueprint prose for `def:knowledge_soundness` updated to match.

Known follow-up (out of scope): `Verifier.StateRestoration.knowledgeSoundness` has the same trap (its extractor's `OracleComp` may fail inside the game).

### DIFF
--- a/ArkLib/OracleReduction/Security/Basic.lean
+++ b/ArkLib/OracleReduction/Security/Basic.lean
@@ -267,10 +267,24 @@ class IsSound (langIn : Set StmtIn) (langOut : Set StmtOut)
   - there exists a straightline extractor `E`, such that
   - for all input statement `stmtIn`, witness `witIn`, and (malicious) prover `prover`,
   - if the execution with the honest verifier results in a pair `(stmtOut, witOut)`,
-  - and the extractor produces some `witIn'`,
 
-  then the probability that `(stmtIn, witIn')` is not valid and yet `(stmtOut, witOut)` is valid
-  is at most `knowledgeError`.
+  then the probability that `(stmtOut, witOut)` is valid and yet the extractor fails to produce
+  a witness `witIn'` such that `(stmtIn, witIn')` is valid is at most `knowledgeError`.
+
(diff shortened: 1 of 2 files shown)


quangvdao:
thanks!

PR #559, opened 2026-06-08
feat(Data/Lattices): Hachi section 3 (more ring stuff)

tobias-rothmann:
## Summary

Formalizes Section 3 of Hachi (*Efficient Lattice-Based Multilinear Polynomial Commitments over Extension Fields*, [NOZ26]): the subfield theory of the power-of-two cyclotomic ring `R_q = Z_q[X]/(X^d + 1)` (`d = 2^α`), which the paper uses to embed extension-field statements into `R_q`. The development lives under `ArkLib/Data/Lattices/CyclotomicRing/`, aggregated by `Subfield.lean` (whose module docstring carries a per-result correspondence to the paper, the deviations, and a file map).

## What's formalized

**Galois layer for `R_q`** (`CyclotomicRing/Galois/`): the automorphisms `σ_i : X ↦ X^i` (`galoisAut`, `Automorphism.lean`), the group `H = ⟨σ_{-1}, σ_{4k+1}⟩` and its order facts (`Hexp`, `Hexp_card`, `Group.lean`), the trace `Tr_H` (`traceH`, `Trace.lean`), and the fixed subring `R_q^H` (`fixedSubring`, `FixedSubring.lean`).

**Lemma 5** — `R_q^H` is a field isomorphic to `F_{q^k}`:
- the symmetric `Z_q`-basis and the cardinality `|R_q^H| = q^k` (Eq. 7) — `vElt`/`vElt_coeff` in `Subfield/Basis.lean`, `card_fixedSubring_eq` in `Subfield/Cardinality.lean`;
- the field structure and isomorphism — `fixedSubring_isField`, `fixedSubringEquivGaloisField` in `Subfield/Field.lean`, with the `q ≡ 5 (mod 8)` number theory in `Subfield/Factorization.lean`.

**Theorem 2** — the packing map and trace identity:
- `ψ : (R_q^H)^{d/k} → R_q` (Eq. 8) — `psi` in `Subfield/Packing.lean`;
- the trace formula `Tr_H(ψ(a)·σ_{-1}(ψ(b))) = (d/k)·⟨a,b⟩`, which turns inner products over `F_{q^k}` into trace statements over `R_q` — `traceH_psi_mul_conj` in `Subfield/TraceInnerProduct.lean`, resting on the monomial-trace vanishing Claims 2 & 3 (`traceH_Xpow_eq_zero`, `traceH_Xpow_half`) in `Subfield/TraceVanishing.lean`;
[...]

PR #582, opened 2026-06-15
refactor(commitments): reorganize into ordinary and functional commitments

tobias-rothmann:
### TDLR

Just reordering commitments into functional commitments (with evaluation) and standard/ordinary commitments (without evaluation)

### Motivation

The commitment formalizations had grown organically under a single
`ArkLib/CommitmentScheme/` tree that mixed two fundamentally different notions:

- **Ordinary** commitments — opening reveals the full message and verification
  recomputes the commitment (e.g. the simple Ajtai/SIS commitment).
- **Functional** commitments — opening reveals an *oracle evaluation* of hidden
  data rather than the message itself (e.g. KZG polynomial commitments, the
  Hachi inner–outer lattice commitment).

This PR splits the tree along that distinction so the two families have clear,
separate entry points, and adds a canonical `Basic` handle for each.

### What changed

- Moved `ArkLib/CommitmentScheme/` → `ArkLib/Commitments/`, split into:
  - `Commitments/Ordinary/` — `Ajtai/Simple/*` (the SIS commitment).
  - `Commitments/Functional/` — `KZG/*` and `Hachi/*` (the renamed
    `Ajtai/InnerOuter/*` + `Gadget`/`GadgetNorms`).
- Renames are content-preserving; the only edits to moved files are updated
  import paths. No definitions, statements, or proofs changed.

### New files

- `Commitments/Ordinary/Basic.lean` — canonical entry point for ordinary
  (non-interactive) commitments. Re-exports VCV-io's
  `CommitmentScheme PP M C D` and its security notions (perfect correctness/
  hiding, game-based hiding/binding experiments, trapdoor extraction) rather
  than redefining them.
- `Commitments/Ordinary/SimpleRO.lean` — a simple random-oracle commitment
  expressed as an ordinary `CommitmentScheme`: the public parameter is a
  uniformly sampled oracle `ro : α × β → γ`; committing to `v` samples
  randomness `r` and outputs `(ro (v, r), r)`; verification recomputes
  `ro (v, r)`. Mirrors VCV-io's standard-model hash commitment with a random
  oracle in place of a keyed hash family.


tobias-rothmann:
@alexanderlhicks CI seems a bit broken. 
Build fails if one generates the `docs/kb` files because it's not allowed, but if one doesn't, the Docs Integrity check fails, because the files have moved and "don't exist anymore".  Ready for review once this is fixed.

alexanderlhicks:
The CI issue with the kb can be ignored somewhat -- it's not completely breaking if it drifts and can be regenerated -- and I need to get around to spending time on a branch to fix this more than anything.

Going to need to resolve the conflicts that came from merging #559 too, though!

tobias-rothmann:
> Going to need to resolve the conflicts that came from merging #559 too, though!

Will do that right now!

ArkLib/Commitments/Ordinary/SimpleRO.lean
```diff
@@ -0,0 +1,60 @@
...
+import ArkLib.Commitments.Ordinary.Basic
+import VCVio
+
+/-!
+  # Simple Random-Oracle Commitment Scheme
+
+  A simple non-interactive commitment scheme in the random-oracle model, expressed as an
+  *ordinary* commitment scheme (VCV-io's `CommitmentScheme`):
+
+  - The public parameter is the random oracle itself, modeled as a uniformly sampled function
```
alexanderlhicks, line 16: Mixing random oracle and standard model terminology in the docstring is a bit confusing I think, there are no oracle queries in the standard model as everyone (adversary included) has full access to the public parameter.

ArkLib/Commitments/Ordinary/SimpleRO.lean
```diff
@@ -0,0 +1,60 @@
...
+  hash family. For commitments with oracle openings, see `ArkLib.Commitments.Functional.Basic`.
+-/
+
+open OracleComp CommitmentScheme
+
+namespace SimpleRO
+
+variable {α β γ : Type}
+
+/-- A random oracle, modeled as a function from queries `α × β` to responses `γ`. -/
```
alexanderlhicks, line 44: See above above mixing rom and standard model terminology; there are no queries here. 
The old formalization that this PR replaces did rely on oraclespec/queries but the impact of this PR is that this is no longer the case.

ArkLib/Commitments/Ordinary/SimpleRO.lean
```diff
@@ -0,0 +1,60 @@
...
+abbrev RandomOracle (α β γ : Type) := α × β → γ
+
+/-- Commit to message `v` under the random oracle `ro` and randomness `r` by hashing `(v, r)`. -/
+def commit (ro : RandomOracle α β γ) (v : α) (r : β) : γ := ro (v, r)
+
+/-- Verify an opening `r` of the commitment `cm` to message `v` by recomputing the hash. -/
+def verify [DecidableEq γ] (ro : RandomOracle α β γ) (v : α) (cm : γ) (r : β) : Bool :=
+  decide (commit ro v r = cm)
+
+/-- The simple random-oracle commitment as an (ordinary) `CommitmentScheme`.
```
alexanderlhicks, line 54: It might be good to signpost what security properties this definition supports / have them linked to or defined in the same file.

alexanderlhicks:
Also thinking about how to frame this with respect to future work on e.g. FRI as PCS (via FRI IOPP -> Merkle Tree commitments in VCV -> BCS). Is it worth having some positional/vector commitment folder here as well in anticipation of this? Right now it would just link to VCV's vector commitment which makes it redundant but it feels like something to think abour organising correctly to pre-empt some future refactor.

ArkLib/Commitments/Ordinary/SimpleRO.lean
tobias-rothmann: Yes right. 
This definition is a bit weird in general.. 
It's a RO commitment scheme without RO. 
Not sure it's even sensible to keep it.

tobias-rothmann: What we could do is define a new ordinary commitment scheme in ArkLib that supports `OracleComp` not just `ProbComp` (what VCV enforces I think). But that would duplicate ofc. I guess the most sensible thing is to maybe extend VCV's `CommitmentScheme` to work with oracles? @alexanderlhicks

tobias-rothmann:
> Also thinking about how to frame this with respect to future work on e.g. FRI as PCS (via FRI IOPP -> Merkle Tree commitments in VCV -> BCS). Is it worth having some positional/vector commitment folder here as well in anticipation of this? Right now it would just link to VCV's vector commitment which makes it redundant but it feels like something to think abour organising correctly to pre-empt some future refactor.

Yes! Right now that all falls into functional, but I could divide further into PCS, VCs, and thats it for now right?

alexanderlhicks:
> > Also thinking about how to frame this with respect to future work on e.g. FRI as PCS (via FRI IOPP -> Merkle Tree commitments in VCV -> BCS). Is it worth having some positional/vector commitment folder here as well in anticipation of this? Right now it would just link to VCV's vector commitment which makes it redundant but it feels like something to think abour organising correctly to pre-empt some future refactor.
> 
> Yes! Right now that all falls into functional, but I could divide further into PCS, VCs, and thats it for now right?

Yep. The compiled non-interactive (i.e. post BCS) PCS would fit in the functional folder, but BCS would consume a vector commitment. Maybe it's fine for BCS to just pull from VCV though... It's ok for now at least I think.

alexanderlhicks (approved):

PR #592, opened 2026-06-26
ci: keep build timing report non-blocking when PR comment fails

ainta:
## Summary
- Grant the `Build Timing Report` `report` job the token permissions it actually uses: `actions: read`, `contents: read`, and `pull-requests: write`.
- Remove the workflow-level `issues: write` / `pull-requests: read` block, and scope the requested permissions to the reporting job.
- Keep the PR comment upsert resilient for expected GitHub comment API failures by logging the HTTP status/message and leaving the rendered report in the job summary.
- Let unexpected JavaScript errors and unexpected API statuses fail the job so real regressions stay visible.

## Context
The report renderer succeeds, but the final comment-upsert step can fail with:

```
POST /repos/Verified-zkEVM/ArkLib/issues/<n>/comments - 403
##[error]Unhandled error: HttpError: Resource not accessible by integration
```

#632 identified the permission side of the same failure: for comments on pull requests, the workflow should request `pull-requests: write`. This PR now folds that permission correction into the existing resilience work, while keeping the rendered timing report available through `$GITHUB_STEP_SUMMARY` if GitHub still rejects the comment API call.

The updated diagnosis is that the workflow should request the correct job-level `GITHUB_TOKEN` scope directly. If an organization/repository setting or rate limit still prevents the comment operation, the workflow emits a warning containing `HTTP <status>: <message>` and does not turn an otherwise-green CI run red for the cosmetic comment.

## Verification
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/build-timing-report.yml'))"`
- `git diff --check`
- Extracted the `Upsert build timing PR comment` `github-script` body and syntax-checked it with bundled Node after wrapping it in an async function.

Posted by Codex on behalf of the user (@ainta) with approval.

### DIFF
--- a/.github/workflows/build-timing-report.yml
+++ b/.github/workflows/build-timing-report.yml
@@ -247,26 +247,47 @@ jobs:
             const body = fs.readFileSync(process.env.BUILD_TIMING_COMMENT, 'utf8');
             const { owner, repo } = context.repo;
             const issue_number = Number('${{ steps.timing-context.outputs.pr-number }}');
-            const comments = await github.paginate(
-              github.rest.issues.listComments,
-              { owner, repo, issue_number, per_page: 100 }
-            );
-            const existing = comments.find(comment =>
-              comment.user?.type === 'Bot' && comment.body?.includes(marker)
-            );
 
-            if (existing) {
-              await github.rest.issues.updateComment({
-                owner,
-                repo,
-                comment_id: existing.id,
-                body,
-              });
-            } else {
-              await github.rest.issues.createComment({
-                owner,
-                repo,
-                issue_number,
-                body,
-              });
+            try {
+              const comments = await github.paginate(
+                github.rest.issues.listComments,
+                { owner, repo, issue_number, per_page: 100 }
+              );
+              const existing = comments.find(comment =>
+                comment.user?.type === 'Bot' && comment.body?.includes(marker)
+              );
+
+              if (existing) {
+                await github.rest.issues.updateComment({
+                  owner,
+                  repo,
+                  comment_id: existing.id,
+                  body,
+                });
+              } else {
+                await github.rest.issues.createComment({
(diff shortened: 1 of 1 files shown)


quangvdao:
Thanks for diagnosing the noisy failure. I’d like to retain the PR comment, so please revise this PR to incorporate the permission correction while keeping the reporting step resilient:

- change `pull-requests: read` to `pull-requests: write`;
- remove `issues: write` so the workflow has only the write scope it actually uses;
- preferably scope these permissions to the `report` job;
- keep expected comment API failures non-blocking and emit the HTTP status/message;
- do not catch every exception indiscriminately—unexpected JavaScript/programming errors should still fail so genuine regressions remain visible;
- update the PR explanation: GitHub permits workflow-level `permissions` to override the repository’s restricted default, so the current “default token cap” diagnosis is not quite right.

I’ve closed #632 in favor of consolidating both fixes here.

ainta:
Addressed in c60e387552fd8bf71a131577571b5e8ba836b900.

Changes made:
- moved the workflow permissions down to the `report` job;
- changed the PR permission to `pull-requests: write` and removed `issues: write`;
- narrowed the non-blocking handling to expected comment API failures, with `HTTP <status>: <message>` in the warning;
- left unexpected JavaScript errors and unexpected API statuses to fail the job;
- updated the PR body to remove the earlier repository-default-token-cap diagnosis and reflect the consolidated #632 permission fix.

Local checks run:
- `python3 -c "import yaml; yaml.safe_load(open(\".github/workflows/build-timing-report.yml\"))"`
- `git diff --check`
- extracted the `github-script` body and syntax-checked it with bundled Node after wrapping it in an async function.

The lightweight PR checks have passed; the full build/docs jobs are still in progress.

quangvdao (approved):

PR #599, opened 2026-07-01
Introduce blocks as in the WHIR paper

ElijahVlasov:
This PR introduces the definition of a block into the `Domain` library along with a few useful lemmas.
The definition is taken from the WHIR paper: 
image

### DIFF
--- a/ArkLib/Data/Domain/CosetFftDomain/Block.lean
+++ b/ArkLib/Data/Domain/CosetFftDomain/Block.lean
@@ -66,6 +66,10 @@ lemma block_x_0 :
 lemma block_k_0 :
   block ω 0 x = if x = 1 then toFinset ω else ∅ := by aesop
 
+@[simp]
+lemma block_k_1 :
+  block ω 1 x = if x ∈ ω then {x} else ∅ := by aesop
+
 /-- An alternative definition of `block` in terms of
   `Polynomial.nthRootsFinset`. -/
 lemma block_eq_nthRootsFinset [NeZero k] :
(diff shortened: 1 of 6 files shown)


ArkLib/Data/Domain/CosetFftDomain/Subdomain.lean
alexanderlhicks: card roots is gone! worth doing a quick pass over these docstrings.

ArkLib/Data/Domain/CosetFftDomain/Block.lean
```diff
@@ -0,0 +1,147 @@
...
+
+variable {n : ℕ}
+variable {D : Type} [FunLike D ι F] [CosetFftDomainClass D ι F]
+variable {ω : D} {k : ℕ} {x y : F}
+
+open Finset Polynomial
+
+/-- The `k`th roots of `x` from the domain `ω`.
+
+  This is the definition 4.16 from [ACFY24]. Note, for simplicity of
```
alexanderlhicks: wondering if there's a good way to flag this more explicitly to avoid anyone reusing the deifnition naively thinking they are pulling straight from the paper.

alexanderlhicks (dismissed):

ArkLib/Data/Domain/CosetFftDomain/Block.lean
ElijahVlasov: Yes, I've fallen into this trap myself a few times

ElijahVlasov: I've reworked blocks and folding to only deal with powers of 2 to match the paper(s) precisely.

alexanderlhicks (approved):

PR #602, opened 2026-07-01
CWSS protocol infrastructure

tobias-rothmann:
## Prepare primitive IORs for CWSS composition

- Add generic CWSS infrastructure for no-challenge protocols and n-ary sequential composition of pure verifiers.
- Prove purity propagation through `Verifier.id`, binary append, and `seqCompose`.
- Refactor/prove CWSS support for core proof-system components: `SendWitness`, `SendClaim`, `CheckClaim`, and `ReduceClaim`.
- Add `SendChallenge` as the verifier-challenge component for fold-style protocols, including its CWSS structure.
- Wire the new modules into the aggregate imports.


alexanderlhicks:
AI-assisted review note: the feedback below was generated from a detailed review conducted collaboratively with the repository reviewer and has been reviewed by the human reviewer.

The new pure/no-challenge/sequential CWSS infrastructure builds cleanly and the inspected theorems are axiom-clean. No concrete soundness bug was established. The actionable feedback is limited to the intentional CheckClaim semantic replacement, its stale documentation, and the completeness obligation for the replacement relation-level design. This is not a request to preserve an obsolete effectful implementation.

Suggested location-specific feedback:

**ArkLib/ProofSystem/Component/CheckClaim.lean:183**

Moving `pred` out of the runtime `guard` and into `oracleRelOut` is what enables `IsPure`, but it changes the public reduction contract: the old oracle verifier could fail after executing an oracle computation, while the new verifier always returns the statement and relies on relation membership later. If the old abstraction is intentionally obsolete, please do not preserve it as canonical. Correct the stale module-level documentation, state the breaking migration, and update the incomplete consumers/examples. Please also add the pure replacement's completeness theorem with the explicit hypothesis that every `relIn` input satisfies `P` (or change the intended input relation accordingly). If equivalence in the no-failure setting is intended, add that theorem too; a compatibility implementation is optional.

**ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/SeqCompose.lean:395**

Non-blocking composition-contract suggestion: please consider adding one small end-to-end usage theorem for the new `IsPure`/no-challenge/`seqCompose` stack using the actual Hachi component shapes. The generic theorems compile, but a concrete composition example would lock down message ordering, relation transport, and the required `Nonempty` assumptions before more Hachi stages depend on this interface.

alexanderlhicks (approved):

PR #609, opened 2026-07-06
refactor(Commitments): abstract scheme into a section variable

ErVinuelas:
## Motivation

In `ArkLib/Commitments/Functional/Basic.lean`, the security definitions
(`correctness`, `perfectCorrectness`, `bindingGame`, `bindingExperiment`, `binding`,
and `extractability`) each re-declared the same explicit parameter

```lean
(scheme : Scheme oSpec Data Commitment Decommitment ComKey VerifKey pSpec)
```

This repetition is boilerplate: the `scheme` is fixed across the whole section and
shared by every definition.

## What this changes

Hoists `scheme` into the section-level `variable` block alongside the other shared
context (`init`, `impl`, the oracle/protocol instances). Each definition now picks it
up implicitly, removing the repeated binder. Call sites that were passing `scheme`
positionally are updated to match the new argument order.

This is a pure refactor — no definitions, statements, or proofs change in meaning.

## Contrast with previous behavior

- **Before:** every security definition explicitly bound its own `scheme` parameter.
- **After:** `scheme` is a shared `variable`, declared once for the section.

### DIFF
--- a/ArkLib/Commitments/Functional/Basic.lean
+++ b/ArkLib/Commitments/Functional/Basic.lean
@@ -79,14 +79,14 @@ variable [DecidableEq ι]
   [∀ i, VCVCompatible (pSpec.Challenge i)]
   [∀ i, SampleableType (pSpec.Challenge i)]
   {σ : Type} (init : ProbComp σ) (impl : QueryImpl oSpec (StateT σ ProbComp))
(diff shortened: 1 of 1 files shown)


tobias-rothmann (approved):
LGTM!

alexanderlhicks (approved):

alexanderlhicks:
Thanks!

PR #596, opened 2026-06-30
AffineGeneratorsWithMCA

katyhr:

PR #610, opened 2026-07-06
Polynomial Generators have MCA for RSCodes

katyhr:
This PR formalises 
- MCA error function for MDS codes (appears in the statement of Theorem 6.1 [BCGM25]).
- Statement of `Theorem 6.1` (MCA for MDS generators) [BCGM25].

- `Definition 8.1 (MCA error for univariate powers)`  [BCGM25].
- `Theorem 8.2` (`polynomial_gen_MCA`) [BCGM25] (statement and proof).

- `Definition 9.1`  [BCGM25].
- Statement  of `Lemma 9.3`  (`univariate_powers_MCA`) [BCGM25].
- `Theorem 9.2`  (`polynomial_gen_MCA_RScode`) [BCGM25] (statement and proof assuming `Lemma 9.3`).

Interesting outcomes: 

- We prove an s-fold tensor product lemma `tensor_of_univ_is_MCA` that is assumed in the proof of `Theorem 9.2.` In fact, we prove a more general version of this result. In the paper, the result is stated only for a Reed-Solomon code. We prove this for a univariate generator of any code!
- Additionally, we explicitly contract the matrix `A` that appears in the proof of `Theorem 9.2`.


ArkLib/Data/CodingTheory/ProximityGenerator/PolynomialGenerator.lean
```diff
@@ -0,0 +1,452 @@
...
+      = ReedSolomon.code (Function.Embedding.subtype (· ∈ s)) (d + 1) := by
+  have hMG : M_G (UnivariatePowersOn s d)
+      = Vandermonde.nonsquare (d + 1) ((Function.Embedding.subtype (· ∈ s)) : s → F) := by
+    ext x j
+    simp [M_G, UnivariatePowersOn, Vandermonde.nonsquare]
+  rw [hMG, genMatIsVandermonde]
+
+/-- The restricted univariate powers generator is an MDS generator (its code is Reed–Solomon,
+hence MDS). -/
+lemma UnivariatePowersOn.isMDSGenerator [DecidableEq F] (s : Set F) [Fintype s] [Nonempty s]
```
alexanderlhicks, line 336: quick note that ismDS will require [Inhabited ι] now that #663 has merged, which will allow you to remove (h : d + 1 ≤ Fintype.card ↥s)

ArkLib/Data/CodingTheory/ProximityGenerator/PolynomialGenerator.lean
```diff
@@ -0,0 +1,452 @@
...
+        simp only [Matrix.vecMul, dotProduct]
+        rw [Fintype.sum_subsingleton _ j]
+        simp [UnivariatePowersOn]
+      rwa [hvec] at hmem
+    rw [prob_uniform_eq_ofReal, Finset.filter_false_of_mem fun x _ => hfalse x]
+    simp
+  · have hℓ : 2 ≤ Fintype.card (Fin (d + 1)) := by rw [Fintype.card_fin]; omega
+    have hdim : LinearCode.dim (fromColGenMat (M_G (UnivariatePowersOn s d)))
+        = Fintype.card (Fin (d + 1)) := by
+      rw [UnivariatePowersOn.code_eq_reedSolomon, ReedSolomon.dim_eq_deg_of_le' h, Fintype.card_fin]
```
alexanderlhicks, line 370: also a note that ReedSolomon.dim_eq_deg_of_le' can be replaced by ReedSolomon.dim_eq_deg_of_le now.

ArkLib/Data/CodingTheory/ProximityGenerator/PolynomialGenerator.lean
```diff
@@ -0,0 +1,452 @@
...
+  Matrix
+
+variable {F : Type} [Field F]
+         {ι : Type} [Fintype ι]
+         (k : ℕ) -- degree of the polynomials
+         (D : ι ↪ F) -- the domain of evaluation
+
+
+/-- Definition 9.1 MCA error function for Reed-Solomon codes [BCGM25]. -/
+noncomputable def ε_mca_RS [Fintype F] (n d m : ℕ) : I → ℝ :=
```
alexanderlhicks, line 156: n is the block length so it should be constrained to be Fintype.card ι (line 47 above exemplifies this).
Same in other statements below.

ArkLib/Data/CodingTheory/ProximityGenerator/PolynomialGenerator.lean
```diff
@@ -0,0 +1,452 @@
...
+  let ρ_sqrt := ReedSolomon.sqrtRate k D
+  fun γ =>
+    if γ ≤ 1 - (1 + (1 / (2 * m : ℝ))) * ρ_sqrt then
+      (|Fintype.card F| : ℝ)⁻¹  *  (m + 1 / 2) ^ 7  * (3 * (ρ_sqrt) ^ 3)⁻¹.toReal * d * n ^ 2
```
alexanderlhicks, line 160: careful with this because  ρ_sqrt := ReedSolomon.sqrtRate is R >= 0, and lean will allow division by 0 which is implicitely gaurded against in the paper definition. k > 0 should prevent this?

alexanderlhicks, line 160: |Fintype.card F| can just be Fintype.card F

ArkLib/Data/CodingTheory/ProximityGenerator/PolynomialGenerator.lean
```diff
@@ -0,0 +1,452 @@
...
+    else
+      1
+
+/-- Let `F` be a field and `k, n, d, m ∈ ℕ` with `m ≥ 3`. Then the univariate powers generator has
+MCA for a Reed-Solomon code over a domain `D` and degree `k` with error `ε_mca_RS` as defined
+above.
+Lemma 9.3 [BCGM25]. -/
+lemma univariate_powers_MCA [Fintype F] (n d m : ℕ) (hm : 3 ≤ m) :
```
alexanderlhicks, line 168: as above for k and n

ArkLib/Data/CodingTheory/ProximityGenerator/PolynomialGenerator.lean
```diff
@@ -0,0 +1,452 @@
...
+    congr 1
+    rw [← Matrix.toLin'_apply', LinearMap.toMatrix'_toLin']
+  rw [hg, LinearMap.toMatrix'_id] at hcomp
+  exact hcomp.symm
+
+/-- Take a Reed–Solomon code of polynomials of degree less than `k` over an evaluation domain
+`D`. Let `G` be a polynomial generator where each `S` is the whole field `F`.
+Then, for every `m ≥ 3`, `G` has MCA for with error `∑ i, ε_mca_RS`.
+Theorem 9.2 [BCGM25]. -/
+lemma polynomial_gen_MCA_RScode [Fintype F] (n m : ℕ) (hm : 3 ≤ m) {ℓ : Type} [Fintype ℓ] {s : ℕ}
```
alexanderlhicks, line 288: k and n

ArkLib/Data/CodingTheory/ProximityGenerator/MCAGenerator.lean
```diff
@@ -0,0 +1,254 @@
...
+  -- Combine: the tensor MCA event implies one of the two, then union + additivity.
+  refine le_trans (Pr_le_Pr_of_implies ($ᵖ (S × S')) _ _ himp) ?_
+  refine le_trans (Pr_or_le ($ᵖ (S × S')) _ _) ?_
+  rw [← hEq]
+  exact add_le_add hA hB
+
+/-- Let `G : S → 𝔽^ℓ` be an MCA generator with error `ε_mca` and `G' : S' → 𝔽^ℓ'` be an MCA
+generator with error `ε_mca'`. Then the (explicit) tensor generator `G ⊗ G' : S × S' → 𝔽^(ℓ × ℓ')`
+is an MCA generator with error `ε_mca + ε_mca'`. -/
+lemma tensor_of_MCA_is_MCA_tight [Nonempty S] {S' : Type} [Fintype S'] [Nonempty S']
```
alexanderlhicks, line 218: this will require generalising IsMCAGenerator to work with interleaved codes (see def 3.3 in the paper which feeds through to lemma 4.4)

ArkLib/Data/CodingTheory/ProximityGenerator/MCAGenerator.lean
```diff
@@ -0,0 +1,254 @@
...
+  calc Pr_{let x' ←$ᵖ S'}[IsMCA (fun x' j' => G (eS x') (eL.symm j')) LC x' U γ]
+      = Pr_{let x' ←$ᵖ S'}[IsMCA G LC (eS x') (fun j => U (eL j)) γ] := Pr_congr hiff
+    _ = Pr_{let x ←$ᵖ S}[IsMCA G LC x (fun j => U (eL j)) γ] := by
+        rw [prob_uniform_eq_ofReal, prob_uniform_eq_ofReal, hcard, Fintype.card_congr eS]
+    _ ≤ ENNReal.ofReal (ε_mca γ) := hGMCA _ γ
+
+/-- Let `G : S → 𝔽^ℓ` be an MCA generator with error `ε_mca` and `G' : S' → 𝔽^ℓ'` be an MCA
+generator with error `ε_mca'`. Then the (explicit) tensor generator `G ⊗ G' : S × S' → 𝔽^(ℓ × ℓ')`
+is an MCA generator. -/
+lemma tensor_of_MCA_is_MCA [Nonempty S] {S' : Type} [Fintype S'] [Nonempty S'] (LC : LinearCode ι F)
```
alexanderlhicks, line 167: do we want to keep this if we resolve tensor_of_MCA_is_MCA_tight ?

docs/kb/papers/BCGM25.md
```diff
@@ -26,16 +26,17 @@ with Reed-Solomon refinements up to the Johnson bound.
...
 - The broader formalization target that polynomial-generator MCA bounds should become reusable
   coding-theory infrastructure rather than WHIR-only assumptions.
 
 ## Main ArkLib Touchpoints
 
-- [`ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean`](../../../ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean)
-- [`ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean`](../../../ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean)
+- [`ArkLib/Data/CodingTheory/ProximityGenerator/Basic.lean`](../../../ArkLib/Data/CodingTheory/ProximityGenerator/Basic.lean)
+- [`ArkLib/Data/CodingTheory/ProximityGenerator/MCAGenerator.lean`](../../../ArkLib/Data/CodingTheory/ProximityGenerator/MCAGenerator.lean)
+- [`ArkLib/ProofSystem/Whir/MutualCorrAgreement.lean`](../../../ArkLib/ProofSystem/Whir/MutualCorrAgreement.lean)
```
alexanderlhicks, line 39: minor but this will mess with the docs CI I guess, can be cleaned up later/when dealing with any merge conflicts.

ArkLib/Data/CodingTheory/Basic/LinearCode.lean
```diff
@@ -257,17 +257,51 @@ scoped macro_rules
...
-def projectedCode (C : Set (ι → F)) (T : Finset ι) : Set (T → F) :=
-  {w | ∃ c ∈ C, w = c|[T]}
-
-notation:60 C "|[" T "]" => projectedCode C T
+def projectedCode [Fintype ι] (C : Set (ι → F)) (T : Finset ι) : Set (T → F) :=
+  {w | ∃ c ∈ C, w = projectedWord c T}
+
+open Submodule
+
+def projectedCode_submod
```
alexanderlhicks, line 270: can be refactored when resolving conflicts with main per the definition change in #596

ArkLib/Data/CodingTheory/Basic/LinearCode.lean
```diff
@@ -257,17 +257,51 @@ scoped macro_rules
 word `c` to `T` as the word obtained by restricting the indexing set of `c` to `T`.
 We denote this by `c|[T]`.
 Definition 3.7 [BCGM25]. -/
-def projectedWord (c : ι → F) (T : Finset ι) : T → F := Set.restrict T c
-
-notation:60 c "|[" T "]" => projectedWord c T
+def projectedWord [Fintype ι] (c : ι → F) (T : Finset ι) : T → F := Set.restrict T c
```
alexanderlhicks, line 260: [Fintype ι] is set above for the section

ArkLib/Data/CodingTheory/Basic/LinearCode.lean
```diff
@@ -257,17 +257,51 @@ scoped macro_rules
...
 
 /-- Let `C` be a code of length `ι`. For every finite `ι`-subset `T`, we define the projected code
 `C|[T]` as the set of projected codewords `c|[T]`, for `c ∈ C`.
 Definition 3.7 [BCGM25]. -/
-def projectedCode (C : Set (ι → F)) (T : Finset ι) : Set (T → F) :=
-  {w | ∃ c ∈ C, w = c|[T]}
-
-notation:60 C "|[" T "]" => projectedCode C T
+def projectedCode [Fintype ι] (C : Set (ι → F)) (T : Finset ι) : Set (T → F) :=
```
alexanderlhicks, line 265: [Fintype ι] is set above for the section

ArkLib/Data/CodingTheory/Basic/LinearCode.lean
```diff
@@ -257,17 +257,51 @@ scoped macro_rules
 word `c` to `T` as the word obtained by restricting the indexing set of `c` to `T`.
 We denote this by `c|[T]`.
```
alexanderlhicks, line 258: minor docstring change after the notation change in #596

ArkLib/Data/CodingTheory/Basic/LinearCode.lean
```diff
@@ -257,17 +257,51 @@ scoped macro_rules
...
 Definition 3.7 [BCGM25]. -/
-def projectedWord (c : ι → F) (T : Finset ι) : T → F := Set.restrict T c
-
-notation:60 c "|[" T "]" => projectedWord c T
+def projectedWord [Fintype ι] (c : ι → F) (T : Finset ι) : T → F := Set.restrict T c
 
 /-- Let `C` be a code of length `ι`. For every finite `ι`-subset `T`, we define the projected code
 `C|[T]` as the set of projected codewords `c|[T]`, for `c ∈ C`.
```
alexanderlhicks, line 263: minor docstring change after the notation change in #596

ArkLib/Data/CodingTheory/ProximityGenerator/PolynomialGenerator.lean
```diff
@@ -0,0 +1,452 @@
...
+
+open unitInterval CoreDefinitions LinearTransformations
+
+variable {F : Type} [Field F]
+         {ι : Type} [Fintype ι] [Nonempty ι]
+
+/-- A function assinging the maximum degree in the `i`-the variable of the collection of
+polynomials `P`. -/
+noncomputable def deg_max {s : ℕ} {ℓ : Type} [Fintype ℓ] (P : ℓ → MvPolynomial (Fin s) F) :
+    Fin s → ℕ := fun i => Finset.sup Fintype.elems (fun j => (P j).degreeOf i)
```
alexanderlhicks, line 39: finset.uni is seemingly more popular/standard than .elems. I don't really know if it's a big deal, but it keeps the definition in terms of finset only.

ArkLib/Data/Probability/Instances.lean
```diff
@@ -451,6 +468,81 @@ lemma Pr_congr {α : Type} {D : PMF α} {P Q : α → Prop}
...
+-/
+theorem Pr_or_le {α : Type} (D : PMF α) (f g : α → Prop) :
+    Pr_{ let r ← D }[ f r ∨ g r ] ≤ Pr_{ let r ← D }[ f r ] + Pr_{ let r ← D }[ g r ] := by
+  classical
+  rw [prob_tsum_form_singleton, prob_tsum_form_singleton, prob_tsum_form_singleton,
+    ← ENNReal.tsum_add]
+  apply ENNReal.tsum_le_tsum
+  intro r
+  rw [← mul_add]
+  refine mul_le_mul_of_nonneg_left ?_ zero_le'
```
alexanderlhicks, line 485: linter says use zero_le

ArkLib/Data/Probability/Instances.lean
```diff
@@ -451,6 +468,81 @@ lemma Pr_congr {α : Type} {D : PMF α} {P Q : α → Prop}
...
+    {c : ENNReal} (h : ∀ b, Pr_{ let a ← Da}[Q a b] ≤ c) :
+    Pr_{ let b ← Db; let a ← Da }[ Q a b ] ≤ c := by
+  classical
+  let D_rest : β → PMF Prop := fun b => (do let a ← Da; return (Q a b))
+  calc Pr_{ let b ← Db; let a ← Da }[ Q a b ]
+      = ∑' b, Db b * (D_rest b) True := prob_tsum_form_split_first Db D_rest
+    _ ≤ ∑' b, Db b * c := by
+        apply ENNReal.tsum_le_tsum
+        intro b
+        exact mul_le_mul_of_nonneg_left (h b) zero_le'
```
alexanderlhicks, line 543: linter says use zero_le

ArkLib/Data/CodingTheory/ProximityGenerator/AffineGenerator.lean
```diff
@@ -0,0 +1,319 @@
...
+
+
+variable {ι : Type} [Fintype ι]
+         {F : Type} [Field F] [Fintype F]
+
+/-- Lemma 7.1. [BCGM25].
+The affine line generator `F → F²`, `x ↦ (1, x)`, having MCA error `ε_mca` for `LC` implies that
+the affine space generator `Fˡ → Fˡ⁺¹`, `x ↦ (1, x)`, has MCA for `LC` with error
+`(1 - 1/|F|)⁻¹ • ε_mca`. -/
+theorem AffineLine_MCA_AffineSpaceMCA {ℓ : ℕ} (hℓ : ℓ ≥ 2) (ε_mca : I → ℝ) (LC : LinearCode ι F)
```
alexanderlhicks, line 265: worth merging in main to get rid of ε_mca : I → ℝ now that #596 is merged

alexanderlhicks:
Following up on the earlier review with four things that have changed underneath this PR since
2026-08-01, and two that came out of reviewing #692. Two are mechanical, one supersedes finding C7,
and one is a re-prioritisation rather than a defect: `isMCAGenerator_of_isMDSGenerator` turns out to
be the only declaration in the whole tree capable of producing a non-trivial `IsMCAGenerator`, which
makes it considerably more load-bearing than "one `sorry` among several".

Merge-order note: #692 stays on `ProximityGap/` and deliberately does not adopt this PR's
`ProximityGenerator/` rename, so #692 needs to land first and this PR rebase on top. That ordering
also resolves F1 and F5 below for free.

---

### F1 · `ArkLib/Data/CodingTheory/ProximityGenerator/MCAGenerator.lean:35` — **[NEW]** will not build after rebase (#701)

`#701` landed on `main` on 2026-08-12 and moved the reusable probability helpers into
`namespace Probability` (see the new `docs/wiki/probability-conventions.md`). This file uses four of
them unqualified — `Pr_le_Pr_of_implies` (:84, :120), `prob_split_uniform_sampling_of_equiv_prod`
(:196), `Pr_seq_le_of_forall_le` (:198), `prob_split_uniform_sampling_of_prod` (:203) — and has no
`open Probability`.

Worth flagging explicitly because **`git merge-tree` reports this merge as clean** — only an actual
build catches it. #692 hit exactly this and it was a one-line fix:

```lean
open NNReal ENNReal unitInterval LinearCode CoreDefinitions Matrix
open scoped ProbabilityTheory
open Probability          -- add
```

I verified on #692 that adding this single line took the merged tree from 10 errors to a clean
`lake build ArkLib` and a green `./scripts/validate.sh`. Please rebase onto current `main` and
re-run the build rather than trusting the mergeability indicator.

---

### F2 · `ProximityGenerator/MCAGenerator.lean:218` — **[UPDATE to C7]** this `sorry` is not known to be provable as stated

C7 said this `sorry` is "stronger than the paper". That was right about the direction but I can now
be precise, because #692 works the same lemma out in full.

As stated here, both hypotheses are MCA **for `LC` itself** (the hypothesis as printed in BCGM25
Lemma 4.4) and the conclusion is at the tight error `ε_mca + ε_mca'`. Reading BCGM25's proof closely:
its Equation (5) step is discharged by the sentence *"which is bounded above by ε′_MCA(γ), the MCA
error of G′"* — and that does not follow from MCA for `C`. Matching the event to `G'`'s MCA experiment
requires the family `w_j := u_(i₀,j)`, where `i₀` comes from an existential that depends on `T`, hence
on the outer seed `x′`. So the family is not fixed and the paper's argument does not reach its own
printed statement.

Two things *are* provable, and both are proved and axiom-clean in #692
(`ProximityGap/TensorGenerator.lean`):

- `tensor_isMCAGenerator` — tight `ε + ε′`, under the hypothesis that `G′` has MCA for the **ℓ-fold
  interleaving** `MC^⋈ℓ`, which is what the paper's proof actually uses;
- `tensor_isMCAGenerator_of_base` — the printed same-code hypothesis, at `ε + ℓ·ε′`; the union bound
  over rows is forced.

Plus `isMCAGenerator_of_moduleInterleavedCode`, showing the interleaved hypothesis is genuinely a
strengthening rather than a restatement.

To be careful about what is *not* being claimed: nobody has shown the statement as written here is
**false**. #692 explicitly withdraws that claim — no separation at equal error is exhibited anywhere.
What is established is that the paper's argument doesn't get there, so this `sorry` should not be
treated as a routine fill.

Suggestion: after rebasing on #692, delete this lemma and re-export its two theorems. That also
removes the last `sorry` from this file.

Worth knowing for the statement itself: expressing the tight form **requires** module alphabets. The
interleaving `C^ℓ` has alphabet `Σ^ℓ`, not `F`, so a `LinearCode ι F`-pinned `IsMCA` cannot state the
hypothesis at all. #692 generalises `IsMCA`/`IsMCAGenerator` to `ModuleCode ι F A` for exactly this
reason.

---

### F3 · `ProximityGenerator/MCAGenerator.lean:245` — **[NEW]** this is the *only* base case in the tree

Not a defect — a re-prioritisation, and I think it's useful context for how to sequence this PR.

While reviewing #692 I checked every declaration in ArkLib that concludes `IsMCAGenerator`. There are
six, and **all six also take an `IsMCAGenerator` hypothesis**: `pseudoinverseGen`, `generatorSubset`,
`AffineLine_MCA_AffineSpaceMCA`, `tensor_isMCAGenerator`, `tensor_isMCAGenerator_of_base`,
`isMCAGenerator_of_moduleInterleavedCode`. Every one of them *transports* MCA; none *establishes* it.

`isMCAGenerator_of_isMDSGenerator` — BCGM25 Theorem 6.1 — is the single declaration anywhere that
would produce MCA from non-MCA hypotheses (`IsMDSGenerator` + dimension + η). While it is `sorry`, the
only `IsMCAGenerator` instances derivable in ArkLib are the trivial ones (`ε ≡ 1` discharges the
hypothesis, since a probability is always `≤ ENNReal.ofReal 1`), and at `ε ≡ 1` every conclusion is
trivial too. Nothing in the tree concludes that any generator has MCA at a non-trivial error.

So this `sorry` is doing much more work than its neighbours, and I'd suggest treating it as its own
piece rather than one line in a large PR. BCGM25 splits it naturally: §5's two counting lemmas, then
§6.1 (unique decoding) and §6.2 (list decoding) as separate regimes. That also lets §8/§9 land
independently of it.

---

### F4 · `ProximityGenerator/Basic.lean:106` — **[UPDATE]** `ε_mca : I → ℝ` now diverges from `main`, and admits negative errors

At the time of the earlier review `main` had `I → I`, so this was one of three incompatible types.
`main` has since settled on **`I → ℝ≥0`**, and #692 keeps it.

Beyond consistency, `ℝ` admits a hazard `ℝ≥0` doesn't: for negative `ε_mca γ`, `ENNReal.ofReal` maps
it to `0`, so `IsMCAGenerator G ε_mca LC` silently becomes the assertion `Pr[...] = 0` rather than
being rejected. Nothing here produces a negative error today, but `ε_MCA_MDS` (:225) is an arithmetic
expression in `η`, `δ_C` and `n` with no non-negativity side condition, so it isn't structurally
prevented either.

Suggest moving to `I → ℝ≥0` on rebase. (BCGM25 Def 3.14 types the error `[0,1] → [0,1]`; `main` uses
`ℝ≥0` rather than `I` so that sums like `ε + ℓ·ε′` are expressible — #692's `TensorGenerator.lean`
docstring notes the resulting bound is vacuous once `ℓ·ε′ ≥ 1`.)

One refinement since I wrote this: #692 has now also dropped the `ENNReal.ofReal` wrapper in favour
of a plain coercion `Pr[…] ≤ ↑(ε_mca γ)`. For `p : ℝ≥0` the two are equal, but `ofReal` is a
round-trip through `ℝ` that forces hand-rolled `ofReal_add` / `ofReal_mul` / `ofReal_natCast`
rewrites at every closing step, where the coercion lets `norm_cast` do it. Worth taking together
with the type change rather than separately — and note that the `ofReal`-swallows-negatives hazard
above is precisely what disappears when the bound is `ℝ≥0`.

---

### F5 · `ArkLib/Data/CodingTheory/Basic/LinearCode.lean:270` — **[UPDATE to C10]** superseded on `main`

C10 asked for `Submodule.map` instead of the hand-rolled `projectedCode_submod`. #692 has now done
this, and generalised it in the same step:

```lean
def projectedCodeSubmod [Semiring F] [Module F A] (MC : ModuleCode ι F A) (T : Finset ι) :
    Submodule F (T → A) := MC.map (LinearMap.funLeft F A (Subtype.val : T → ι))
```

Note `[Field F]` is weakened to `[Semiring F]` and the alphabet is now general. So C10 is best resolved
by dropping this definition on rebase and taking `main`'s, rather than by fixing it in place — which
also removes the naming divergence (`projectedCode_submod` here vs `projectedCodeSubmod` on `main`).

---

### F6 · `ProximityGenerator/PolynomialGenerator.lean:168` — **[NEW]** a gap in BCGM25 itself, upstream of C5/C6

Surfaced while checking #692's docstrings, and it may bear on why Lemma 9.3 and Theorem 9.2 are hard
to state correctly here.

BCGM25's proof of Theorem 9.2 writes *"By Lemma 9.3, `G_d` has mutual correlated agreement for any
linear code `C` with error `ε_MCA,RS,d`"* — but Lemma 9.3 is stated **only** for `RS[F, D, k]`, and its
proof is irreducibly Reed–Solomon-specific (it builds the Guruswami–Sudan `Q(X,Y,Z)` of
[BCIKS20, Thm 5.1] and factors `disc*_Y(Q)`). The parallel sentence in the proof of Theorem 8.2 *is*
justified — by Theorem 6.1, which holds for every F-linear code with error depending on `C` only
through `n` and `δ_C` — but the Theorem 9.2 one is not justified by the lemma it cites.

So Theorem 9.2's printed error does not follow from its printed proof, independently of anything in
this PR. That's worth recording in the module docstring, and is probably worth raising with the
authors alongside the Lemma 4.4 point in F2.

Caveat on how far I checked this: I verified the statement scopes and the section structure, not
Lemma 9.3's proof line by line. Flagging it as something to confirm before relying on it, not as a
settled finding.

---

---

## Provenance

F1, F3, F6 and the `main`-side half of F4/F5 come from a review of #692 carried out 2026-08-12
(build, `#print axioms`, `validate.sh` and a merged-tree build all run first-hand; write-up in
`PR692_REVIEW_2026-08-12/`). F2's analysis of BCGM25 Lemma 4.4 was re-derived from the source PDF and
cross-checked against #692's proofs, which are axiom-clean.

The earlier C1–C18 findings — including the compiled refutation of Lemma 9.3 as stated
(`Probe_610_Lemma93_refutation.lean`) — were **not** re-verified on 2026-08-13. They were checked
against `e3e8d1e1`, which is still this PR's head, so they should still apply, but say so if relaying
them as current.

ArkLib/Data/CodingTheory/ProximityGenerator/MCAGenerator.lean
katyhr: I think it depends if we manage to prove the tight version without having the hypothesis on interleaved codes as the paper and Ignacio claims to be possible. If we have a hypothesis on the interleaved codes, it might be good to have both versions in case there are instances where the interleaved code assumption fails

ArkLib/Data/CodingTheory/ProximityGenerator/MCAGenerator.lean
katyhr: This is the impression I got too even though this is not what the paper says. So I left it as sorry for now to try to prove it without this hypothesis

ArkLib/Data/CodingTheory/ProximityGenerator/PolynomialGenerator.lean
katyhr: oh yes! thanks!

ArkLib/Data/CodingTheory/ProximityGenerator/PolynomialGenerator.lean
katyhr: Changed

alexanderlhicks:
closing due to #766.

PR #617, opened 2026-07-09
Slight fix to the statement of grandMCAChallenge+ proof of grandMCA…

katyhr:
…Challenge, serving as counter-example to correctness of formalisation.

Slight fix to `grandMCAChallenge`, to account for the case, where the maxima is not attainable, say that the set of values `δ`, for which `epsMCA (F := F) (A := F) ((C : Set (ι → F))) δ > (ε_star : ENNReal)` hold, are the values `< 3/4`, then no maximal exists. 

Also, including proof of `grandMCAChallenge`, demonstrating the importance of the constructivity of `δ_C`.


alexanderlhicks:
Thanks for flagging!

PR #618, opened 2026-07-09
ProximityChallenges

katyhr:
formalisation of challenges in [ABF26]. Note: restriction of `δ_C` and `δ_C'` to a rational number to be able to enforce constructivity.


alexanderlhicks:
Thanks for flagging!

PR #621, opened 2026-07-09
refactor: move constRq and Rq degree/coef lemmas into Rq.lean

ErVinuelas:
Move some lemmas in `Gadget.lean` to `Rq.lean` since it seems that they belong there. I also got rid of a small lemma (`constRq_one`) that it is not used and does not seem very relevant.


alexanderlhicks (approved):

alexanderlhicks:
Thank yoU!

PR #631, opened 2026-07-10
feat(SendWitness): prove reduction completeness; add Prover.runToRound_succ

Abraxas1010:
## Summary

Fills the open `sorry` in `SendWitness.reduction_completeness` — per the coordination note on #627 (no claim was raised, so bringing the proof as offered) — and adds the general step lemma its investigation surfaced:

* **`Prover.runToRound_succ`** (`Execution.lean`, beside `runToRound_zero_of_prover_first`): `prover.runToRound i.succ = prover.processRound i (prover.runToRound i.castSucc)` — a one-line term via `Fin.induction_succ`. This is the round-by-round unfolding brick for collapsing concrete protocol executions; its absence is why component completeness proofs were stuck.
* **`SendWitness.reduction_completeness`** — proven. The interesting discovery: for this component the entire `Reduction.run` collapses **definitionally** — `have hrun : (reduction …).run stmtIn witIn = pure ((Transcript.concat witIn default, (stmtIn, witIn), ()), (stmtIn, witIn)) := rfl` — the kernel reduces the monad lifts, the direction match, the `Fin.induction`, and the transcript concat in one step (the key is ascribing `default` at the literal index `0`, where the `Transcript` instance synthesizes). The probabilistic legs then follow the `Reduction.id_perfectCompleteness` template exactly (`one_le_probEvent_iff` → `probEvent_eq_one_iff` → the `OptionT`/`simulateQ`/`StateT` change-chains).

`#print axioms`: `reduction_completeness` = `[propext, Classical.choice, Quot.sound]`; `runToRound_succ` = `[propext, Quot.sound]`.

## Notes

* The oracle-reduction variants further down the file (their own `sorry`s) are untouched — this PR is deliberately the smallest complete unit; the same `rfl`-collapse + template recipe should apply there and we're happy to follow up.
* This is the completeness demonstrator for the BCS transform plan in #627: our phase-2 (reveal round) completeness is exactly SendWitness-shaped.
* Full `lake build` of both touched modules green; no new axioms; no `native_decide`/`bv_decide`.

---

*Contributed by The Institute for Ontological Mathematics (IAOM) / Equation Capital dba Apoth3osis.*

PR #629, opened 2026-07-10
Refactor/small clean up

ErVinuelas:
Very small PR. Got rid of an unused variable and moved a lemma where I think it belongs.

PR #626, opened 2026-07-10
feat/refactor[Hachi]: fig. 3 protocol + CWSSPackages abstraction + partial reorg of Hachi folder

tobias-rothmann:
## TLDR

This PR adds the (sub)protocol of fig. 3 from the Hachi paper and along the way, introduces a new abstraction called `CWSSPackage` that allows for easy composition of CWSS protocols (even has nice notation). It also slightly restructures the Hachi folder to be more readable (typically with subfolders for subprotocols and more documentation of where what is to be found). 

## Figure 3 Protocol

This PR adds the `Hachi/QuadEval/` development for the Figure 3 polynomial-evaluation reduction.

The reduction rewrites a multilinear evaluation claim `f(x) = y` as the quadratic form `bᵀ M a`, then folds the `2^r` carrier blocks under the verifier’s challenge vector. In Lean, the protocol is split into:

- `QuadEval/Gadgets.lean`: gadget algebra for `PublicParamsD`, the carrier `w`, its decomposition `ŵ`, the short commitment `v = D ŵ`, the `J` decomposition, and the `tensorG` / `tensorG1` identities used by the extractor.
- `QuadEval/Reduction.lean`: the Figure 3 protocol data, including `QuadEvalStatement`, `QuadEvalResponse`, `QuadEvalWitness`, `ShortChallenge`, `relIn`, `relOut`, `verifier`, and the honest `prover` skeleton.
- `QuadEval/Soundness.lean`: Hachi Lemma 8 as `quadEval_coordinateWiseSpecialSound`, together with `buildWitness` and the composable `quadEvalPackage`.
- `QuadEval/Bridge.lean`: the zero-round polynomial-level bridge from `PolyEvalStatement` to `QuadEvalStatement`, with `relPolyEval`, `bridge_coordinateWiseSpecialSound`, and `bridgePackage`.

The implementation is organized as a two-package CWSS chain. The first package is a zero-round
adapter into the Figure 3 statement shape; the second package is the actual two-round Figure 3
protocol. At the relation level, the chain looks like this:

```mermaid
flowchart LR
  R0(["<b>relPolyEval</b><br/>on PolyEvalStatement"])
  R1(["<b>relIn</b><br/>on QuadEvalStatement"])
  R2(["<b>relOut</b><br/>Eq. (20) + range checks"])

  R0 -- "bridgePackage<br/><i>0-round adapter<br/>toQuadEvalStatement</i>" --> R1
[...]

PR #632, opened 2026-07-10
fix(ci): Build Timing Report cannot post its PR comment (pull-requests: write)

Abraxas1010:
Every recent **Build Timing Report** run fails at the final *Upsert build timing PR comment* step with `Resource not accessible by integration` (observed on PRs #602, #626, #629, #631 — see the workflow history on `main`).

**Root cause:** the report posts to the triggering PR via `issues.createComment`. For comments on *pull requests*, that endpoint needs `pull-requests: write`; `issues: write` only covers plain issues. The workflow currently grants `issues: write` + `pull-requests: read`, so the token is rejected at the last step.

**Fix:** one line — `pull-requests: read` → `pull-requests: write`. No other behavior change; the job already runs in the base-repo context (`workflow_run`), so no fork-token hazard is introduced.

We noticed this because our PR #631's timing report failed; the failure predates and is independent of that PR.

— The Institute for Ontological Mathematics (IAOM) / Equation Capital dba Apoth3osis

### DIFF
--- a/.github/workflows/build-timing-report.yml
+++ b/.github/workflows/build-timing-report.yml
@@ -13,7 +13,7 @@ permissions:
   actions: read
   contents: read
   issues: write
-  pull-requests: read
+  pull-requests: write
 
 jobs:
   report:


quangvdao:
Thanks for tracking this down. We do want to retain the build-timing PR comment, but I’m going to consolidate the permission fix with the resilience work in #592 rather than land two competing patches. The revised approach should use `pull-requests: write`, remove the redundant `issues: write`, and keep expected comment-posting failures non-blocking while leaving unexpected script errors visible. Closing this PR in favor of that combined change.

PR #615, opened 2026-07-07
feat(ring-switching): add reusable packing proofs and protocol integrations

alexanderlhicks:
This adds reusable ring-switching algebra and protocol proofs, used by FRI-Binius tensor batching and Hachi's trace head. Generic scalar and public-family reductions compose checked reconstruction, batching and product sumcheck to an opening of the same committed polynomial.

- Independent finite bases support different packing/opening algebras and ranks. Finite weighted observation is shared by Boolean-table packing and Hachi's monomial/ψ representation. Checked observation supplies honest checking and exact witness readback while preserving each protocol's commitment predicate.
- Binius retains its tensor message and vector challenge. Its packing/layout, guard and round-zero relation consume the shared proofs. The batching head has proved completeness and worst-case knowledge security, instantiated with the production codeword commitment's unique-distance binding.
- Hachi's trace-head checking, readback, completeness and coordinate-wise special soundness consume the shared reconstruction lemmas while retaining the scaled trace, weak opening and norm conditions.
- Generic pipelines use guarded composition with explicit extractors, knowledge states and per-prefix bounds. Challenge errors sum to `batching error + m * (2/|C|)`. Randomized security requires functionality, suitable challenge domains and injective coefficient transport; honest coverage is separate. `PackedOpening` appends a supplied downstream reduction at the same `pc.evalRel`.
- Tensor coordinates have two-sided inverse laws. Failed verifier checks are absorbing, and terminal checks forward the received packed value, including when the multiplier is zero. The FRI-Binius terminal preserves its accepted output fields and forwarded oracles.

The APIs use semantic layout names (`packedPrefixLayout`, `packedSuffixLayout`, `quirkyLayout`) and `ProfileLayout` for the Boolean-table correspondence. Declaration docstrings describe their Lean declarations; module docs carry protocol context and references. The [coverage reference](https://github.com/Verified-zkEVM/ArkLib/blob/feat/generic-ring-switch/docs/kb/audits/ring-switching-model-coverage.md), concept/paper pages, wiki and blueprint describe current interfaces, source correspondence and proof boundaries.

[...]

PR #654, opened 2026-07-16
feat/refactor(Hachi): ring switching subprotocol

tobias-rothmann:
Ring switching for hachi as a proper subprotocol :)


ArkLib/ProofSystem/RingSwitching/Basic.lean
```diff
@@ -0,0 +1,102 @@
...
+
+/-!
+# Ring Switching — a family of constructions, not one protocol
+
+Umbrella for `ProofSystem/RingSwitching/`. "Ring switching" names several reductions that
+move an evaluation or linear claim between a small ring and a large ring — so that each part
+of a proof system can run where it is cheap or sound: commit over the small ring, evaluate
+and open over a large extension; or state a relation over a structured quotient ring, check
+it inside a field. The constructions share *algebra*, not *protocol*; no single generic
+ring-switching reduction exists in this library, and none is planned. The two construction
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[cross-PR coordination]** This umbrella declares that no generic ring-switching reduction exists or is planned, and reframes the DP24 packing machinery as `Packing/`. But PR #615 (`feat/generic-ring-switch`) builds exactly such a generic reduction in `RingSwitching/Generic/` (S1–S6 + an RBR-KS bound). The two framings are directly contradictory. There's no file-level conflict (`Generic/` vs `Packing/`+`Lift/` are disjoint), but if both land the repo will hold two generic packing ring-switches and a doc denying one exists. Please coordinate with #615's owner on which framing wins before either merges, and update this sentence accordingly. (Both PRs also edit `docs/kb/concepts/ring-switching.md`, `repo-map.md`, `references.bib`, `FRIBinius/Prelude.lean`, `Multilinear.lean` — expect textual conflicts there.)

ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/ScalarRound.lean
```diff
@@ -64,29 +72,276 @@ instance : ∀ i, SampleableType ((pSpecScalar Msg C).Challenge i)
...
-`mem_of_pure_accepting`. -/
+  {σ : Type}
+
+/-- **Generic scalar-round CWSS assembly.** Any pure statement-extending verifier of the
+two-round scalar `pSpecScalar` is coordinate-wise special sound for `scalarStructure k`,
+provided a witness assembler `mkWitness` that turns `k` per-branch `relOut`-witnesses at
+pairwise-distinct challenges into a `relIn`-witness. This is the engine behind Hachi Lemma 9
+(`k = 2d`, interpolation) and Lemma 11 (`k = deg + 1`, per sumcheck round): all tree/extractor
+plumbing is discharged here once; the protocol-specific work lives entirely in `hmk`. -/
 theorem coordinateWiseSpecialSound_of_mkWitness_scalar
```
alexanderlhicks, line 345: 🤖 _AI-generated (Claude Code)_

**[praise · merge-order]** Nice — this closes the `sorry` still open on `main` and in #659 (`ScalarRound` grows 106→380 lines; I verified the theorem is axiom-clean via `#print axioms`). Because #659's guarded/escape work sits on top of this engine, landing #654 first and having #659 rebase onto it (rather than re-diverging) is the natural merge order. Flagging so the ScalarRound engine isn't duplicated across the two PRs.

ArkLib/Commitments/Functional/Hachi/RingSwitch/Reduction.lean
```diff
@@ -63,183 +54,145 @@ import ArkLib.OracleReduction.Security.CoordinateWiseSpecialSoundness.ScalarRoun
...
+`short_zOk` hypothesis; the interpolation/descent recovery and the escape/collision extractor
+are supplied by the generic layer. -/
+theorem vecLInftyNorm_le_of_liftShort (s : RlinStatement Φ n μ) (w : LiftedWitness Φ μ n)
+    (hshort : liftShort Φ bound ρBound w) (hside : bound ≤ s.bound) :
+    vecLInftyNorm Φ w.z ≤ s.bound :=
+  le_trans hshort.1 hside
+
+/-! ### Why the CWSS certificate is exposed only as `liftPackage.isCWSS`
+
+We deliberately do **not** provide a standalone `lift_coordinateWiseSpecialSound` restated in
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[fragility guard]** The documented reason for not restating a standalone `lift_coordinateWiseSpecialSound` (whnf heartbeat timeout through `Rq`'s computable layer) is reasonable, and I confirmed `liftPackage` elaborates cleanly with no heartbeat error. Since this is a known fragility, consider a small CI guard / regression test that the `Composition.lean` `▷`-seams keep elaborating, so a future refactor perturbing the `rfl` seams fails loudly rather than silently timing out.

ArkLib/Data/Lattices/CyclotomicRing/QuotientLift.lean
```diff
@@ -0,0 +1,89 @@
...
+variable {R : Type*} [Field R]
+
+variable [BEq R] [LawfulBEq R] (Φ : CyclotomicModulus R) [IsCyclotomic Φ]
+
+/-! ## Degree bound and injectivity of canonical representatives -/
+
+/-- A reduced representative has `natDegree` below `d = deg φ` — the any-modulus version of
+`Rq.natDegree_val_toPoly_lt` (which is pinned to `powTwoCyclotomic`), with the dimension
+positivity `0 < d` as an explicit hypothesis. -/
+theorem Rq.natDegree_val_toPoly_lt' (hd : 0 < Φ.φ.natDegree) (a : Rq Φ) :
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[dedup · non-blocking]** This generalizes main's `powTwoCyclotomic`-pinned `natDegree_val_toPoly_lt` to any modulus with `0 < deg φ` — a genuine generalization, not a duplicate. But the two now coexist; consider re-deriving the pinned version from this general one (or noting the relationship) so they don't drift. Cleanliness only.

alexanderlhicks (commented):

ArkLib/ProofSystem/RingSwitching/Basic.lean
tobias-rothmann: This is mostly what we talked about, I think. I don't think there is a sensible way to unify the structurally so different ring switching techniques into one. Both reuse some parts of the "Transport" and verifier rounds files where possible, but fundamentally having Lift vs Packing makes sense imho.

ArkLib/ProofSystem/RingSwitching/Lift/Reduction.lean
```diff
@@ -0,0 +1,221 @@
...
+    CommittedScalar.buildWitness_mem
+      (by have := hd ▸ P.natDegree_modulus_pos; omega) K (fun w => w.z)
+      (checkAt P φF getM getY sideCond) (relLin getM getY zOk)
+      (fun s' w fam' hinj' hcheck hshort =>
+        recover P φF getM getY zOk sideCond hφF hd short_zOk s' w fam' hinj' hcheck hshort)
+      s t fam resp hresp hinj
+
+/-- **CWSS of `Lift`**, at plain `k = 2d` special soundness, from the
+generic committed-scalar theorem. -/
+theorem coordinateWiseSpecialSound [IsPresentation P] (hφF : Function.Injective φF)
```
alexanderlhicks, line 187: **This certificate has no security content once `K.esc` is inhabited.**

Both `relIn` and `relOut` here are `withEscape K.esc`, and the following compiles against this
branch — with no hypothesis about the verifier, the challenge structure, the tree, or acceptance
(Lean even flags the `SampleableType` instance as unused):

```lean
theorem cwss_trivial_of_escIn_nonempty
    (D : CWSSStructure pSpec) (relIn : Set (StmtIn × WitIn)) (relOut : Set (StmtOut × WitOut))
    (esc : Set E) (escOut : Set EOut) (he : esc.Nonempty)
    (verifier : Verifier oSpec StmtIn StmtOut pSpec) :
    verifier.coordinateWiseSpecialSound init impl D
      (relIn.withEscape esc) (relOut.withEscape escOut) := by
  obtain ⟨e, he⟩ := he
  exact ⟨fun _ _ => .inr e, fun _ _ _ _ => he⟩
```

`Set.mem_withEscape_language_iff` degenerates the other half: with `esc` nonempty every statement
lies in `(relOut.withEscape esc).language`, so `ChallengeTree.IsAccepting` — the only place the
verifier enters `treeSpecialSound` — carries no information either.

`Escape.lean:51-55` concedes this is "the formal price", and the usual escape hatch is "for
kernel-free keys the set is empty". That hatch is unavailable in the regime the scheme needs: for a
**compressing** key, `z ↦ A *ᵥ z` maps the box `{‖z‖∞ ≤ γ}` (size `(2γ+1)^(d·cols)`) into
`Rq^rows` (size `q^(d·rows)`), which is non-injective once `(2γ+1)^cols > q^rows`; the difference
of two colliding short vectors is a nonzero kernel element of `ℓ∞`-norm `≤ 2γ`, i.e. exactly an
escape. Module-SIS is only *computationally* hard, and `withEscape` asks only for set membership.

**The good news is that the real content is already here and is not affected.**
`buildWitness_mem` (line 168) constrains the *concrete* extractor and is genuinely non-trivial on
the all-`inl` branch; so do `recover`, `Presentation.mulVec_eq_of_evalAt_rowSum` and
`eq_of_evalAt_eq`. The vacuity is entirely in the last step, where that concrete extractor is
discarded for `∃ E : Extractor.TreeBased, …`.

Suggested fix, cheap and losing no mathematics: add an extractor-pinned form
(`treeSpecialSoundWith (E := buildWitness …)`), advertise *that* as the security theorem, and keep
`coordinateWiseSpecialSound` as the composition interface, labelled as such. Alternatively require
`esc = ∅` at top level and surface the MSIS branch as "*this named algorithm outputs an element of
the escape set*" rather than "an element exists".

This is pre-existing framework behaviour rather than something #654 introduces, so it needn't block
the merge — but it is far cheaper to fix now than after rows 7-12 are built on it, and the KB claim
at `docs/kb/papers/NOZ26.md:43-45` should be scoped either way. Same issue applies to #659's
`QuadEvalEscapeMap`, where I've left the matching comment.

alexanderlhicks (commented):
*Part of a cross-PR review of #654, #656 and #659 read together, in isolated builds, against NOZ26 (ePrint 2026/156, SHA-256 `c7c98591…afd5e6a15`) and the current HMZ25 revision. Where a finding interacts with a sibling PR I say so explicitly.*

## #654 — reviewed at `215b8868`

**This is the strongest of the three Hachi PRs and I think it should merge first.**

The quotient-presentation algebra is correct. I re-derived the degree accounting independently:
`rowSum` has `natDegree ≤ 2d−2`, `rep (y i) + modulus * ρ` has `natDegree ≤ 2d−1`, both `< 2d`, so
the `2d` constant is **tight** against Lemma 9's "degree at most `2d − 1`". Both directions of the
quotient-witness correspondence are proved, the exactness layer is a real strengthening of the
coset laws (not a restatement), and stating the extraction target over `Rq Φ` rather than the
paper's `Z_q[X]` is equivalent by exactly those two lemmas — the cleaner choice.

The whole chain is **axiom-clean including `liftPackage` itself**: `#print axioms` on 12
declarations (`Presentation.mulVec_eq_of_evalAt_rowSum`, `exists_rowSum_eq_of_mulVec_eq`,
`eq_of_evalAt_eq`, `Lift.recover`, `Lift.buildWitness_mem`, `Lift.coordinateWiseSpecialSound`,
`Lift.package`, `CommittedScalar.buildWitness_mem`, `CommittedScalar.coordinateWiseSpecialSound`,
`isPresentation_cyclotomic`, `vecLInftyNorm_le_of_liftShort`, `liftPackage`) returns exactly
`[propext, Classical.choice, Quot.sound]` in every case. Reuse discipline is good too: `evalAt`
bundles Mathlib's `eval₂RingHom`, the interpolation kernel routes through
`eq_zero_of_natDegree_lt_card_of_eval_eq_zero`, nothing is forked, and every changed file is inside
the 1500-line cap.

I found **no false theorem and no unsound bound.** The comments below are claim/scope/statement
defects, and all are cheap.

Highest-value items:

1. The escape-threaded CWSS certificate has no security content once `K.esc` is inhabited — and
   Hachi's escape is a Module-SIS solution set, which is inhabited for any *compressing* key
   (comment on `Lift/Reduction.lean:187`, with a compiled witness). Pre-existing on `main`, so not
   a blocker for this PR, but the KB should stop describing it as proven soundness.
2. `docs/kb/papers/NOZ26.md:45` and `HMZ25.md:43` advertise `lift_coordinateWiseSpecialSound`,
   which **does not exist anywhere in the tree**, and `NOZ26.md:79` contradicts `NOZ26.md:43-45`.
3. `references.bib:496` has the wrong HMZ first author. #659 already has the correct entry — take
   theirs verbatim (this also resolves the add/add conflict).
4. One avoidable new `sorry` with **zero consumers**; a complete 18-line proof is in the comment.
5. `docs/skills/prove-milestone.md` relaxes a repo-wide axiom policy out of scope, and states it
   incorrectly ("exactly" should be "subset" — the "exactly" form would fail several lemmas in
   this very PR).

**Scope, for the body/title:** what is formalized is the *simplified Figure 4 / Lemma 9 extraction
kernel*, not the p. 18 protocol that commits `(z, r₁, …, r_log_b(q))` with per-digit norm bounds —
the paper's own words are "*there is a hidden gadget decomposition of `r`*". There is also no
honest-prover completeness theorem in the Lift layer. `RingSwitch/Reduction.lean:41-46` states this
correctly; the KB pages should match it.

**On the open ring-switching PR #615** (head `68e9a752`): its `Generic/Carrier.lean:7` imports
`RingSwitching.Prelude`, which this PR renames, so #615 will not build after this merges — one-line
import fix on their side. Its taxonomy interaction is in the `Basic.lean:19` comment. Worth noting
that #615 is *not* exposed to item 1 above: it states security as an RBR knowledge-soundness
probability bound and never uses `withEscape`, so there is no constant-extractor escape hatch there.
The fix I suggest for the Hachi certificates moves them toward the shape #615 already has.

**Ordering: #654 → #659 → #656.** Two concrete reasons beyond the architecture: this PR proves
`ScalarRound.coordinateWiseSpecialSound_of_mkWitness_scalar`, which #659 still carries as `sorry`;
and it deletes the six root `HACHI_*.md` notes while cleaning up **every** reference to them (0
dangling), where #659 does the same deletion and leaves 21 dangling across 10 files — so #659
should adopt this PR's cleanup rather than redo it.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
```diff
@@ -72,64 +73,112 @@ variable (m₀ m₁ : ℕ)
...
+vanishing polynomial of the symmetric range `{−(b−1), …, b−1}`. Its multilinear extension over
+the table entries is `H₀`. -/
+def rangeProduct (b : ℕ) (v : F) : F :=
+  v * ∏ j ∈ Finset.Icc 1 (b - 1), ((v - (j : F)) * (v + (j : F)))
+
+/-- Over a field, `P_b(v) = 0` iff `v` is the image of an integer in the symmetric range
+`{−(b−1), …, b−1}`. The engine of the `H₀ ⇒ liftShort` bridge (the corrected [NOZ26, Lemma 10]
+reads the field roots back as centered representatives under `2b − 1 < q`). **Sorried**
+(skeleton). -/
+theorem rangeProduct_eq_zero_iff {b : ℕ} {v : F} :
```
alexanderlhicks, line 103: **This `sorry` is avoidable — here is the proof.**

Three things I checked: the statement is **true**, it has **zero Lean consumers** at this head (the
only other mention is prose at `ZeroCheck/Batch.lean:79`), and it takes 18 lines. So the trust gap
buys nothing. Verified axiom-clean against this exact definition of `rangeProduct`:

```lean
theorem rangeProduct_eq_zero_iff {b : ℕ} {v : F} :
    rangeProduct b v = 0 ↔ ∃ j : ℕ, j ≤ b - 1 ∧ (v = (j : F) ∨ v = -(j : F)) := by
  classical
  rw [rangeProduct, mul_eq_zero, Finset.prod_eq_zero_iff]
  constructor
  · rintro (hv | ⟨j, hjmem, hfac⟩)
    · exact ⟨0, Nat.zero_le _, Or.inl (by simpa using hv)⟩
    · have hjle : j ≤ b - 1 := (Finset.mem_Icc.mp hjmem).2
      rcases mul_eq_zero.mp hfac with h | h
      · exact ⟨j, hjle, Or.inl (sub_eq_zero.mp h)⟩
      · exact ⟨j, hjle, Or.inr (by linear_combination h)⟩
  · rintro ⟨j, hjle, hv | hv⟩
    · rcases Nat.eq_zero_or_pos j with rfl | hjpos
      · exact Or.inl (by simpa using hv)
      · exact Or.inr ⟨j, Finset.mem_Icc.mpr ⟨hjpos, hjle⟩, by rw [hv]; simp⟩
    · rcases Nat.eq_zero_or_pos j with rfl | hjpos
      · exact Or.inl (by simpa using hv)
      · exact Or.inr ⟨j, Finset.mem_Icc.mpr ⟨hjpos, hjle⟩, by rw [hv]; ring_nf⟩
```

(#656 also proves this one, axiom-clean — but pasting the above keeps #654 self-contained and
avoids a cross-PR dependency in the direction opposite to the merge order.)

One scope note for the docstring while you're here: this yields `v = ±(j : F)`, **not** a centered
`ZMod q` representative. Getting from here to a norm bound additionally needs injectivity of
`ℕ → F` on `[0, b−1]`, i.e. the `2b − 1 < q` condition the docstring alludes to — in #656 that step
is `valMinAbs_natAbs_le_of_rangeProduct_eq_zero`.

ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/CommittedScalar.lean
```diff
@@ -0,0 +1,270 @@
...
+Two distinct short openings of one commitment produce an element of `esc`.  This is the common
+interface needed by committed scalar phases: exact binding uses an empty escape type, while
+Hachi's weak binding instantiates `E` with its Module-SIS escape budget. -/
+structure BindingCommitment (W E : Type) (Short : W → Prop) where
+  /-- Commitment/message type. -/
+  TCom : Type
+  /-- Deterministic commitment map. -/
+  com : W → TCom
+  /-- Valid binding-break artifacts. -/
+  esc : Set E
```
alexanderlhicks, line 74: `esc` is constrained only from *below*: `collision_mem` (line 78) says a collision lands in `esc`,
and nothing bounds it above. So `esc := Set.univ, escOfCollision := fun _ _ => e` satisfies the
structure — and by the comment on `Lift/Reduction.lean:187`, any nonempty `esc` already empties the
downstream certificate.

Credit where it's due: the `example` at `RingSwitch/Reduction.lean:115-121` (empty escape, injective
commitment) is exactly the right non-vacuity witness, and it is more than either sibling PR
provides. The gap is only that `liftPackage` still accepts an arbitrary `K`, so the example does not
constrain the shipped entry point.

Two options if you want the type to carry the guarantee: an exact-binding variant with
`E := Empty`, or an extra field bounding `esc` above by the image of `escOfCollision` on colliding
short pairs, so `esc` cannot be inflated past what a collision can actually produce.

docs/kb/papers/NOZ26.md
```diff
@@ -37,9 +37,25 @@ Commitment layer:
...
 
-- The **extension-field → cyclotomic-ring reduction**: Hachi reduces evaluation proofs over `F_{q^k}`
-  to equivalent statements over a power-of-two cyclotomic ring `R_q`. This is the ring-switching
-  shape ArkLib factors out as `RingSwitchingProfile`.
+- The **extension-field → cyclotomic-ring reduction** (§3): Hachi reduces evaluation proofs over
+  `F_{q^k}` to equivalent statements over a power-of-two cyclotomic ring `R_q`. This is the
+  ring-switching shape ArkLib factors out as `RingSwitchingProfile`.
+- The **cyclotomic-ring → extension-field lift** (§4.3, Figure 4 / **Lemma 9**, following
+  [`HMZ25`](HMZ25.md)): **formalized and proven** (sorry-free, axiom-clean) as
+  `liftPackage` / `lift_coordinateWiseSpecialSound` in Hachi's
```
alexanderlhicks: Two problems in this sentence.

**1. `lift_coordinateWiseSpecialSound` does not exist.** Grepping the whole tree at this head, the
only two occurrences are this line and `docs/kb/papers/HMZ25.md:43`. The certificate is the
`liftPackage.isCWSS` field; the generic theorem is
`RingSwitching.Lift.coordinateWiseSpecialSound`.

**2. "formalized and proven (sorry-free, axiom-clean)" conflates two different claims.** The
axiom-cleanliness is real and worth stating — I re-ran `#print axioms` across the whole chain
*including `liftPackage` itself* and every declaration reports exactly
`[propext, Classical.choice, Quot.sound]`. But:

- the escape-threaded certificate carries no security content for a nonempty escape set (see the
  comment on `ArkLib/ProofSystem/RingSwitching/Lift/Reduction.lean:187`), so axiom-cleanliness is
  not evidence that the statement says what a reader will assume; and
- what is formalized is the **simplified Figure 4 / Lemma 9 extraction kernel**, not the p. 18
  protocol committing `(z, r₁, …, r_log_b(q))` with per-digit norm bounds — the paper's own words
  are "*there is a hidden gadget decomposition of `r`*". `RingSwitch/Reduction.lean:41-46` states
  this correctly under "Paper-model boundary"; this page should match it.

Worth recording here too: there is no honest-prover completeness theorem anywhere in the Lift
layer, so the paper's "completeness follows directly" is unformalized.

docs/kb/papers/NOZ26.md
```diff
@@ -80,10 +96,12 @@ Ring-switching layer:
...
-- Formalize Hachi-specific soundness separately (does not reuse the field/domain soundness theorem).
+  `decomposeRows_spec` / `decomposeColumns_spec` via Theorem 2, with `2^κ_pack = d/k`
+  (the §3 packing head; the §4.3 HMZ25 lift, Lemma 9, is done — see above).
+- Formalize Hachi-specific soundness separately (does not reuse the field/domain soundness
+  theorem): done through Lemma 9 (rows 1–4 of the opening chain); Lemmas 10–11 and the
+  recursion adapters remain skeletons (see `Hachi/Composition.lean`'s inventory).
 - The norm-growth and short-element invertibility inputs (`Mic07`, `LS18`) are deferred.
-- The sumcheck / ring-switching evaluation machinery of the paper is not yet formalized.
 
 ## Version Notes
```
alexanderlhicks, line 109: Two structural problems in this page, both visible from here (the lines I would rather anchor to,
77-90, fall outside this PR's diff hunks).

**1. Duplicated sections.** `## Version Notes` appears at both line 85 and line 106 (this line), and
`## Known Divergences From ArkLib` at both 77 and 90. Looks like a merge artefact.

**2. The line-79 block contradicts the page's own body.** It says "ArkLib has not yet built the
Hachi ring-switching instance" and that Hachi soundness "is out of scope for the current
ring-switching module" — versus lines 43-45, written by this PR, saying it is "formalized and proven
… in Hachi's `RingSwitch/Reduction.lean`". The line-79 text appears to predate the PR and to
describe the *Packing* / `RingSwitchingProfile` situation, so it needs re-scoping to say so
explicitly rather than "the Hachi ring-switching instance".

docs/kb/papers/HMZ25.md
```diff
@@ -0,0 +1,83 @@
...
+- The quotient-witness correspondence and its interpolation soundness kernel:
+  `ArkLib/Data/Lattices/CyclotomicRing/QuotientLift.lean` (generic, monic cyclotomic modulus,
+  abstract extension embedding `φF : R →+* F`) — proven and axiom-clean.
+- The lift as a two-round subprotocol (commit to the lifted witness, evaluate at a random `α`)
+  with `k = 2d` plain special soundness, **formalized generically** in
+  `ArkLib/ProofSystem/RingSwitching/Lift/` over any monic-modulus presentation
+  `S ≅ R[X]/(φ)` (`Presentation`/`IsPresentation`), on the committed-scalar shell
+  `ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/CommittedScalar.lean`;
+  Hachi's presentation instance, norms, local predicate,
+  interpolation recovery, `liftPackage`, and `lift_coordinateWiseSpecialSound` are in
```
alexanderlhicks: Same nonexistent declaration as `docs/kb/papers/NOZ26.md:45`: `lift_coordinateWiseSpecialSound` is
not defined anywhere in the tree. Use `liftPackage.isCWSS` (the Hachi certificate) or
`RingSwitching.Lift.coordinateWiseSpecialSound` (the generic theorem).

blueprint/src/references.bib
```diff
@@ -492,6 +492,22 @@ @inproceedings{WW23
   publisher    = {Springer}
 }
 
+@misc{HMZ25,
+      author = {Meng-Yuan Michael Huang and Xinyu Mao and Jiapeng Zhang},
```
alexanderlhicks: **Wrong first author.** The paper's title page reads **"Mi-Ying (Miryam) Huang"**, not "Meng-Yuan
Michael Huang" — checked against the ePrint 2025/199 PDF (revision dated 2026-05-21, SHA-256
`c1832c6b…aef662`).

**The corrected entry already exists in #659.** `main` has no HMZ25 entry, both PRs add one, and
#659's spelling is right — taking theirs verbatim also resolves the add/add conflict, since this PR
uses `@misc`/`howpublished` and #659 uses `@article`/`journal` for the same key:

```bibtex
@article{HMZ25,
  title        = {Sublinear Proofs over Polynomial Rings},
  author       = {Huang, Mi-Ying Miryam and Mao, Xinyu and Zhang, Jiapeng},
  journal      = {Cryptology ePrint Archive},
  year         = {2025},
  url          = {https://eprint.iacr.org/2025/199}
}
```

Worth pinning the revision too — `year = {2025}` with no revision, for a revision dated 2026-05-21.

Heads-up on resolution order: this file is a **four-way** conflict point — #654, #656, #659 and the
open #615 all edit `blueprint/src/references.bib`. Probably cleanest to settle the bib in one PR and
have the others adopt it.

**Attribution boundary**, since it affects how the module docstrings should cite: the current HMZ
states its quotient-witness identity in the Galois-ring Ring-R1CS form ("there exists a function
`D : {0,1}^· → Gal(Q,d)[X]`"), not Hachi's linear `Mz = y + (X^d+1)·r` with `2d`-point extraction.
NOZ26 p. 18 is the exact source of the theorem formalized here and itself credits the originating
observation to HMZ. So cite NOZ26 for Lemma 9 and HMZ25 for the idea — which is what the module
docstrings already do. (HMZ also distinguishes its own "ring switching" from DP24's, which
independently corroborates this PR's Lift/Packing taxonomy split.)

docs/skills/prove-milestone.md
```diff
@@ -50,11 +55,17 @@ Maintain these invariants throughout the task:
...
-6. Do not use `native_decide`, `classical`, any `Classical.*` declaration, `axiom`, `admit`,
-   `Lean.ofReduceBool`, an unsafe/opaque proof surrogate, or a theorem whose transitive axiom set
-   contains `sorryAx`. Do not introduce project axioms. Treat `Classical.choice` as forbidden even
-   when it enters indirectly through a reused theorem. Record unavoidable foundational use of
-   `propext` or `Quot.sound`; do not call a result “axiom-free” if either appears.
+6. Do not use `native_decide`, a new `classical`/`Classical.*` shortcut in the milestone's own
+   proofs, `axiom`, `admit`, `Lean.ofReduceBool`, an unsafe/opaque proof surrogate, or a theorem
+   whose transitive axiom set contains `sorryAx`. Do not introduce project axioms. ArkLib's
+   accepted **axiom-clean** baseline — the standard the proven QuadEval milestone meets — is
+   exactly `{propext, Classical.choice, Quot.sound}` (Mathlib's foundational axioms);
```
alexanderlhicks, line 62: This file is out of scope for a ring-switching subprotocol PR, and it relaxes a **repo-wide** proof
policy — I would split the whole change out, or get explicit maintainer sign-off before it lands
inside a feature PR.

Two of the new claims are also wrong as written:

- "**exactly** `{propext, Classical.choice, Quot.sound}`" should be a **subset**. Several lemmas in
  this very PR need fewer (some elementary degree facts need none of `Quot.sound`), so the
  "exactly" form would reject them.
- Line 65, "attempting to would be impossible without rewriting Mathlib", is too strong as a global
  claim. An individual theorem can often be reproved constructively or routed around a
  choice-using combinator without touching Mathlib.

One more, if the section stays: checking only the milestone's own proof bodies leaves a loophole for
`Classical.choice` entering through a newly added target-specific helper. The audit should cover the
transitive closure of the target, which is what `#print axioms` on the headline declaration
actually gives you.

The olean/style/citation workflow notes in the same file are useful and can stay if they are
deliberately in scope.

ArkLib/ProofSystem/RingSwitching/Basic.lean
alexanderlhicks: "no single generic ring-switching reduction exists in this library, **and none is planned**" — the
**open** PR #615 `feat: even more general ring switch` (head `68e9a752`) is one: it adds
`RingSwitching/Generic/{Carrier,Packing,Batching,Recombine,Relations,Reduction}.lean` as a generic
framework intended to subsume `RingSwitchingProfile`.

The *substance* of the taxonomy survives, and it is worth stating precisely because it also bounds
#615's scope. #615 generalises the **Packing / DP24 tensor-carrier** direction — its vocabulary is
`Carrier`/`Packing`/`Batching`/`Recombine`, and `ringSwitchPhase_rbrKnowledgeSound`
(`Generic/Reduction.lean:405`) carries `[IsDomain car.E]`, a hypothesis `Rq Φ` does **not** satisfy.
Hachi's switch is the [HMZ25] lift-and-substitute, i.e. this PR's `Lift`. So #615 is a successor to
Packing and **cannot subsume `Lift`** while it needs a domain carrier. HMZ's own paper also
distinguishes its "ring switching" from DP24's, which independently corroborates the split.

Suggested wording: "no single *protocol* unifies Lift and Packing; the generic work in #615
generalises Packing." Since #615 also edits `docs/kb/concepts/ring-switching.md` — as does this PR —
it is worth settling the taxonomy once on that page rather than in two PRs with different wording.

**Mechanical follow-up, verified rather than inferred:** #615's `Generic/Carrier.lean:7` reads
`import ArkLib.ProofSystem.RingSwitching.Prelude`, and this PR renames that module to
`RingSwitching/Packing/Prelude.lean`. That file is new in #615, so it will not move with the rename
and #615 will fail to build after this merges. One-line fix on their side
(`import ArkLib.ProofSystem.RingSwitching.Packing.Prelude`) — worth a heads-up in this PR's body,
since #615 is already `CONFLICTING`.

ArkLib/Commitments/Functional/Hachi/RingSwitch/Reduction.lean
```diff
@@ -4,55 +4,46 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+
+  This file is the **cyclotomic instance** of generic `Lift`
+  (`ProofSystem/RingSwitching/Lift/`): the presentation is `Rq Φ` with canonical
+  reduced representatives (`cyclotomicPresentation`, laws discharged from the `Rq` quotient
+  bridge in `Data/Lattices/CyclotomicRing/QuotientLift.lean`). The generic layers supply
+  everything construction-shaped: the lifted witness and `checkAt` predicate, the
+  `2d`-point interpolation/descent recovery, the escape/collision/common-opening extractor,
+  and the CWSS plumbing. The composable package is therefore assembled **wholesale from generic
+  `Lift.package`** at `cyclotomicPresentation`; the single Hachi-specific obligation handed to it
+  is the norm implication `vecLInftyNorm_le_of_liftShort`. (See the note above `liftPackage` on
```
alexanderlhicks, line 21: Dangling forward reference: "(See the note above `liftPackage` on why the CWSS certificate is
exposed as `liftPackage.isCWSS` …)". There is no such note at lines 168-171 — only a restatement of
the same fact. Either add the explanation or drop the pointer.

It is worth adding rather than dropping, because the answer is interesting: exposing the certificate
as a package field instead of a named theorem in Hachi's own relation vocabulary is what lets
`Lift.package` be consumed wholesale, and it is also why the KB pages went looking for a
`lift_coordinateWiseSpecialSound` that does not exist.

ArkLib/Commitments/Functional/Hachi/RingSwitch/Reduction.lean
```diff
@@ -63,183 +54,133 @@ import ArkLib.OracleReduction.Security.CoordinateWiseSpecialSoundness.ScalarRoun
...
+
+/-- Coefficient-range predicate on the quotient polynomials. -/
 def RhoShort (ρBound : ℕ) (ρ : Fin n → Polynomial (ZMod q)) : Prop :=
   ∀ i k, ((ρ i).coeff k).valMinAbs.natAbs ≤ ρBound
 
-/-- The combined shortness predicate of the lifted witness — the norm side of `relLift`, and the
-`Short` parameter of the abstract commitment `LiftCom` (weak binding is norm-conditioned,
-[NOZ26] Lemma 7). -/
+/-- Hachi's norm-conditioned admissibility predicate for a lifted opening. -/
 def liftShort (bound ρBound : ℕ) (w : LiftedWitness Φ μ n) : Prop :=
```
alexanderlhicks, line 112: `RhoShort` is never consumed. `liftPackage` discharges the generic `short_zOk` obligation through
`vecLInftyNorm_le_of_liftShort`, which uses only `hshort.1`.

It isn't wrong — Figure 4's verifier does check `‖r‖∞ ≤ b − 1`, so modelling it is faithful, and
`R^lin` correctly bounds only `z`. But note the *direction*: `Short` sits inside
`CommittedScalar.rel`, i.e. in the theorem's **hypothesis**, so strengthening it shrinks the
adversary class the CWSS statement covers while buying nothing in the conclusion.

Either consume it (state the `ρ`-side bound in the output relation — #656 does use it, via
`hZero_eq_zero_imp_liftShort`, which derives an `r`-side bound as well as a `z`-side one), or add a
line saying it is a forward-compatibility hook for the digit layer so the next reader doesn't spend
time looking for the consumer.

ArkLib/ProofSystem/RingSwitching/Lift/Presentation.lean
```diff
@@ -0,0 +1,311 @@
...
+    linear_combination -hdiv
+
+/-- **The per-row recovery engine**, over an arbitrary presentation: if a row's lifted
+equation (with quotient `ρ` of degree `≤ d − 1`) holds under `evalAt` at `2d`
+pairwise-distinct points of a field `F`, the row equation holds over `S`
+([NOZ26] Lemma 9). The defect
+polynomial has degree `< 2d`, so the `2d` roots kill it; the resulting `R[X]` identity
+descends along the coset laws. The degree `d` is an explicit parameter tied to the modulus by
+`hd`, so instances can state witnesses against their own degree expression. -/
+theorem mulVec_eq_of_evalAt_rowSum {F : Type*} [Field F] {φF : R →+* F}
```
alexanderlhicks, line 299: The degree accounting here is correct and **tight**, and I think it deserves a docstring line
because it is the single most load-bearing constant in the PR: `rowSum` has
`natDegree ≤ 2d − 2` (two degree-`< d` representatives multiplied), `rep (y i) + modulus * ρ` has
`natDegree ≤ max(d−1, d + (d−1)) = 2d − 1`, both are `< 2d`, so `2d` pairwise-distinct points force
equality — matching NOZ26 Lemma 9's "degree at most `2d − 1`" exactly.

Also worth a note: `hA : Function.Injective A` with `A : Fin (2 * d) → F` silently requires
`|F| ≥ 2d`. For a smaller carrier the hypothesis is unsatisfiable and the lemma is vacuously true.
Benign in context — the CWSS tree could not exist either, and the paper's `(2d−1)/|F_{q^k}|` error
is only meaningful for `|F_{q^k}| > 2d` — but one sentence would stop someone instantiating at a
small carrier and believing something had been proved.

docs/kb/papers/NOZ26.md
alexanderlhicks: **The KB names a declaration that doesn't exist.** `lift_coordinateWiseSpecialSound` isn't defined anywhere in the tree — grepping `215b8868` gives 0 hits outside this page and `HMZ25.md:43`. The CWSS certificate is the `liftPackage.isCWSS` field; the generic theorem behind it is `RingSwitching.Lift.coordinateWiseSpecialSound`.

I've also narrowed "formalized and proven" to the simplified Figure 4 kernel here, with the fuller scope note suggested a few lines below.

```suggestion
- The **cyclotomic-ring → extension-field lift** (§4.3, Figure 4 / **Lemma 9**, following
  [`HMZ25`](HMZ25.md)): the *simplified* Figure 4 extraction kernel is **formalized and proven**
  (sorry-free, axiom-clean) as `liftPackage` — its CWSS certificate is the `liftPackage.isCWSS`
  field, and the generic theorem behind it is `RingSwitching.Lift.coordinateWiseSpecialSound` — in
  Hachi's
```

---

**Related, and not anchorable as a suggestion:** adding this claim creates a contradiction with the bullet at **`NOZ26.md:79`** — "ArkLib has not yet built the Hachi ring-switching instance" and Hachi soundness "is out of scope for the current ring-switching module". I checked: that text is **pre-existing on `main`** and falls outside this PR's diff hunks, so GitHub won't let me attach a suggestion to it. Re-scoping it to the packing abstraction resolves the clash, since that is what it was always about:

```markdown
- ArkLib has not yet built a Hachi instance of the **`RingSwitchingProfile`** (packing) abstraction;
  that abstraction is designed to admit one, but only the Binius instance is implemented. This is a
  statement about `ProofSystem/RingSwitching/Packing/` only — Hachi's own ring switch is the
  [`HMZ25`](HMZ25.md) quotient-evaluation lift, which *is* built, as `liftPackage` above.
```

Entirely your call whether to fold that in here or leave it for a separate cleanup — the two duplicated headings further down (`## Known Divergences From ArkLib` at 77 and 90, `## Version Notes` at 85 and 106) are likewise pre-existing and not this PR's doing.

docs/kb/papers/NOZ26.md
```diff
@@ -37,9 +37,25 @@ Commitment layer:
...
+  `OracleReduction/Security/CoordinateWiseSpecialSoundness/CommittedScalar.lean`), with the
+  presentation law-discharge lemmas in
+  `Data/Lattices/CyclotomicRing/QuotientLift.lean`. It is consumed at row 4 of the Hachi opening
+  chain (composed in `Hachi/Composition.lean`).
+  Design decisions recorded there: the never-sent `(z, r)` is the output-relation witness
+  (D6); the `w̃`-commitment is the abstract, norm-conditioned weak-binding `LiftCom`
+  (Remark 2 / Lemma 7), its binding break threaded backwards through all seams as an escape
+  budget (`Set.withEscape`, design G1); the witness type carries `deg ρᵢ ≤ d − 1`
+  (the paper's `Z_q^{<d}`); the extraction target is `R^lin` over `R_q`, equivalent to the
+  paper's `Z_q[X]` identity by the quotient-witness correspondence.
```
alexanderlhicks, line 60: Small addition: record the paper-model scope, so this page matches the "Paper-model boundary" note already in `RingSwitch/Reduction.lean:41-46`. What's formalized is the simplified raw-`(z, r)` Figure 4 / Lemma 9 kernel; the p. 18 honest protocol commits `(z, r₁, …, r_log_b(q))` with per-digit norm bounds (the paper's own "there is a hidden gadget decomposition of `r`"), and that encoding, its reconstruction identity, and an honest-prover completeness bound aren't formalized. Worth having on the KB page since that's what gets read first.

```suggestion
  (the paper's `Z_q^{<d}`); the extraction target is `R^lin` over `R_q`, equivalent to the
  paper's `Z_q[X]` identity by the quotient-witness correspondence.
  **Scope.** What is formalized is the simplified raw-`(z, r)` Figure 4 / Lemma 9 kernel. The
  paper's p. 18 honest protocol commits `(z, r₁, …, r_log_b(q))` with per-digit norm bounds —
  "there is a hidden gadget decomposition of `r`" — and that encoding, its reconstruction
  identity, and an honest-prover completeness bound are **not** formalized; `RhoShort` records
  the resulting admissibility requirement abstractly. See the "Paper-model boundary" note in
  `Hachi/RingSwitch/Reduction.lean`.
```

docs/kb/papers/HMZ25.md
alexanderlhicks: Same phantom declaration as `NOZ26.md:45` — `lift_coordinateWiseSpecialSound` doesn't exist. Suggest pointing at the package field instead:

```suggestion
  interpolation recovery, and `liftPackage` (whose CWSS certificate is its `isCWSS` field) are in
```

blueprint/src/references.bib
alexanderlhicks: **Wrong first author.** The ePrint 2025/199 title page reads **Mi-Ying (Miryam) Huang**, not "Meng-Yuan Michael Huang".

Worth fixing here rather than downstream: `main` has no `HMZ25` entry, so #654, #656 and #659 all add one, and **#656 and #659 already carry the correct spelling**. If this lands first with the typo, whoever resolves that three-way add/add conflict has to notice and pick the sibling's version. The suggestion below matches #656's spelling exactly.

(Both siblings also still say `year = {2025}` with no revision pin, for a revision dated 2026-05-21 — worth pinning at some point, but not here.)

```suggestion
      author = {Huang, Mi-Ying Miryam and Mao, Xinyu and Zhang, Jiapeng},
```

docs/skills/prove-milestone.md
```diff
@@ -50,11 +55,17 @@ Maintain these invariants throughout the task:
...
+   `Classical.choice` arriving through Mathlib or through an unmodified generic combinator (e.g.
+   `ReduceClaim.verifier_coordinateWiseSpecialSound`, which every CWSS milestone reuses) is
+   acceptable and need not be eliminated — attempting to would be impossible without rewriting
+   Mathlib. Confirm each target declaration's `#print axioms` is a subset of that baseline with no
```
alexanderlhicks: **Two things, and this file is also out of scope for a ring-switching PR** — it changes a repo-wide proof policy, `main` has neither wording, and no other open branch touches the file, so nothing downstream will correct it. Splitting it out would be cleanest; if it stays, these two need fixing:

1. **"exactly `{propext, Classical.choice, Quot.sound}`" contradicts line 66 of this same item**, which asks you to confirm `#print axioms` is a *subset* of the baseline. Line 66 is the right one — several declarations in this PR legitimately need fewer (some elementary degree facts don't need `Quot.sound`), so the "exactly" reading would reject your own work.
2. **"impossible without rewriting Mathlib" is too strong as a global claim.** An individual theorem can often be reproved constructively or routed around a choice-using combinator without touching Mathlib. Worth softening to "not required" rather than "impossible".

```suggestion
   `{propext, Classical.choice, Quot.sound}` (Mathlib's foundational axioms); a milestone may use
   any *subset* of it, and many results legitimately need fewer.
   `Classical.choice` arriving through Mathlib or through an unmodified generic combinator (e.g.
   `ReduceClaim.verifier_coordinateWiseSpecialSound`, reused by the CWSS milestones that go through
   it) is acceptable and need not be eliminated; eliminating it from an individual theorem is
   sometimes possible by reproving constructively or routing around a choice-using combinator, but
   it is not required. Confirm each target declaration's `#print axioms` is a subset of that
   baseline with no
```

ArkLib/ProofSystem/RingSwitching/Basic.lean
```diff
@@ -0,0 +1,102 @@
...
+families, one folder each:
```
alexanderlhicks: "and none is planned" — draft PR #615 `feat: even more general ring switch` (head `e5b94f4d`) is one.

The *taxonomy* is right and worth keeping, and it's worth saying why precisely, because it also bounds #615's scope: #615 generalises the **Packing** side (its vocabulary is `Carrier`/`Packing`/`Batching`/`Recombine`, and `ringSwitchPhase_rbrKnowledgeSound` carries `[IsDomain car.E]`, which `Rq Φ` doesn't satisfy), so it can't subsume `Lift`.

Unrelated heads-up while you're here: #615's `Generic/Carrier.lean:7` imports `ArkLib.ProofSystem.RingSwitching.Prelude`, which this PR renames to `Packing.Prelude`. That file is new in #615 so it won't move with the rename — one line on their side after this merges.

```suggestion
it inside a field. The constructions share *algebra*, not *protocol*; no single *protocol* in this
library unifies Lift and Packing. (Generic work on the Packing side is in progress — see draft
PR #615 — but it does not subsume Lift: it needs a domain carrier, which `Rq Φ` is not.) The two
construction families, one folder each:
```

alexanderlhicks (commented):
*Follow-up to my cross-PR review, from a pre-merge pass on the current head `215b8868`.*

**The code side of this PR looks ready to me.** I merged it with current `main` (`3975ccab`) and the
full gate set is green: `lake build` 4134 jobs, `check-imports.sh` (the auto-merged generated umbrella
is correct), `check-docs-integrity.py`, `scripts/kb/lint.py`, style lint 672 vs 674 on `main`. The
core is axiom-clean including `liftPackage` itself — `Lift.coordinateWiseSpecialSound`,
`CommittedScalar.coordinateWiseSpecialSound` and `liftPackage` all report exactly
`[propext, Classical.choice, Quot.sound]` post-merge. `main`'s #668 doesn't break anything (the
affected `KZG` files aren't touched here, so the merge takes `main`'s fixed versions). I also checked
that the six `Packing/` renames preserve their named API — the raw count differences turned out to be
multi-line anonymous `instance` declarations plus one docstring line my pattern misread, so no finding
there.

So everything below is documentation, and all of it is committable as-is. I've built and validated the
exact set locally on top of `215b8868`: build green (4133 jobs), all four validators pass, no lines
over 100 chars, no statement/term/proof touched.

Two things I deliberately left out:

- **`rangeProduct_eq_zero_iff` at `ZeroCheck/Constraints.lean:100`** is still `sorry` with zero Lean
  consumers. #656 proves it (axiom-clean, verified at `512a4179`), so it will be fixed when that
  lands; happy to paste the 18-line proof here instead if you'd rather not wait.
- The duplicated `## Known Divergences From ArkLib` / `## Version Notes` headings in `NOZ26.md` are
  **pre-existing on `main`**, not introduced here, so I've left them alone.

ArkLib/ProofSystem/RingSwitching/Basic.lean
tobias-rothmann: Added this comment, but removed the PR mention.

docs/skills/prove-milestone.md
tobias-rothmann: Second point is true, I updated that. First one, I think not.

docs/kb/papers/NOZ26.md
tobias-rothmann: Fixed the doc, but without the "sorry-free, axiom-clean" part. I think thats a but polluting the doc.

alexanderlhicks (approved):

PR #650, opened 2026-07-14
feat[Hachi]: skeleton

tobias-rothmann:
Skeleton for the Hachi opening proof with a corresponding artifact: 
in `Hachi/hachi-overview.html`

cc @ErVinuelas

PR #656, opened 2026-07-17
feat(Hachi): formalize lemma 10 of the paper

ErVinuelas:
## Motivation

Formalize Hachi Lemma 10 and the zero-check links built on the protocol skeleton from #650, while keeping protocol-facing polynomial computations executable.

## Previous behavior

The skeleton did not provide a sound extraction argument for uniform vector challenges: an axis cross does not determine a multilinear polynomial. Some zero-check polynomial seams also depended on noncomputable Mathlib representations.

## Changes

- derive zero-check evaluation points from scalar seeds along Kronecker curves
- prove the multilinear Kronecker root-counting and injectivity results used by extraction
- represent protocol-facing constraint polynomials with computable CompPoly multilinear values
- keep Mathlib polynomial views confined to proof bridges
- derive lift shortness from the batched range identity
- document the deviation from the paper and the resulting soundness argument

## Relationship

Builds on #650.


tobias-rothmann:
@ErVinuelas is there a reason why you used `ArkLib.Data.MvPolynomial` and not the CompPoly `CMvPolynomial` (these should be roughly the same, I guess, but CompPoly is probably more developed). 
Iirc the ArkLib local polys will also disappear at some point.

tobias-rothmann:
Also, I think you could use multilinear polynomials, no? $H_0$ and $H_\alpha$ should be multilinear iirc. So you could maybe use `CompPoly.CMlPolynomial`. 
@ErVinuelas

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Reduction.lean
```diff
@@ -62,134 +55,257 @@ import ArkLib.Commitments.Functional.Hachi.ZeroCheck.Batch
...
-def relZeroCheck (K : LiftCom (LiftedWitness Φ μ n) E (liftShort Φ bound ρBound))
+/-- The zero-check's output relation (Figure 5's residual claims): `w̃` opens `t`, is short
+(`liftShort`), and both batched constraint polynomials vanish at the points derived from the
+seeds — `H₀` at `κ_{m₀}(ρ₀)` and `H_α` at `κ_{m₁}(ρ_α)`. -/
+def relZeroCheck (K : LiftCom (LiftedWitness Φ μ n) E (liftShort Φ bound rBound))
     (φF : ZMod q →+* F) (b : ℕ) :
     Set (ZeroCheckStatement Φ K.TCom F n μ × LiftedWitness Φ μ n) :=
   {p |
     K.com p.2 = p.1.t ∧
+    liftShort Φ bound rBound p.2 ∧
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[high · the assumption]** `relZeroCheck` (and `relBatched`/`relLift`/`roundRel`) carry `liftShort` — a norm bound on the *opened* witness — as an explicit conjunct. In the paper, shortness is a *consequence* of the protocol: `H₀(τ₀)=0` whp ⇒ `H₀≡0` ⇒ every committed coefficient is in `[−(b−1),b−1]` ⇒ the opening is short (this is what feeds Lemma 7 weak binding). Here it is *assumed* instead.

Why it matters: the verifier is a pure pass-through, so all checks live in the output relation. `H₀`/`H_α` are checks a real verifier *can* perform (the downstream sumcheck discharges them), but `liftShort` is a norm bound a real verifier *cannot* perform (it never opens the commitment). So this conjunct adds an unverifiable check — exactly the guarantee the range machinery is meant to supply — and the formalized knowledge soundness ends up *assuming* the committed witness is short rather than *proving* it. See the linked comments on `Batch.lean` and `rangeProduct_eq_zero_iff`.

_Suggested fix (resolution "option 1", already set up in the code):_ prove `H₀≡0 ⇒ liftShort` and drop the conjunct, deriving it.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Batch.lean
```diff
@@ -43,64 +40,78 @@ open OracleComp OracleSpec ProtocolSpec CoordinateWise
...
-
-/-- **The batching bridge as a `CWSSPackage`**: zero-round `ReduceClaim` at `mapStmt := id`,
-reducing `relLiftE` to `relBatchedE` with no soundness error (the whole content is the sorried
-un-batching pull-back). -/
+    (h : (X, w) ∈ relBatchedE Φ m₀ m₁ bound rBound K φF b) :
+    (X, w) ∈ relLiftE Φ bound rBound K φF := by
+  rcases w with w | e
+  · -- real witness: only the per-row equation needs work
+    simp only [relBatchedE, Set.mem_withEscape_inl, relBatched, Set.mem_setOf_eq] at h
+    obtain ⟨hcom, hshort, _hZero, hAlphaZ, hbound⟩ := h
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[high · the discard]** The `_hZero` binder shows this proof destructures the `H₀≡0` hypothesis and then **discards it** — shortness is taken from the `hshort` (`liftShort`) conjunct instead. So the Kronecker root-counting spends real effort establishing `H₀≡0` as an output obligation, and nothing downstream consumes it: the range/smallness half of the zero-check is inert. This is the concrete site of the finding on `ZeroCheck/Reduction.lean`.

_Suggested fix:_ use `_hZero` to prove shortness (via `rangeProduct_eq_zero_iff`) rather than reading it off `hshort`.

alexanderlhicks (commented):

### Additional notes (lines outside the diff hunks)

- **`ArkLib/ProofSystem/Component/CheckClaim.lean:39`** — **[high · regression]** This docstring references `oracleReduction_completeness`, but the theorem itself was **deleted** — `git diff origin/main` shows a pure deletion introduced by the merge commit `ffcc20ea` ("Merge branch 'main' into lemma10-on-650"). It's a proven, sorry-free theorem (55 lines on main) that every sibling component (`DoNothing`, `SendWitness`, `ReduceClaim`) still keeps. The build isn't broken (nothing depends on it), but a proven theorem was lost as a merge-resolution casualty and this docstring now dangles. Please restore it from `main`.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
```diff
@@ -63,182 +62,269 @@ namespace ArkLib.Lattices.Ajtai.InnerOuter
...
+/-! ## The range factor and the table -/
+
+/-- Hachi Eq. (23)'s per-entry range factor `P_b(v) = v·∏_{j=1}^{b-1} (v - j)·(v + j)`, the
+vanishing polynomial of the symmetric range `{-(b-1), …, b-1}`. -/
+def rangeProduct (b : ℕ) (v : F) : F :=
+  v * ∏ j ∈ Finset.Icc 1 (b - 1), ((v - (j : F)) * (v + (j : F)))
+
+/-- Over a field, `P_b(v) = 0` iff `v` is the image of an integer in the symmetric range
+`{-(b-1), …, b-1}`. -/
+theorem rangeProduct_eq_zero_iff {b : ℕ} {v : F} :
```
alexanderlhicks, line 98: 🤖 _AI-generated (Claude Code)_

**[high · the unused bridge]** This is precisely the tool that would bridge `H₀≡0 ⇒ liftShort` (it characterizes the range product's roots as `{0,±1,…,±(b−1)}`), but it has **zero consumers** (grep-confirmed). Its presence-but-disuse is the tell that resolution option 1 was set up and then deferred. To close the finding: apply `rangeProduct_eq_zero_iff` (you'll need to reintroduce the `2b ≤ q+1` / `2b−1 < q` hypothesis dropped from `batchPackage`, plus injectivity of `φF` on centered representatives) to derive `liftShort` from `_hZero`, then remove `liftShort` from the relations.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
```diff
@@ -63,182 +62,269 @@ namespace ArkLib.Lattices.Ajtai.InnerOuter
...
 
-/-- Per-round univariate degree of the range sumcheck (`F_{0,τ₀}`): the `2b − 1`-factor range
-product over the multilinear `w̃`, times the multilinear `eq̃` — degree `2b` (pin R8). -/
-def roundDegZero (b : ℕ) : ℕ := 2 * b
+/-- The number of distinct challenge seeds the zero-check needs per coordinate,
+`D := max(2, 2^{m₀}, 2^{m₁})`. `2^{m₀}` and `2^{m₁}` are the degrees of the univariate Kronecker
+restrictions of `H₀` and `H_α`, so that many roots determine each polynomial; the floor of `2`
+meets the `2 ≤ k` requirement of the coordinate-wise special soundness structure. The paper's
+Lemma 10 uses `max(2d, 2b−1)`. -/
+def zeroCheckD (m₀ m₁ : ℕ) : ℕ := max 2 (max (2 ^ m₀) (2 ^ m₁))
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[praise · no action]** For the record: this deviation from the paper's `D = max(2d, 2b−1)` is correct and well-documented. The corrected Lemma 10 (Kronecker-curve repair for the real axis-cross flaw in NOZ26's proof) is a genuinely valuable contribution, and `D` here is exactly the per-polynomial pullback degree bound. Confirming the deviation is sound and the docstring explains it well.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
```diff
@@ -63,182 +62,269 @@ namespace ArkLib.Lattices.Ajtai.InnerOuter
...
+/-- The Boolean-point coefficients of `H_α` (Eq. (22)): the `M̃_α`-contracted per-row value
+`i ↦ (∑_{u,ℓ} M̃_α(i,u)·w̃(u,ℓ)·α̃(ℓ)) − yᵢ(α)`.
+
+By the "represent the constraints by polynomials" identity of [NOZ26] §4.3, this contraction
+equals the `α`-evaluated per-row defect of the lift relation `relLift`, namely
+`evalAt α (∑ⱼ Mᵢⱼ·zⱼ) − evalAt α yᵢ − evalAt α (X^d+1)·evalAt α rᵢ`. It is defined here as that
+defect, encoded into the `m₁`-cube via `rowPoint` and zero-padded on rows `≥ n`; the row equation
+of `relLift` (Fig. 4) is recovered at the Boolean points via `hAlphaEvals_rowPoint`. The `b`
+argument is unused here and kept for signature compatibility with `hAlpha`. -/
+noncomputable def hAlphaEvals (φF : ZMod q →+* F) (_b : ℕ) (s : RlinStatement Φ n μ) (α : F)
```
alexanderlhicks, line 155: 🤖 _AI-generated (Claude Code)_

**[faithfulness note]** `hAlphaEvals` is defined directly as the α-evaluated per-row defect, so the paper's three-case matrix `M̃_α(i,u)` and `α̃(ℓ)=α^ℓ` (Eq. 22) are not separate objects and the "represent constraints by polynomials" identity is baked into the definition rather than derived. I checked the collapse is mathematically correct, so this is faithful in *result*. A one-line docstring noting the collapse (cross-referencing Eq. 22) would let readers verify against the paper without re-deriving it. (Also: the `_b` parameter is dead — drop it or note why it's kept for signature compatibility.)

docs/kb/audits/noz26-zero-check-lemma10.md
```diff
@@ -0,0 +1,110 @@
...
+  *soundness* `H₀ ≡ 0 ⇒ liftShort` (reconciling the single `rangeProduct b` with `liftShort`'s two
+  bounds `bound`/`ρBound`), and the sumcheck-summand stubs (`sumcheckPoly*`, `hypercubeSum`,
+  `sum_*`).
+- **Link 5 (batching bridge).** The un-batching pull-back `mem_relLiftE_of_relBatchedE`
+  (`relBatchedE → relLiftE`, `ZeroCheck/Batch.lean`) is now **proof-`sorry`-free**: `H_α ≡ 0` gives
+  every `eq̃`-coefficient zero (`MvPolynomial.MLE_eq_zero_iff`), and `hAlphaEvals_rowPoint`
+  identifies the coefficient at row `i` with `relLift`'s per-row `α`-equation; `K.com`/`liftShort`/
+  bound are carried verbatim. It needs only the arity pin `n ≤ 2 ^ m₁`; the range-side hypotheses
+  `2b ≤ q+1`, `b-1 ≤ bound` were **removed as unused** (shortness is carried directly, resolution
+  option 2). Its residual `sorryAx` is inherited solely from the `wTable`/`H₀` conjunct in the
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[audit doc · false claim]** This "residual sorryAx is inherited solely from the wTable/H₀ conjunct" claim is false and contradicts line 35 of this same doc ("proven, axiom-clean"). Confirmed by `#print axioms` in a built worktree: both `zeroCheck_coordinateWiseSpecialSound` and `mem_relLiftE_of_relBatchedE` depend only on `[propext, Classical.choice, Quot.sound]` — no `sorryAx`. `relBatched` references only concrete sorry-free defs, never the sorried sumcheck stubs. The claim is stale from when `wTable`/`hZero` were themselves sorried. Per the repo's own rule ("a wrong audit doc is worse than none") please correct this line.

alexanderlhicks (commented):
*Part of a cross-PR review of #654, #656 and #659 read together (plus the open ring-switching PR #615), in isolated builds, against NOZ26 (ePrint 2026/156, SHA-256 `c7c98591…afd5e6a15`) — Eqs. (21)–(23), Fig. 5, Lemmas 10–11. Where a finding interacts with a sibling PR I say so explicitly.*

## #656 — reviewed at `1c94b646`

**The headline mathematics here is the best in the three-PR stack, and I'd like that not to get
diluted in the rework.**

I derived the Kronecker repair independently from the paper before reading the code and arrived at
exactly `multilinear_eq_zero_of_kronecker_roots`: writing `H` in the monomial basis,
`∏_{j∈S} Xⱼ ↦ ρ^{Σ_{j∈S} 2^j}`, and `S ↦ Σ_{j∈S} 2^j` is the base-2 bijection onto `{0,…,2^m−1}`, so
distinct multilinear monomials land on distinct univariate powers. That makes `powAlgHom` injective
on multilinears and gives a **deterministic** polynomial-identity equivalence rather than a
Schwartz–Zippel bound — strictly stronger than what Lemma 10 needed.

And the paper genuinely is broken as printed. `H₀` is multilinear in `t`, CWSS only pins vanishing
on the axis cross through the centre, and `H(t₁,t₂) = (t₁−a)(t₂−b)` is an explicit nonzero
multilinear vanishing on both lines through `(a,b)`. Shipping
`exists_nonzero_multilinear_vanishing_on_kronecker_seeds` alongside the positive result is exactly
the right instinct, and it turns out to be load-bearing for the design (see the `relZeroCheck`
comment).

Verified by `#print axioms` in this PR's own build: `powAlgHom_eq_zero_iff`,
`multilinear_eq_zero_of_kronecker_roots`, `exists_nonzero_multilinear_vanishing_on_kronecker_seeds`,
`rangeProduct_eq_zero_iff`, `hZero_eq_zero_imp_liftShort`, `mem_relLiftE_of_relBatchedE` and
`zeroCheck_coordinateWiseSpecialSound` are all clean —
`[propext, Classical.choice, Quot.sound]`, no `sorryAx`. Changed-file `sorry` bodies go **29 → 21,
eight removed, none added**. I also re-derived the repair's cost and it's affordable: `D = 2^{m₀}`
raises the branching arity from the paper's `O(d+b)` to `O(μ·d)`, so by Lemma 4 the knowledge error
becomes about `2(μ+n)d/|F_{q^k}|²` — still negligible, tree still polynomial. Worth saying in the
docstring; it's the obvious reviewer objection and the answer is favourable.

Four things I'd like changed, none of them a false theorem:

1. **Four sorried *definitions*** (`sumcheckPolyZero`, `sumcheckPolyAlpha`, `zcTargetAlpha`,
   `hypercubeSum`) with two sorried theorems relating them **to each other**. That's a sharper
   hazard than a sorried theorem: the seam is currently unfalsifiable. Good news — I checked both
   theorem *statements* against the paper and they are faithful, sign included.
2. **`docs/kb/audits/noz26-zero-check-lemma10.md:103` states a false arity mismatch** and prescribes
   two fixes that would each break the (correct) `sum_sumcheckPolyAlpha`. This one will actively
   mislead whoever implements F7.
3. **Two undocumented paper divergences that are *correct* but need writing down**: the `1_{≤μ}`
   factor the paper's `F_{0,τ₀}` carries and Eq. (23)'s `H₀` doesn't, and Lemma 10's `D` vs
   `SS(F,2,D)`'s `2D−1`.
4. **Dead scope**: `ToCompPoly/Multivariate/Restrict.lean` and most of the new ToComp multilinear
   module have no consumers; `docs/skills/make-pr-ready.md` is unrelated to Lemma 10.

On `relZeroCheck` carrying `liftShort`: I went looking for circularity and the loop does close
correctly — `relBatched` carries the *identity*, not shortness, and the norm bounds are derived from
it. The residual limitation is real but honestly documented and provably unavoidable in the collision
branch. So that one is a claims issue, not a code issue.

**On #615** (open, head `68e9a752`): it adds `OracleReduction/Security/ChallengeRound.lean`, so
there'll be two `ChallengeRound.lean` modules under `Security/` — see that comment. #615 also
overlaps this PR on `ArkLib/Data/MvPolynomial/Multilinear.lean` and `blueprint/src/references.bib`
(the latter is a four-way conflict point across #615/#654/#656/#659).

**Ordering: #654 → #659 → #656.** This should land third and as a *semantic port* rather than a
merge — #654 replaces the `RingSwitch/Reduction.lean` and `ZeroCheck/*` shapes this branch builds on
with the generic `Lift.Presentation` / `BindingCommitment` layer, and this branch modifies
`HACHI_RING_SWITCHING_PLAN.md`, which both #654 and #659 delete. Please carry across the Kronecker
repair, its sharpness companion, `rangeProduct_eq_zero_iff` (#654 currently `sorry`s it — this PR's
proof is the fix), `hZero_eq_zero_imp_liftShort`, `mem_relLiftE_of_relBatchedE`, and the generic
`ChallengeRound`.

ArkLib/Data/MvPolynomial/LinearMvExtension.lean
```diff
@@ -241,11 +241,237 @@ lemma powContraction_is_right_inverse_to_linearMvExtension
...
+    Submodule.sub_mem _ H.2 H'.2
+  have hz : powAlgHom (H.val - H'.val) = 0 := by rw [map_sub, h, sub_self]
+  exact Subtype.ext (sub_eq_zero.mp ((powAlgHom_eq_zero_iff hmem).mp hz))
+
+/-- **Root-counting on the Kronecker curve (the extraction step of repaired Lemma 10).** If a
+multilinear `H` vanishes on the curve at `2^m` distinct scalar seeds, then `H = 0`. Concretely: the
+pullback `powAlgHom H` has degree `< 2^m` but acquires `≥ 2^m` distinct roots, so it is the zero
+polynomial, whence `H = 0` by Kronecker injectivity. This is the deterministic replacement for the
+false "vanishes on the star ⇒ zero polynomial" step, giving `k = D = 2^m`-special soundness. -/
+theorem multilinear_eq_zero_of_kronecker_roots [IsDomain F]
```
alexanderlhicks: Before any criticism: **this is correct, and it is the right repair.** I derived the argument
independently from the paper before reading the PR and arrived at exactly this. Writing `H` in the
monomial basis, `∏_{j∈S} Xⱼ ↦ ρ^{Σ_{j∈S} 2^j}`, and `S ↦ Σ_{j∈S} 2^j` is the base-2 bijection onto
`{0,…,2^m−1}`, so distinct multilinear monomials land on **distinct** univariate powers — hence
`powAlgHom H = 0 ↔ H = 0`, and `2^m` distinct seeds kill a degree-`< 2^m` univariate.

Two things worth emphasising because they're easy to undersell:

- this is a **deterministic identity equivalence**, not a Schwartz–Zippel bound, so it is *stronger*
  than what Lemma 10 actually needed;
- the paper really is broken as printed. `H₀` is multilinear in `t`, so CWSS pins vanishing only on
  the axis cross through the centre, and `H(t₁,t₂) = (t₁−a)(t₂−b)` is an explicit nonzero multilinear
  vanishing on both lines through `(a,b)`. Confirmed axiom-clean here.

ArkLib/Data/MvPolynomial/LinearMvExtension.lean
```diff
@@ -241,11 +241,237 @@ lemma powContraction_is_right_inverse_to_linearMvExtension
...
+push it up with `linearMvExtension`; `powAlgHom` inverts that, so `H ≠ 0`, while
+`eval_powAlgHom_eq_eval_kronecker` turns each root of `p` into a curve evaluation of `H`.
+
+This is the exact companion of `multilinear_eq_zero_of_kronecker_roots`, and it is why the Hachi
+zero-check's seam relations must carry the range *identity* rather than deriving it from their point
+evaluation: the extractor's collision branch shares an opening across only `2^m − 1` seeds — one
+short of the root count — so it could never conclude `H₀ ≡ 0` for a single colliding opening. See
+`ArkLib/Commitments/Functional/Hachi/ZeroCheck/Reduction.lean` (`relZeroCheck`) and
+`docs/kb/audits/noz26-zero-check-lemma10.md`. -/
+theorem exists_nonzero_multilinear_vanishing_on_kronecker_seeds [Nontrivial F] (s : Finset F)
```
alexanderlhicks: Shipping the sharpness companion next to the positive result is exactly right, and it turns out to
be **load-bearing for the design argument**, not just a nicety: it is what proves that the
extractor's collision branch — which sees at most `2^m − 1` seeds for any single opening — can never
derive `H₀ ≡ 0`, which in turn is why `relZeroCheck` has to carry `liftShort` semantically. I'd
reference it from the `relZeroCheck` docstring for that reason.

I'd also surface this and `multilinear_eq_zero_of_kronecker_roots` in the PR description. Right now
the strongest content in the PR is buried in a `Data/` file and easy to miss.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
```diff
@@ -61,184 +75,480 @@ import Mathlib.Algebra.MvPolynomial.Basic
...
+
+/-- The range sumcheck summand `F_{0,τ₀}`, characterized by `∑_{x} F_{0,τ₀}(x) = H₀(τ₀)`
+(`sum_sumcheckPolyZero`) with per-variable degree `roundDegZero b = 2b`.
+
+Unlike `H₀`/`H_α` this one is genuinely evaluated by the prover in every sumcheck round. It is not
+multilinear, so it remains a general `CMvPolynomial`. -/
 def sumcheckPolyZero (φF : ZMod q →+* F) (b : ℕ) (τ₀ : Fin m₀ → F)
-    (w : LiftedWitness Φ μ n) : MvPolynomial (Fin m₀) F :=
+    (w : LiftedWitness Φ μ n) : CMvPolynomial m₀ F :=
   sorry
```
alexanderlhicks: Four sorried **definitions** in a row — `sumcheckPolyZero` (here), `sumcheckPolyAlpha:448`,
`zcTargetAlpha:454`, `hypercubeSum:460` — and then `sum_sumcheckPolyZero:463` and
`sum_sumcheckPolyAlpha:478` are sorried *theorems relating those opaque definitions to each other*.
(The primed aliases at 471/487 just forward.)

This is a sharper hazard than a sorried theorem. The terms are opaque, so nothing constrains them,
and a statement whose both sides are opaque cannot be wrong **yet** — it can only become wrong
later, silently, when the bodies arrive. There's no `example`/`#check` anywhere pinning even one
value of `hypercubeSum`, so there's nothing to catch a mismatch at that point.

Suggestion: either supply the four definitions, or demote them to `opaque`/section parameters and
take the two identities as **hypotheses** of the consumers. Then the seam is honestly a parameter
rather than a `theorem` that reads as proved.

Credit where it's due, and this is the part most likely to have been subtly wrong: I checked both
theorem **statements** against the paper and they're faithful, sign included. Paper p. 22 gives
`Σ_{u,ℓ} F_{α,τ₁} = a` with `a = Σ_i eq̃(τ₁,i)·y_i(α)`, and
`H_α(τ₁) = Σ_{u,ℓ} w̃·α̃·Σ_i eq̃·M̃_α − Σ_i eq̃·y_i(α)`, so `Σ F = H_α(τ₁) + a` — which is exactly
`hypercubeSum … = eval τ₁ hAlphaML + zcTargetAlpha`. Same for `sum_sumcheckPolyZero` against
`H₀(τ₀)`.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
```diff
@@ -61,184 +75,480 @@ import Mathlib.Algebra.MvPolynomial.Basic
...
+/-- Partial sum of `H` over the trailing cube coordinates: `hypercubeSum H i cs =
+∑_{x ∈ {0,1}^{m₀ − i}} H(cs, x)`, where `cs` fixes the first `i` coordinates. Operates on the
+computable representation — this is the fold the prover actually runs. -/
+def hypercubeSum (H : CMvPolynomial m₀ F) (i : ℕ) (cs : Fin i → F) : F :=
   sorry
 
-/-- The sum of `F_{0,τ₀}` over the cube is `H₀(τ₀)` — the algebraic identity behind the
-sumcheck bridge (`Sumcheck/Bridge.lean`). **Sorried (F5).** -/
+/-- The full-cube sum of the range summand `F_{0,τ₀}` equals `H₀(τ₀)`. -/
 theorem sum_sumcheckPolyZero (φF : ZMod q →+* F) (b : ℕ) (τ₀ : Fin m₀ → F)
```
alexanderlhicks, line 779: Undocumented (and **correct**) divergence from the printed paper, worth recording next to the other
Lemma-10 corrections since a reader comparing against Figure 5 will otherwise assume a bug.

The paper's `F_{0,τ₀}` (p. 22) carries a trailing indicator factor `1_{≤μ}(x,y)`, restricting the
range check to the `z` rows. Eq. (23)'s `H₀` has **no** such factor, sums over all `ℓ,u`, and the
bullet above it imposes the constraint "for each `u ∈ [μ + n]` and `ℓ ∈ [d]`" — i.e. on `r` too,
consistent with the earlier `‖z‖∞, ‖r‖∞ ≤ b − 1`. So the paper's own
`Σ_{u,ℓ} F_{0,τ₀}(u,ℓ) = H₀(τ₀)` is **false as printed**; the two sides differ by the indicator.

This PR follows the Eq. (23) reading — no indicator, range constraint on both the `z` and `ρ` rows,
visible in `hZero_eq_zero_imp_liftShort` discharging a `z`-side *and* an `r`-side bound. That's the
self-consistent choice and I agree with it; it just needs to be written down as a deliberate
divergence.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
```diff
@@ -61,184 +75,480 @@ import Mathlib.Algebra.MvPolynomial.Basic
...
+/-- The Boolean-point coefficients of `H_α` (Eq. (22)): the `M̃_α`-contracted per-row value
+`i ↦ (∑_{u,ℓ} M̃_α(i,u)·w̃(u,ℓ)·α̃(ℓ)) − yᵢ(α)`.
+
+By the "represent the constraints by polynomials" identity of [NOZ26] §4.3, this contraction
+equals the `α`-evaluated per-row defect of the lift relation `relLift`, namely
+`evalAt α (∑ⱼ Mᵢⱼ·zⱼ) − evalAt α yᵢ − evalAt α (X^d+1)·evalAt α rᵢ`. It is defined here as that
+defect, encoded into the `m₁`-cube via `rowPoint` and zero-padded on rows `≥ n`; the row equation
+of `relLift` (Fig. 4) is recovered at the Boolean points via `hAlphaEvals_rowPoint`. The `b`
+argument is unused here and kept for signature compatibility with `hAlpha`. -/
+def hAlphaEvals (φF : ZMod q →+* F) (_b : ℕ) (s : RlinStatement Φ n μ) (α : F)
```
alexanderlhicks: `hAlphaEvals` defines the desired `α`-evaluated row defect **directly**, so Eq. (22)'s public
contraction `Σ_u M̃_α(i,u) · w̃(u,ℓ) · α̃(ℓ)` is never constructed and never proved equivalent to it.

`hachi-overview.html:607` already admits this ("the paper's Eq. (22) contraction … and its
equivalence to that defect, are not formalized"), which is good — but the gap is currently visible
only in the HTML, while downstream `finalCheck` promises a *public* `M̃_α` evaluation that nothing
here supplies, and the six sumcheck stubs hide the join. Either build the contraction and prove the
equivalence, or state the scope as an abstract direct-defect variant at the *relation* level so it
survives someone reading only the Lean.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
```diff
@@ -61,184 +75,480 @@ import Mathlib.Algebra.MvPolynomial.Basic
...
+  LinearMvExtension.kroneckerPoint (m := m) ρ
 
-/-- **Eq. (23)**: the `eq̃`-batched range polynomial
-`H₀(τ) := ∑ᵢ eq̃(τ, i)·w̃(i)·∏_{j=1}^{b−1}(w̃(i) − j)(w̃(i) + j)` — multilinear in `τ ∈ F^{m₀}`;
-`H₀ ≡ 0` iff every table entry is a root of the range product, i.e. lies in `[−(b−1), b−1]`
-(under `2b − 1 < q`). **Sorried (F5).** -/
-def hZero (φF : ZMod q →+* F) (b : ℕ) (w : LiftedWitness Φ μ n) :
-    MvPolynomial (Fin m₀) F :=
-  sorry
+/-- The number of distinct challenge seeds the zero-check needs per coordinate,
```
alexanderlhicks: Two things worth adding to this docstring, both of which strengthen the PR's case rather than
weaken it.

**1. The repair's cost, since it's the obvious objection.** `zeroCheckD = max 2 (max (2^m₀) (2^m₁))`
replaces the paper's `max(2d, 2b−1)`, and since `2^m₀ ≥ (μ+n)·d` the branching arity grows from
`O(d + b)` to `O(μ·d)`. By NOZ26 Lemma 4 the knowledge error becomes `ℓ·k/|S|^ℓ = 2·D/|F_{q^k}|²`,
i.e. roughly `2(μ+n)d/|F_{q^k}|²` instead of `2·max(2d,2b−1)/|F_{q^k}|²` — still negligible at
Hachi's field size, and the tree stays polynomial. One sentence pre-empts the question.

**2. A second printed-Lemma-10 defect that supports the repair.** Lemma 10 asks for "`D` valid
transcripts … `∈ SS(F_{q^k}, 2, D)`", but `SS(S,ℓ,k)` has `ℓ(k−1)+1` elements, so at `ℓ=2, k=D`
that's **`2D−1`** — which is what this code actually uses. And the prose immediately above Lemma 10
says to treat `(τ₀,τ₁)` as `log μ + log d + log n` coordinates, contradicting the lemma's own
`ℓ = 2`. (Also, `τ₀` is drawn from `F^{log μ + log d}` on p. 20 although `w̃`'s domain is
`[μ+n] × [d]`, so the arity there should be `log(μ+n) + log d`.)

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Reduction.lean
```diff
@@ -61,135 +54,275 @@ import ArkLib.Commitments.Functional.Hachi.ZeroCheck.Batch
...
+The primary polynomials `hZero`/`hAlpha` are `CMlPolynomialEval` Boolean-value vectors.
+The relation evaluates them directly with `CMlPolynomialEval.eval`; the derived Mathlib views
+`hZeroML`/`hAlphaML` are used only inside the Kronecker root-counting proof.
+
+The shortness conjunct is a temporary semantic admissibility condition needed by the
+norm-conditioned weak-binding escape `K.collision_mem`: a single point evaluation of `H₀` does
+not imply that the corresponding opening is short. In particular, this relation deliberately
+does not assume the global identity `H₀ ≡ 0`; that identity remains the conclusion extracted by
+the zero-check into `relBatchedE`. -/
+def relZeroCheck (K : LiftCom (LiftedWitness Φ μ n) E (liftShort Φ bound rBound))
```
alexanderlhicks: `relZeroCheck` requires `liftShort Φ bound rBound p.2`, which a real Figure 5 verifier cannot
check — it sees `t = Com(w̃)` and two point evaluations, and establishing the range bound is the
entire *purpose* of `H₀`. So the CWSS theorem's hypothesis excludes precisely the adversary this step
exists to catch: a non-short prover falls outside `relZeroCheck` and the theorem says nothing about
it.

I want to be fair about severity, because I went looking for circularity and **did not find it**:

- the loop closes in the right direction — `relBatched` (`Batch.lean:55-62`) carries the **identity**
  `hZero = 0`, not shortness, and `hZero_eq_zero_imp_liftShort` derives the norm bounds from it for
  both the `z` and `ρ` rows;
- the obstruction is provably unavoidable in the collision branch, and this PR proves it itself: one
  seed per divergent opening means at most `2^m − 1`, and
  `exists_nonzero_multilinear_vanishing_on_kronecker_seeds` shows that can never suffice;
- the docstring at 124-128 and the overview row are both candid about it.

So the ask is about **claims, not code**: nothing here should be described as a complete
formalization of Lemma 10, and the residual obligation — make `relZeroCheck` carry only
verifier-checkable predicates, or add a third extractor branch for the not-both-short case — deserves
to be a named gap (e.g. in `docs/kb/audits/`) rather than an adjective in a docstring.

docs/kb/audits/noz26-zero-check-lemma10.md
```diff
@@ -0,0 +1,214 @@
...
+  vector identities.
+- **Still stubs, hence not computable in substance.** The sumcheck summands
+  `sumcheckPolyZero`, `sumcheckPolyAlpha`, the public target `zcTargetAlpha` and the prover fold
+  `hypercubeSum` have CompPoly *types* (`CMvPolynomial m₀ F`, `F`) but `sorry` bodies, as do
+  `sum_sumcheckPolyZero`/`sum_sumcheckPolyAlpha` (milestone F7). Their intended computable shapes
+  are `F_{0,τ₀} = eq̃(τ₀, ·) · P_b(mle[w̃])` (per-variable degree `2b = roundDegZero b`) and
+  `F_{α,τ₁} = eq̃(τ₁, ·) · (mle[rowSum(α)] − φ(α)·mle[r(α)])` (per-variable degree
+  `2 = roundDegAlpha`), with `zcTargetAlpha = ∑ᵢ eq̃(τ₁, i)·yᵢ(α)` accounting for the public
+  right-hand side.
+- **Arity mismatch to resolve before F7.** `sumcheckPolyAlpha` returns `CMvPolynomial m₀ F` while
```
alexanderlhicks: **This "Arity mismatch to resolve before F7" claim is false, and both fixes it proposes would break
something that is currently correct.** Flagging it as blocking because it's a doc that actively
misdirects implementation work.

In the paper, `F_{α,τ₁}(x,y) := w̃(x,y) · α̃(y) · Σ_i eq̃(τ₁,i) · M̃_α(i,x)` is summed over the `w̃`
coordinates `(u,ℓ)` — the `m₀`-cube — while the row index `i ∈ [n]` (the `m₁` side) is summed
**internally**, inside the `Σ_i eq̃(τ₁,i)·M̃_α(i,x)` factor. It is not a cube coordinate, so there is
nothing to over-count. The existing `CMvPolynomial m₀ F` signature with `τ₁ : Fin m₁ → F` is the
correct one.

Why the proposed remedies are harmful:

- inserting `∏_{j ≥ m₁}(1 − Xⱼ)` would change the sum away from `H_α(τ₁) + a`, breaking
  `sum_sumcheckPolyAlpha` — whose statement I checked against p. 22 and which is currently faithful;
- re-typing to `CMvPolynomial m₁ F` would make the `w̃(x,y)` factor untypable, since `w̃` lives on the
  `m₀`-cube.

Please delete the claim (and the matching NOZ26 action item).

ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/ChallengeRoundTree.lean
```diff
@@ -0,0 +1,327 @@
+/-
```
alexanderlhicks: Naming collision worth catching before both land: the **open** ring-switching PR #615 (head
`68e9a752`) adds `ArkLib/OracleReduction/Security/ChallengeRound.lean` — 113 lines,
`probEvent_challengeRound_le` / `probEvent_rbrGame_le`, probability factoring for RBR games.

The contents are genuinely different from this file's CWSS challenge-round tree and extractor, so
there's no duplicated mathematics. But two `ChallengeRound.lean` modules under
`OracleReduction/Security/`, one nested a level deeper, is a routing trap for
`docs/wiki/repo-map.md` and for `git grep`. Renaming one — this to `ChallengeRoundTree.lean`, or
#615's to `RbrChallengeRound.lean` — is free now and awkward once both are merged.

(#615 also overlaps this PR on `ArkLib/Data/MvPolynomial/Multilinear.lean` and
`blueprint/src/references.bib`.)

ArkLib/ToCompPoly/Multivariate/Restrict.lean
```diff
@@ -0,0 +1,59 @@
+/-
```
alexanderlhicks: This module (3 declarations) is imported **only** by the generated umbrella `ArkLib.lean:287` — no
production consumer anywhere in the tree. Same story for 13 of the 14 declarations in the new ToComp
multilinear module, where only `eval_eq_MvPolynomial_MLE` is actually used.

Under ArkLib's reuse/no-duplication guidance this is scope that should either be dropped from the PR
or split out together with a consumer, so it doesn't accrete as unowned surface area.

docs/skills/make-pr-ready.md
```diff
@@ -16,12 +16,21 @@ Work through these in order. Do not stop until every item is complete.
```
alexanderlhicks, line 16: Out of scope for a Lemma-10 PR — I'd split it out.

But on the merits the change is an **improvement**, and I want to be explicit because it would be
easy to misread as the opposite: it *adds* canonical base-ref resolution (inspect the PR with
`gh pr view … --json baseRefName,headRefName`, detect an `origin` that is a personal fork whose
`main` lags, fetch the canonical base into a separate tracking ref) where `main` hard-codes
`origin/main`. That's the right fix for stacked and fork-based PRs, and it's the behaviour I'd want
kept.

Cross-PR note: #659 edits overlapping lines of this same file differently — it keeps the
`origin/main` hard-coding and adds a policy permitting deletion of untracked user Markdown, which
I've asked them to drop. Merging both will need manual resolution: keep **this** §0 rewrite, and at
most a corrected form of #659's addition.

ArkLib/Commitments/Functional/Hachi/hachi-overview.html
```diff
@@ -645,9 +645,9 @@ <h2>Beyond the current line</h2>
     ["RingSwitch","Rlin.lean","bridge","§4.3","3","wip",3],
     ["RingSwitch","Reduction.lean","subprotocol","§4.3 · Fig4 · L9","4","wip",1],
     ["","ZeroCheck.lean","umbrella","§4.3","—","umbrella",0],
-    ["ZeroCheck","Batch.lean","bridge","§4.3 · Eq22–23","5","wip",1],
-    ["ZeroCheck","Constraints.lean","infrastructure","§4.3 · Eq21–23","—","wip",12],
-    ["ZeroCheck","Reduction.lean","subprotocol","§4.3 · Fig5 · L10*","6","wip",1],
+    ["ZeroCheck","Batch.lean","bridge","§4.3 · Eq22–23","5","wip",0],
+    ["ZeroCheck","Constraints.lean","infrastructure","§4.3 · Eq21–23","—","wip",6],
+    ["ZeroCheck","Reduction.lean","subprotocol","§4.3 · Fig5 · L10*","6","proven",0],
```
alexanderlhicks: Narrow one. This flat table marks `ZeroCheck/Reduction.lean` as `status:"proven"`, but the legend at
line 427 defines that pill as "Sorry-free **and complete**". Sorry-free yes; complete no — it's a
repaired variant carrying a temporary semantic shortness assumption.

The row *description* at line 575 needs no change at all and is genuinely a model of honest status
reporting — "**Corrected Lemma 10**", "CWSS is sorry-free and axiom-clean at (ℓ,k)=(2,D)",
"**Temporary shortness assumption**: `relZeroCheck` carries a `liftShort` conjunct…". It's only the
one-word status in the summary table that overreaches. Maybe add a `repaired` pill, or move the row
to `wip`.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
ErVinuelas: Those definitions are part of what we consider the "sumcheck part" of the protocol. I am working on it in a separate branch because I thought it made more sense to separate the different subprotcols in different PRs. Since we got the corrections regarding the Lemma 10, I will try to incorporate them before doing the sumcheck part. I think this definitions should be left empty as part of the skeleton of the protocol until I tackle that part. 

Does it make sense or should I fill the definitions so they are not hanging?

ArkLib/ToCompPoly/Multivariate/Restrict.lean
ErVinuelas: This was an old module that was used by the previous representation of some polynomials. It led to a small PR in CompPoly, but I deleted the parts that are no longer neccesary. Thank you!

docs/kb/audits/noz26-zero-check-lemma10.md
```diff
@@ -0,0 +1,265 @@
...
+
+Because `hμn` pins `2^{m₀} ≥ (μ + n)·deg φ`, the branching arity of the extraction tree grows from
+the paper's `O(d + b)` to `O(μ·d)`. This is the obvious objection to the repair and the answer is
+favourable. By [NOZ26] Lemma 4 the knowledge error of a coordinate-wise special-sound family is
+`ℓ·k/|S|^ℓ`, so at `(ℓ, k) = (2, D)` over `S = F_{q^k}` it is `2·D/|F_{q^k}|²`, i.e. roughly
+`2(μ + n)d/|F_{q^k}|²` in place of `2·max(2d, 2b−1)/|F_{q^k}|²`. That is still negligible at Hachi's
+field size, and `D` is polynomial in the witness dimensions so the transcript tree stays polynomial.
+What the repair buys in exchange is a *deterministic* identity equivalence
+(`multilinear_eq_zero_of_kronecker_roots`), strictly stronger than the Schwartz–Zippel bound Lemma 10
+actually needed. Recorded in the `zeroCheckD` docstring.
```
alexanderlhicks: Small suggestion, tweak accordingly:

```suggestion
Because `hμn` pins `2^{m₀} ≥ (μ + n)·deg φ`, the branching arity of the extraction tree grows from
the paper's `O(d + b)` to `O(μ·d)` — and, more importantly, the soundness error degrades. For a
nonzero `H₀` the univariate pullback `powAlgHom H₀` has degree `≤ 2^{m₀} − 1`, so a seed lands on a
root with probability up to `2^{m₀}/|F_{q^k}|`. At the paper's Figure 9 parameters (`q ≈ 2^32`,
`k = 4` so `|F_{q^k}| ≈ 2^128`; `deg φ = 2^10` and `w̃` length `≤ 2^26`, hence `m₀ ≈ 26`, which is
also what `hμn` pins) that is `≈ 2^-102`, against `≈ 2^-123` for Figure 5's uniform `τ₀` by
Schwartz–Zippel — a **~21-bit regression** on a `λ = 128` target.

Buying those bits back is not cheap. `k` must divide `d/2 = 512` ([NOZ26] Lemma 1 / Theorem 1), so
`k` is a power of two: `k = 5` would suffice numerically but is illegal, and the next legal value is
`k = 8`. The paper's sumcheck cost `26·k·32·(16+2)` bits `≈ 7.3 KB` at `k = 4` then becomes
`≈ 14.6 KB`, i.e. **double**, plus `F_{q^8}` arithmetic throughout.

Two accounting notes, since both are easy to get wrong:

- [NOZ26] Lemma 4's `ℓ·k/|S|^ℓ` must **not** be read as `2·D/|F_{q^k}|²` here. `H₀` depends only on
  `ρ₀` and `H_α` only on `ρ_α`, so a cheating prover needs one seed bad, not both; the `|S|^ℓ`
  denominator would price in an independence the protocol does not have. A knowledge error below the
  direct `2^{m₀}/|F_{q^k}|` bound is in any case impossible, since knowledge error dominates the
  success probability of a witness-less prover.
- The comparison baseline is Figure 5's Schwartz–Zippel error `m₀/|F_{q^k}|`, not
  `2·max(2d, 2b−1)/|F_{q^k}|²`; the latter mixes the paper's `D` with a denominator that does not
  apply.

No bound better than `2^{m₀}/|F_{q^k}|` is currently proved for this reparametrisation, and whether
a *realisable* table attains it is open: `exists_nonzero_multilinear_vanishing_on_kronecker_seeds`
gives tightness for arbitrary multilinears, but a protocol adversary must exhibit an `H₀` of the
form `∑ᵢ eq̃(t,i)·P_b(w̃(i))`, which that lemma does not construct.

What the repair buys in exchange is a *deterministic* identity equivalence
(`multilinear_eq_zero_of_kronecker_roots`), which is what the printed Lemma 10's root-counting step
needed and did not have. Whether that trade is the right one is a protocol question for the paper's
authors rather than a formalization question, and it is **not settled here**.

### Alternative repair routes under discussion (not yet claims)

The axis-cross gap was raised with the [NOZ26] authors directly, and this subsection records the
state of that correspondence. It is **not** a set of established results: nothing below is
formalized in ArkLib, and the statements are the authors' and our suggestions, not theorems.

In a reply of 2026-07-31, George O'Rourke confirmed the diagnosis — that Schwartz–Zippel cannot be
invoked under CWSS, because the CWSS tree constrains only the coordinate-wise structure of the
challenges and says nothing about their distribution — and also noted that the printed analysis
takes `(τ₀, τ₁)` as a two-coordinate vector even though the two coordinates are drawn from
`F_{q^k}^{log μ + log d}` and `F_{q^k}^{log n}`, neither of which is `F_{q^k}`. Two alternatives
were proposed there, both of which leave Figure 5's protocol unchanged:

1. drop CWSS for this step and argue knowledge soundness directly by Schwartz–Zippel, at the cost
   of no longer being able to compose the chain through a single generic CWSS-to-knowledge-soundness
   step;
2. treat each coordinate of `(τ₀, τ₁)` as a separate challenge round and prove plain
   `(k, …, k)`-special soundness by root counting with induction on the number of variables.

Route 2 looks the more promising one for this formalization, since it would keep the error at
Figure 5's order (`≈ 2^-122` by [NOZ26] Lemma 4 at `ℓᵢ = 1`, `kᵢ = 2`, summed over the
`log(μ + n) + log d + log n` coordinates) while still landing on a CWSS-shaped notion. Two caveats
before anyone relies on that: the required multivariate root-counting lemma is **not** in the tree,
and the relevant per-coordinate degree is `1` (both `H₀` and `H_α` are multilinear *in the
challenge*; the `2b − 1` and `2` appearing in the paper are witness-side sumcheck degrees), so the
per-coordinate parameter would be `k = 2` rather than the larger values a first reading suggests.

Until the authors settle the route, treat the Kronecker reparametrisation in this file as
provisional. Also recorded in the `zeroCheckD` docstring.
```

The arithmetic here comes out ~21 bits too optimistic, and I think it flips the conclusion.

`ℓ·k/|S|^ℓ` can't be read as `2·D/|F_{q^k}|²` in this protocol: `H₀` depends only on `ρ₀` and `H_α`
only on `ρ_α`, so a cheating prover needs *one* seed bad, not both — the `|S|^ℓ` denominator prices
in an independence that isn't there.

A direct bound settles it either way. For nonzero `H₀`, `powAlgHom H₀` has degree `≤ 2^{m₀} − 1`, so
a seed hits a root with probability up to `2^{m₀}/|F_{q^k}|`; at Figure 9's parameters (`m₀ ≈ 26`,
`k = 4`, `|F_{q^k}| ≈ 2^128`) that's `≈ 2^-102`, against `≈ 2^-123` for Figure 5's uniform `τ₀` by
Schwartz–Zippel. Knowledge error dominates a witness-less prover's success probability, so `2^-229`
isn't reachable. The baseline in the current text is off too — Figure 5's error is `m₀/|F_{q^k}|`,
not `2·max(2d, 2b−1)/|F_{q^k}|²`.

Worth recording that it's awkward to buy back: `k | d/2 = 512`, so `k` is a power of two — `k = 5`
would do numerically but is illegal, and `k = 8` doubles the sumcheck proof (`≈ 7.3 → 14.6 KB`).

To be clear about what I'm *not* claiming: `2^-102` is an upper bound, and
`exists_nonzero_multilinear_vanishing_on_kronecker_seeds` shows it's tight for arbitrary
multilinears — but a protocol adversary needs an `H₀` of the form `∑ᵢ eq̃(t,i)·P_b(w̃(i))`, which that
lemma doesn't construct. So the truth is in `[?, 2^-102]`. The reason I'd still change the text is
that nothing proves better than `2^{m₀}/|F_{q^k}|` here while Figure 5 proves `m₀/|F_{q^k}|`, and
this page is what the next person reads before setting parameters.

I built and validated the replacement below on top of `512a4179`: `lake build` green (4135 jobs),
`check-docs-integrity.py` and `scripts/kb/lint.py` pass, no new lines over 100 chars, and
`#print axioms` unchanged on the eleven core declarations — it's comment text only, no statement,
term or proof touched.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Constraints.lean
```diff
@@ -1,244 +1,851 @@
...
+Since the arity pin `hμn` gives `2^{m₀} ≥ (μ + n)·deg φ`, this raises the branching arity of the
+extraction tree from the paper's `O(d + b)` to `O(μ·d)`. That is the obvious objection to the
+Kronecker repair, and the answer is favourable. By [NOZ26] Lemma 4 the knowledge error of a
+coordinate-wise special-sound family is `ℓ·k/|S|^ℓ`, so at `(ℓ, k) = (2, D)` over `S = F_{q^k}` it
+becomes `2·D/|F_{q^k}|²`, i.e. roughly `2(μ + n)d/|F_{q^k}|²` in place of the paper's
+`2·max(2d, 2b−1)/|F_{q^k}|²`. At Hachi's field size that is still negligible, and since `D` is
+polynomial in the witness dimensions the transcript tree stays polynomial. The repair therefore
+buys a *deterministic* identity equivalence (`multilinear_eq_zero_of_kronecker_roots`, strictly
+stronger than a Schwartz–Zippel bound) at the price of a constant-factor-in-the-exponent-free
+arity increase.
```
alexanderlhicks: Small suggestion, tweak accordingly — same correction as on `docs/kb/audits/noz26-zero-check-lemma10.md`, since this docstring is where that note points with "Recorded in the `zeroCheckD` docstring". Kept terse here and left the detail (including the alternative repair routes raised with the paper authors, marked as suggestions rather than results) in the KB note:

```suggestion
Since the arity pin `hμn` gives `2^{m₀} ≥ (μ + n)·deg φ`, this raises the branching arity of the
extraction tree from the paper's `O(d + b)` to `O(μ·d)` — and, more importantly, it degrades the
soundness error. For a nonzero `H₀` the univariate pullback `powAlgHom H₀` has degree
`≤ 2^{m₀} − 1`, so a seed lands on a root with probability up to `2^{m₀}/|F_{q^k}|`. At the paper's
Figure 9 parameters (`m₀ ≈ 26`, `k = 4`, `|F_{q^k}| ≈ 2^128`) that is `≈ 2^-102`, against
`≈ 2^-123` for Figure 5's uniform `τ₀` by Schwartz–Zippel. Buying the ~21 bits back is not cheap:
`k` must divide `d/2 = 512`, so `k` is a power of two and `k = 4 → 8`, doubling the sumcheck proof
(`≈ 7.3 KB → ≈ 14.6 KB`).

[NOZ26] Lemma 4's `ℓ·k/|S|^ℓ` must **not** be read as `2·D/|F_{q^k}|²` here: `H₀` depends only on
`ρ₀` and `H_α` only on `ρ_α`, so a cheating prover needs one seed bad, not both, and the `|S|^ℓ`
denominator would price in an independence the protocol does not have.

No bound better than `2^{m₀}/|F_{q^k}|` is currently proved for this reparametrisation. Whether a
*realisable* table attains it is open: `exists_nonzero_multilinear_vanishing_on_kronecker_seeds`
gives tightness for arbitrary multilinears, but a protocol adversary must exhibit an `H₀` of the
form `∑ᵢ eq̃(t,i)·P_b(w̃(i))`, which that lemma does not construct.

What the repair buys in exchange is a *deterministic* identity equivalence
(`multilinear_eq_zero_of_kronecker_roots`), which is what the printed Lemma 10's root-counting step
needed and did not have. Whether that trade is the right one is a protocol question for the paper's
authors and is not settled here; see `docs/kb/audits/noz26-zero-check-lemma10.md`.
```

alexanderlhicks (commented):
*Follow-up to my cross-PR review, from a pre-merge pass at `512a4179`.*

Two small suggestions, both committable — same fix in two places, since the cost claim appears in
this file's `zeroCheckD` docstring as well as in the audit note it defers to.

Everything else I re-checked at this head looks good and I don't want it lost in the noise: the
Eq. (22) work is **faithful** (I checked `mAlphaTilde` case-by-case and `alphaContract`
term-for-term against the paper, not just the axiom sets — `−(α^d+1)` is correctly generalised to an
arbitrary cyclotomic modulus, and the reordered branches are provably disjoint), it's genuinely wired
in rather than proved-and-parked (`hAlpha_eq_zero_iff_alphaDefect` has 7 call sites), **`wTable` is
now concrete** where it used to be a sorried `def`, and the build is green with the `sorry` delta
still at −8. The new "second defect in the printed Lemma 10" subsection and the `1_{≤μ}` write-up are
both accurate. Nice work on the turnaround.

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Reduction.lean
```diff
@@ -62,217 +101,543 @@ import ArkLib.Commitments.Functional.Hachi.ZeroCheck.Batch
...
 
-/-- **The corrected Lemma 10 extraction algorithm (skeleton).**
-
-**Sorried** — this def is the extraction *algorithm* itself (the case split of the proof plan on
-`zeroCheck_coordinateWiseSpecialSoundWithEscape`). -/
-noncomputable def zeroCheckExtractor
+open Classical in
+/-- Classically select an output-relation witness for a leaf, with an arbitrary nonempty fallback
+outside accepting trees. -/
+noncomputable def nestedPathResponse
```
tobias-rothmann, line 423: Can you make this extractor definition computable?

ArkLib/Commitments/Functional/Hachi/ZeroCheck/Reduction.lean
```diff
@@ -62,217 +101,543 @@ import ArkLib.Commitments.Functional.Hachi.ZeroCheck.Batch
...
+  | _, .chalNode _ _ _ children => .chal 0 (nestedLeftPath (children 0))
+
+/-- Tree-based extractor for the scalar-round zero-check: the opening selected at the all-left
+leaf.
+
+There is no case analysis to do, exactly as in `CommittedScalar.mkWitness`. On trees whose leaves
+do **not** share an opening the escape event `nestedZeroCheckEsc` fires and the certificate's left
+disjunct carries the conclusion, so the choice of leaf is soundness-irrelevant; on trees that do
+share one, every leaf gives the same answer. -/
+noncomputable def nestedZeroCheckExtractor
```
tobias-rothmann, line 454: Same here, but that should follow from the previous def ig

tobias-rothmann (approved):
Soft approval from my side! 
The only thing is that the extractor(s) is(are) non-computable. 
If you can change that in this PR already great, otherwise I guess I'll have to change it in my PR #697

alexanderlhicks (approved):

alexanderlhicks:
> Soft approval from my side! The only thing is that the extractor(s) is(are) non-computable. If you can change that in this PR already great, otherwise I guess I'll have to change it in my PR #697

easier to do this in the #697  to ensure consistency in how it's done imo!

PR #603, opened 2026-07-02
Block Relative distance refactor + proofs

ElijahVlasov:
Improvements to Block Relative distance stuff + proof of the lemma 4.19 from the WHIR paper.

PR #534, opened 2026-06-05
Multilinear folding completeness

ElijahVlasov:
This PR introduces iterated folding operation as defined in the WHIR paper, proves it completeness, and its multilinear counterpart.

PR #657, opened 2026-07-17
Degree-2 folding that avoids using `log`

ElijahVlasov:
The title. + A couple nice lemmas for domains.

### DIFF
--- a/ArkLib/Data/CodingTheory/ProximityGap/Folding.lean
+++ b/ArkLib/Data/CodingTheory/ProximityGap/Folding.lean
@@ -232,6 +232,62 @@ theorem foldWord_k_1 [NeZero n] {i : Fin (2 ^ (n - 1))} {α : F} :
     ((f i + f i') / 2) + α * ((f i - f i') / (2 * x)) := by
   simp [foldWord, foldValue_k_1]
 
+/-- An explicit formula for `foldWord` when `k = 1` that
+  does not use Lagrange interpolation and avoids using `log`. -/
+theorem foldWord_k_1_no_choice {i : Fin (2 ^ (n - 1))} {α : F}
+  {j j' : Fin (2 ^ n)} (hjj' : j ≠ j')
+  (hj : domain j ^ 2 = domain.subdomain 1 i) (hj' : domain j' ^ 2 = domain.subdomain 1 i) :
+  foldWord domain f 1 α i =
+    ((f j + f j') / 2) + α * ((f j - f j') / (2 * domain j)) := by
+  have hn : n ≠ 0 := fun contra ↦ by
+    aesop (add safe [cases Fin, (by omega)])
+  have : NeZero n := ⟨hn⟩
+  rw [foldWord, foldValue_k_1]
+  extract_lets x u u'
+  have hroots := square_roots_explicit
+    (i := 0) (ω := domain) (x := domain.subdomain 1 i) (y := domain j)
+    (by omega) (by simp) hj
+  have hj' : domain j' ∈ ({domain j, -domain j} : Finset _) := by
+    simp [←hroots]
+    aesop (add safe (by rw [mem_subdomain_0_iff_mem]))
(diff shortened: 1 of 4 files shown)


alexanderlhicks (commented):
> **AI-generated review.** This review was produced by Codex and has not been written by a human reviewer. Findings were validated at exact head `3d8d860b779bc210f2ed38f6b2ef9f5f3cfdaa8b` against the Lean proof, the WHIR fold definition, current ArkLib APIs, exact-head builds, and the related PR stack.

**Conclusion: the mathematical change is acceptable; the inline API comments are non-blocking cleanup.** The theorem gives a correct explicit-root form of the binary fold, handles either ordering of the two roots, derives nontriviality from distinct indices, and is a useful bridge for #534.

**Validation performed**

- `lake build ArkLib` completed all 4,109 jobs at this exact SHA. The validation wrapper's remaining failure is only the inherited missing `NOZ26` bibliography key. The currently red GitHub build/site checks are likewise inherited main-line validation/`SendClaim.completeness` issues, not Lean failures caused by this PR.
- `#print axioms ProximityGap.foldWord_k_1_no_choice` reported only standard Lean/Mathlib axioms and no `sorryAx`.
- Inspected the proof's two-root classification and both root-order cases against the binary formula in [WHIR Definition 4.14, p. 26](https://eprint.iacr.org/2024/1586).
- Built a synthetic current-main + #603 + #657 stack successfully (4,128 jobs). #534 overlaps the insertion point, so its iterated/MLE work should consume or explicitly bridge this theorem during conflict resolution.

This does not by itself close WHIR Claim 4.15 or STIR Lemma 4.9, but it is sound, compatible foundational progress toward the canonical fold interface.

ArkLib/Data/CodingTheory/ProximityGap/Folding.lean
```diff
@@ -232,6 +232,62 @@ theorem foldWord_k_1 [NeZero n] {i : Fin (2 ^ (n - 1))} {α : F} :
     ((f i + f i') / 2) + α * ((f i - f i') / (2 * x)) := by
   simp [foldWord, foldValue_k_1]
 
+/-- An explicit formula for `foldWord` when `k = 1` that
+  does not use Lagrange interpolation and avoids using `log`. -/
+theorem foldWord_k_1_no_choice {i : Fin (2 ^ (n - 1))} {α : F}
```
alexanderlhicks: **AI-generated inline feedback — naming/integration (non-blocking).** The proof is correct, but `no_choice` is misleading because callers still choose and supply `j` and `j'`. Please consider `foldWord_k_1_of_square_roots`. I validated both proof branches: swapping the two roots negates both the numerator and denominator of the odd term, so the expression is order-invariant. Document that fact and make #534's iterated/MLE development use this theorem or an explicit equivalence, avoiding two unconnected binary-fold APIs.

ArkLib/Data/Domain/CosetFftDomain/Ops.lean
```diff
@@ -116,6 +116,25 @@ lemma neg_mem_domain_iff_mem [nz : NeZero n] :
 lemma domain_implies_char_ne_2 [NeZero n] (ω : D) :
   ¬CharP F 2 := FftDomainClass.domain_implies_char_ne_2 (toFftDomain ω)
 
+@[simp]
```
alexanderlhicks: **AI-generated inline feedback — simp API (non-blocking).** The explicit witness `ω` does not occur in the conclusion `(2 : F) ≠ 0`, so `simp` cannot infer it from a matching goal. I checked current call sites and the neighboring `domain_implies_char_ne_2`: please remove `[simp]` here, or expose an inferable/scoped characteristic fact whose domain witness is available from context. This avoids registering a globally surprising rule that cannot normally fire.

alexanderlhicks (approved):

alexanderlhicks (approved):

PR #659, opened 2026-07-20
refactor(Hachi/CWSS): cwss packages with escape

tobias-rothmann:
# CWSS packages with escape threading

Adds a **protocol-agnostic escape abstraction** to the coordinate-wise special-soundness
(CWSS) composition machinery, and refactors the package layer into a clean **2×2 lattice**
of package kinds joined by a single universal append `▷`.

## Motivation: escapes belong to composition, not to relations

In a composed special-soundness chain, a downstream extractor may fail to produce a *real*
witness and instead return a cryptographic **escape** — e.g. a commitment binding-break that
is a Module-SIS solution via weak binding (Hachi's `w̃`-commitment, [NOZ26] Remark 2 /
Lemma 7). Composed extraction feeds each extractor's output into the *previous* seam relation,
so every relation upstream of the escape's origin needs somewhere to put it.

The naive fix — bake an escape disjunct into each subprotocol's own `relIn`/`relOut` — pollutes
every upstream relation definition with cryptographic plumbing it shouldn't know about.

**This PR keeps the subprotocol relations ordinary and threads the escape on a parallel seam
that is only widened *into* the relation inside the special-soundness certificate.** Escape
handling shifts from *relation definitions* to *appending / protocol composition*. Because an
escape (an MSIS/collision solution) is checkable against the commitment key alone, it is
**statement-independent**: it passes through statement maps trivially, and the escape branch of
every seam extractor is the identity `Sum.inr`.

## Core primitive: `Set.withEscape`

```
withEscape (rel : Set (S × W)) (esc : Set E) : Set (S × (W ⊕ E))
```

Widens an ordinary relation to accept escapes on the right summand (`+` lemmas:
`mem_withEscape_inl/inr`, `mem_withEscape_language_iff`, `withEscape_empty_iff`). The widening
lives *inside* each package's certificate — the public `relIn`/`relOut` stay escape-free.

## The four package kinds — the 2×2 lattice (escape? × guarded?)

Every kind carries `verifier`, `struct`, `relIn`, `relOut`, and an `isCWSS` certificate; the
escape kinds add `escIn`/`escOut` plus `escape_mono : escOut ⊆ escIn` (escapes may grow up the
chain but are never discarded).

| Package                | Escape-aware | Guarded (verifier may reject at runtime) | Certificate is over…                                | Native infix | File            |
|------------------------|:------------:|:-------------------------------:|-----------------------------------------------------|:------------:|-----------------|
| `CWSSPackage`          | –            | – (pure)                        | `relIn` / `relOut`                                   | `▷`          | `Package.lean`  |
| `GCWSSPackage`         | –            | ✓                               | `relIn` / `relOut`                                   | `▷ᵍ`         | `Guarded.lean`  |
| `EscapeCWSSPackage`    | ✓            | – (pure)                        | `relIn.withEscape escIn` / `relOut.withEscape escOut`| `▷ₑ`         | `Escape.lean`   |
| `EscapeGCWSSPackage`   | ✓            | ✓                               | `relIn.withEscape escIn` / `relOut.withEscape escOut`| `▷ₑᵍ`        | `Escape.lean`   |

The *guarded* axis (from the sumcheck track) models a left factor whose verifier may **reject**:
`if check stmt tr then pure (out …) else failure`. This is needed exactly where a runtime check
reads sent/input data that the downstream statement type drops (sumcheck round targets,
final-eval check, recursion handoff), which pure/relation-only verifiers cannot express.

## Lifting: an escape-free / pure package enters a richer chain for free

Escape packages need only be **defined** for the subprotocols that genuinely produce escapes;
everything else is lifted automatically at composition time.

| Lift                                         | From → To                                             | Semantics                                                        |
|----------------------------------------------|-------------------------------------------------------|-----------------------------------------------------------------|
| `CWSSPackage.toGuarded` / `EscapeCWSSPackage.toGuarded` | pure → guarded (both escape columns)       | lossless — purity is the `check := true` special case            |
| `CWSSPackage.withEscape` / `GCWSSPackage.withEscape`    | escape-free → escape-aware (both rows)     | constant budget `escIn = escOut` (passes downstream escapes through unchanged) |
| `coordinateWiseSpecialSound.withEscape`      | certificate-level widening to any escape set `E`      | underpins the package lifts; `treeSpecialSound.withEscape` at the tree level |

## Composition: one universal `▷` over all 16 pairings

`L₁ ▷ L₂` composes **any** two package kinds at the *join* of their kinds, inserting the
`withEscape` / `toGuarded` lifts on the fly and choosing escape budgets so the escape seam holds
definitionally (only the ordinary relation seam `hRel`, and between two escape-aware factors the
escape seam `hEsc`, remain — both discharged by `rfl`). Join semantics: escape-free ▷ escape-free
stays escape-free, pure ▷ pure stays pure, and a single escape-aware or guarded factor lifts the
rest of the chain.

- All **16** (4×4) appends are implemented (`append`, `appendEscape`, `appendGuarded`,
  `appendEscapeGuarded`, `appendPure`, …) and wired into a single dispatch table `univAppendFn`.
- `▷` is a **single scoped elaborator** that reads each factor's head constant and applies the
  unique compatible append — **linear** in chain length. Overloaded-notation `choice` nodes would
  re-elaborate nested alternatives exponentially (a 5-link Hachi chain already exhausts the
  heartbeat budget). The kind-marked `▷ₑ`, `▷ᵍ`, `▷ₑᵍ` remain as explicit single-target synonyms;
  repo code composes with the universal `▷` throughout.


ArkLib/Commitments/Functional/Hachi/QuadEval/Reduction.lean
```diff
@@ -100,20 +100,34 @@ instance : Nonempty
...
+  ⟨{ message := fun _ _ => 0, innerDecomp := fun _ _ => 0, challenge := fun _ => 0 }⟩
+
+/-- A concrete local escape produced by the three-case extractor of Hachi Lemma 8. The public
+matrix is carried with the kernel vector, making validity statement-independent once the escape
+has been produced. This is essential for threading the escape backwards through earlier protocol
+relations without attaching the originating `QuadEvalStatement`. -/
+inductive QuadEvalSISBreak (Φ : CyclotomicModulus R)
+    (innerRows messageDigits outerRows blocks innerDigits dRows : Nat) where
+  /-- A short nonzero kernel vector for the outer commitment matrix `B`. -/
+  | msisB
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[blocker · cause]** The escape object now carries its **own** public matrix as a constructor field. On `main`, `msisB` carried only the `solution` and validity was checked against the *statement's* commitment key (`stmt.pp.outerMatrix`). Carrying the matrix here is what lets the escape be validated against an arbitrary matrix rather than the real key — see the linked comments on `quadEvalSISSet` and `eval_coordinateWiseSpecialSound`.

_Suggested fix:_ drop the `matrix` field and check the solution against the statement's `outerMatrix`/`dMatrix` at the use site (restore main's shape), or pass the statement matrices into `quadEvalSISSet`.

ArkLib/Commitments/Functional/Hachi/QuadEval/Reduction.lean
```diff
@@ -184,6 +198,45 @@ def evalConsistency (base : ZMod q) (a : PolyVec (Rq Φ) (2 ^ m)) (b : PolyVec (
 def dShort (γ : ℕ) : ModuleSIS.Solution Φ (blocks * messageDigits) → Bool :=
   fun z => decide (vecLInftyNorm Φ z ≤ subLInftyNormBound γ)
 
+/-- The statement-independent set of valid Module-SIS escapes that `QuadEval` may add while
+extracting. Each escape carries the matrix against which its solution is checked: `B` for a
+divergent inner decomposition, or `D` for a divergent carrier decomposition. -/
+def quadEvalSISSet (γ : ℕ) :
+    Set (QuadEvalSISBreak Φ innerRows messageDigits outerRows blocks innerDigits dRows) :=
+  { e | match e with
+    | .msisB matrix solution =>
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[blocker · mechanism]** This validates the solution against the escape's *carried* `matrix`, not the statement's commitment key. Since `ModuleSIS.relation isShort A z = (z ≠ 0) && isShort z && (A *ᵥ z = 0)` and `Simple.PublicParams = PolyMatrix`, choosing `matrix := 0` makes `A *ᵥ z = 0` hold for free, so **any** short nonzero `z` inhabits `quadEvalSISSet γ`. Confirmed with a compiling proof against this branch:

```lean
example (γ z) (hz : z ≠ 0) (hs : outerShort Φ γ z = true) :
    QuadEvalSISBreak.msisB 0 z ∈ quadEvalSISSet Φ γ := by
  show ModuleSIS.relation Φ (outerShort Φ γ) 0 z = true
  simp only [ModuleSIS.relation, Bool.and_eq_true, decide_eq_true_eq]
  refine ⟨⟨hz, hs⟩, ?_⟩; funext i; simp [matVecMul_apply, ArkLib.Lattices.dot]
```

So `quadEvalSISSet γ` is nonempty for every `γ ≥ 1`. This is the composition-security gap: Remark 2 / Lemma 7 require an MSIS solution for the *actual* `[A|B]`, checkable against the commitment key alone — an escape checkable against a self-carried (possibly zero) matrix is free.

_Suggested fix:_ `ModuleSIS.relation Φ (outerShort Φ γ) stmt.pp.outerMatrix solution` (and the `dMatrix` analogue for `msisD`), as on `main`.

ArkLib/Commitments/Functional/Hachi/Composition.lean
```diff
@@ -130,51 +135,58 @@ section Composition
...
-Figure 3], `Rq`-level), `sorry`-free.** This is the certificate carried by `evalChain`: the composed
-verifier (bridge ⧺ `QuadEval`) is CWSS for `ofIsEmpty.append foldStructure`, reducing the
-polynomial-level `relPolyEval` (a weak eval-consistent opening, or MSIS(B), or MSIS(D)) to Hachi
-Eq. (20) (`relOut`), pinned to `𝓜(q, α)` with the [LS18] hypotheses of
+Figure 3], `Rq`-level), `sorry`-free.** The public flow of `evalChain` is the ordinary
+`relPolyEval → relIn → relOut` flow. Only this certificate widens its endpoints by the
+parallel escape sets; `QuadEval` adds its mapped local Module-SIS(B/D) outcomes to the input
+budget. The result is pinned to `𝓜(q, α)` with the [LS18] hypotheses of
 `quadEval_coordinateWiseSpecialSound`. -/
 theorem eval_coordinateWiseSpecialSound (init : ProbComp σ)
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[blocker · consequence]** Because the escape set (`esc ∪ Q.localEsc`) is unconditionally nonempty — every `QuadEvalEscapeMap` has `mapsTo (quadEvalSISSet γ) localEsc`, and `quadEvalSISSet γ` is nonempty for `γ ≥ 1` — this CWSS statement (and `hachi_iteration_coordinateWiseSpecialSound` below) is **vacuously true**. `coordinateWiseSpecialSound` is existential in the extractor, and `treeSpecialSound.withEscape`'s nonempty-escape branch (`CWSS/.../Escape.lean:144`) discharges it with the *constant* extractor `fun _ _ => Sum.inr e`, which ignores the accepting transcript entirely. So the theorem holds for any verifier and certifies nothing about extraction.

This is **not** a proof bug: the theorem is genuinely proven and axiom-clean (`#print axioms` → `propext, Classical.choice, Quot.sound`), and the *honest* extractor `buildWitness` is faithful. The problem is only that the theorem's target set is wider than the honest extractor needs, and that wider set is free — so the existential collapses. Fixing the `quadEvalSISSet` escape (see the two linked comments) restores content; no change is needed on this line.

alexanderlhicks (commented):

### Additional notes (lines outside the diff hunks)

- **`ArkLib/Commitments/Functional/Hachi/QuadEval/Soundness.lean:95`** — **[documented deviation · no action required]** For the record (already noted in the docstrings): `quadEvalBetaSq` is a squared-ℓ₂ over-approximation with dimensional factors, not the paper's `β̄ = 2b^k` (Lemma 8). This is the same `quadEvalBetaSq`/`VerifiedOpening` vs `(2b^k, 2ω, b)` parameter bridge owed since #626 — worth tracking as the parameter-instantiation task so the escape norm bounds eventually reconcile to the paper's constants.

ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/Escape.lean
```diff
@@ -75,3 +100,502 @@ theorem withEscape_empty_iff (rel : Set (S × W)) (s : S) (w : W ⊕ Empty) :
...
+  /-- The ordinary input relation. -/
+  relIn : Set (StmtIn × WitIn)
+  /-- The ordinary output relation. -/
+  relOut : Set (StmtOut × WitOut)
+  /-- Escapes that may be returned when extracting an input witness. -/
+  escIn : Set E
+  /-- Escapes accepted from the next extractor in the chain. -/
+  escOut : Set E
+  /-- Extraction may preserve or grow the escape set, but never discard an accepted escape. -/
+  escape_mono : escOut ⊆ escIn
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[design]** The escape-package abstraction constrains only monotonicity of the escape set — nothing requires the escape to be a *statement-tied hardness* predicate (an MSIS break of a fixed commitment key). That is what lets a downstream instance (QuadEval) supply a free-standing escape set and silently produce a vacuous certificate. Consider adding an invariant to the package contract — e.g. carry the commitment key as a package parameter and require escapes to be MSIS solutions against *it* — so composing escape packages can't yield free escapes. Without it this footgun is available to every future escape instance, not just Hachi.

ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/Escape.lean
```diff
@@ -75,3 +100,502 @@ theorem withEscape_empty_iff (rel : Set (S × W)) (s : S) (w : W ⊕ Empty) :
...
+    [∀ i, SampleableType (pSpec₁.Challenge i)]
+    (L₁ : EscapeGCWSSPackage init impl E StmtA WitA StmtB WitB pSpec₁)
+    (L₂ : EscapeCWSSPackage init impl E StmtB WitB StmtC WitC pSpec₂)
+    (hRel : L₁.relOut = L₂.relIn := by rfl)
+    (hEsc : L₁.escOut = L₂.escIn := by rfl) :
+    EscapeGCWSSPackage init impl E StmtA WitA StmtC WitC (pSpec₁ ++ₚ pSpec₂) :=
+  L₁.append L₂.toGuarded hRel hEsc
+
+@[inherit_doc EscapeGCWSSPackage.append]
+scoped infixr:65 " ▷ᵍ " => EscapeGCWSSPackage.append
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[hygiene]** `▷ᵍ` now has four overloaded declarations (here, :530, :533, and `Guarded.lean:219`) — exactly the overloaded-notation `choice`-node pattern the universal `▷` elaborator was introduced to avoid. Harmless in practice because repo code uses `▷`, but it undercuts the stated motivation. Suggest dropping the kind-marked `▷ᵍ` synonyms (keep only `▷`) or reducing `▷ᵍ` to a single target.

ArkLib/Commitments/Functional/Hachi/QuadEval/Bridge.lean
```diff
@@ -161,43 +161,35 @@ omit [NeZero q] in
...
-  | msisD z => exact h
+  obtain ⟨hvo, hec⟩ := h
+  refine ⟨hvo, ?_⟩
+  change CMlPolynomial.eval (Hachi.toPolynomial (derivedMsgMatrix Φ base w)) (s.xl ++ s.xh)
+    = s.y
+  rw [← Hachi.splitForm_monomialBasis_eq_eval (derivedMsgMatrix Φ base w) s.xl s.xh]
+  exact hec
 
 omit [NeZero q] in
 /-- **CWSS of the bridge.** The zero-round `ReduceClaim` head is coordinate-wise special sound for
```
alexanderlhicks: 🤖 _AI-generated (Claude Code)_

**[dead code]** This theorem now has no consumers (grep across `ArkLib/` finds no use outside this file); `bridgePackage.isCWSS` inlines a widened variant instead. Either wire it in or remove it, so the file doesn't accumulate an unused public API surface.

_(anchored near L200; the referenced line is just outside the diff hunk.)_

ArkLib/Commitments/Functional/Hachi/QuadEval/Reduction.lean
```diff
@@ -184,6 +206,54 @@ def evalConsistency (base : ZMod q) (a : PolyVec (Rq Φ) (2 ^ m)) (b : PolyVec (
...
+structure QuadEvalEscapeMap
+    (pp : Hachi.PublicParamsD Φ innerRows messageRows messageDigits outerRows blocks innerDigits
+      dRows) (γ : ℕ) (E : Type) where
+  /-- Ambient encodings of valid escapes introduced by this `QuadEval` invocation. -/
+  localEsc : Set E
+  /-- Embed a concrete `QuadEval` SIS break into the common escape type. -/
+  toEscape :
+    QuadEvalSISBreak Φ innerRows messageDigits outerRows blocks innerDigits dRows → E
+  /-- Valid local breaks land in `localEsc`. -/
+  mapsTo : Set.MapsTo toEscape (quadEvalSISSet Φ pp γ) localEsc
```
alexanderlhicks: `mapsTo` constrains `localEsc` only from *below*. Nothing bounds it above, so
`localEsc := Set.univ, toEscape := id` satisfies the structure and widens the headline certificate to
nothing.

But the problem is worse than `Set.univ`, and this is what matters for choosing a fix: **mere
nonemptiness of the escape set already collapses the certificate.** The following compiles with no
hypothesis about the verifier, the challenge structure, the tree, or acceptance (Lean even flags the
`SampleableType` instance as unused):

```lean
theorem cwss_trivial_of_escIn_nonempty
    (D : CWSSStructure pSpec) (relIn : Set (StmtIn × WitIn)) (relOut : Set (StmtOut × WitOut))
    (esc : Set E) (escOut : Set EOut) (he : esc.Nonempty)
    (verifier : Verifier oSpec StmtIn StmtOut pSpec) :
    verifier.coordinateWiseSpecialSound init impl D
      (relIn.withEscape esc) (relOut.withEscape escOut) := by
  obtain ⟨e, he⟩ := he
  exact ⟨fun _ _ => .inr e, fun _ _ _ _ => he⟩
```

and `Set.mem_withEscape_language_iff` degenerates the other half — with `esc` nonempty every
statement lies in the widened output language, so `IsAccepting` carries no information either.

So "expose the paper theorem via the canonical `.self` map" does **not** repair it; see the comment
on line 246. What has to change is the *quantifier*, not the set:
`buildWitness_mem_relIn_withEscape` already constrains the concrete extractor and is axiom-clean, so
make that the advertised entry point, and present the Module-SIS branch as "*this named algorithm
outputs an element of `quadEvalSISSet pp γ`*" rather than "an element exists". Tightening
`QuadEvalEscapeMap` to require exact-image membership is still worth doing for composition hygiene,
but on its own it restores no security content.

This is pre-existing framework behaviour (`Escape.lean` is on `main` and its docstring concedes "the
formal price"), so it needn't block the merge by itself — but it is much cheaper to fix now than
after rows 7-12 are built on this interface, and the same comment is on #654's `liftPackage`.

ArkLib/Commitments/Functional/Hachi/QuadEval/Reduction.lean
```diff
@@ -184,6 +206,54 @@ def evalConsistency (base : ZMod q) (a : PolyVec (Rq Φ) (2 ^ m)) (b : PolyVec (
...
+
+namespace QuadEvalEscapeMap
+
+/-- The canonical embedding when the ambient escape type is exactly the concrete `QuadEval` SIS
+break type. This is useful for running `QuadEval` by itself; composed Hachi chains instead provide
+an embedding into their shared escape type. -/
+def self
```
alexanderlhicks: `self` sets `localEsc := quadEvalSISSet Φ pp γ`, so it is **not** a repair for the previous comment —
it is the canonical *instance* of the problem.

The docstring at line 214 offers "for kernel-free keys the set is empty", but that escape hatch is
unavailable in the only regime the scheme is built for. For a **compressing** key the set is nonempty
by pigeonhole: `z ↦ A *ᵥ z` maps the coefficient box `{‖z‖∞ ≤ γ}` (size `(2γ+1)^(d·cols)`) into
`Rq^rows` (size `q^(d·rows)`), which is non-injective once `(2γ+1)^cols > q^rows`; the difference of
two colliding short vectors is a nonzero kernel element of `ℓ∞`-norm `≤ 2γ` — exactly an element of
`quadEvalSISSet` at `outerShort`/`dShort γ = 2γ`. Hachi's parameters (`cols = δ·2^m`, `rows = n_A`,
`γ = b`) sit far inside that regime, and a non-compressing commitment is useless.

Module-SIS is only *computationally* hard, while `withEscape` asks only for set membership. So the
disjunction being formalized reads "extract a witness **or** produce an object that exists
unconditionally". Worth adjusting the line-214 docstring too, since "kernel-free keys" suggests the
emptiness case is the normal one when it's the degenerate one.

ArkLib/Commitments/Functional/Hachi/QuadEval/Reduction.lean
```diff
@@ -184,6 +206,54 @@ def evalConsistency (base : ZMod q) (a : PolyVec (Rq Φ) (2 ^ m)) (b : PolyVec (
 def dShort (γ : ℕ) : ModuleSIS.Solution Φ (blocks * messageDigits) → Bool :=
   fun z => decide (vecLInftyNorm Φ z ≤ subLInftyNormBound γ)
 
+/-- The statement-independent set of valid Module-SIS escapes that `QuadEval` may add while
+extracting, **relative to the fixed commitment key `pp`**: solutions are checked against `pp`'s
+`B` matrix (a divergent inner decomposition) or its `D` matrix (a divergent carrier
+decomposition). Checkability against the key alone — never against statement data — is what
+makes the set a genuine hardness target: an element is a Module-SIS break of the *actual* key
+([NOZ26] Remark 2 / Lemma 7), and for kernel-free keys the set is empty. -/
+def quadEvalSISSet
```
alexanderlhicks, line 217: Worth saying explicitly: **the fixed-key repair here is real and is the right shape.** Solutions are
checked against `pp.outerMatrix` / `pp.dMatrix` — the actual key carried by the statement's public
parameters — so a break can no longer be smuggled in by supplying a convenient matrix alongside the
solution, which was the original defect.

I also checked the underlying relation for the obvious hole and it isn't there: `ModuleSIS.relation`
(`Data/Lattices/ModuleSIS.lean:85-89`) is `decide (z ≠ 0) && isShort z && decide (A *ᵥ z = 0)`, so
NOZ26 Definition 1's `0 < ‖z‖∞` really is enforced. That matters more than it looks — a missing
`z ≠ 0` would have turned the nonemptiness argument in my comment on line 246 from a
parameter-regime observation into an unconditional one.

ArkLib/Commitments/Functional/Hachi/Composition.lean
```diff
@@ -31,52 +33,54 @@ sibling `Commitment.lean`.)
...
-   loop's closing seam.
+   loop's closing seam. The universal `▷` lifts each pure factor into the escape-guarded world
+   automatically (`Escape.lean`, package-lattice section); no explicit `.toGuarded` calls.
 
 ## The composed verifier chain, seam by seam
 
 Top-to-bottom is one opening iteration of the ArkLib Hachi commitment, whose committed data is an
 `Rq`-valued multilinear polynomial. The opening starts with the Figure 3 path, not with §3: the §3
 packing head (extension-field claims into `Rq`-claims, via the generalized `RingSwitching` packing
 phase — see `HACHI_RING_SWITCHING_PLAN.md`, Phases B–E) is a separate track that wraps *external*
```
alexanderlhicks: This PR deletes six root design documents (4,096 lines) but **21 references to them survive** across
10 tracked files. Six are in this file alone: lines 53 and 388 (`HACHI_RING_SWITCHING_PLAN.md`), 87
(`HACHI_LEMMA10_GAP.md`), and 93, 278, 394 (`HACHI_RECURSION_GAP.md`).

Full list so it can be fixed in one pass:

| file | lines / count |
|---|---|
| `Hachi/Composition.lean` | 53, 87, 93, 278, 388, 394 |
| `Hachi/ZeroCheck/Constraints.lean` | 45, 75, 188 |
| `Hachi/ZeroCheck/Reduction.lean` | 3 refs |
| `Hachi/ZeroCheck/Basic.lean` | 1 |
| `Hachi/Recursion/ZBatchBridge.lean` | 2 |
| `Hachi/Recursion/PartialEval.lean` | 1 |
| `CWSS/ScalarRound.lean`, `CWSS/Guarded.lean` | 1 each |
| `docs/wiki/repo-map.md` | 124, 132 |
| `Hachi/hachi-overview.html` | 1 |

Some are **load-bearing section pointers**, not decoration: `ZeroCheck/Constraints.lean:45`, `:75`
and `:188` cite `HACHI_LEMMA10_GAP.md` **§3.K / §3.K.2** as the derivation of the Kronecker-seed
challenge design the code implements, and line 93 here points at `HACHI_RECURSION_GAP.md` to back
"All other sorries in the chain are honest skeleton work". Deleting the documents removes the only
written justification for designs the code depends on — and `check-docs-integrity.py` doesn't see
paths inside docstrings, so CI stays green either way.

**The fix is already written**: #654 deletes the same six files and leaves **zero** dangling
references (grep count 0 at its head). Please adopt its cleanup on rebase rather than redo it — and
migrate the surviving mathematics first (the axis-cross counterexample, the §3.K Kronecker
derivation, the recursion gap analysis) into `docs/kb/audits/`, since those are currently the only
detailed statements of several known unproved gaps in the repo.

alexanderlhicks (commented):
*Part of a cross-PR review of #654, #656 and #659 read together (plus the open ring-switching PR #615), in isolated builds, against NOZ26 (ePrint 2026/156, SHA-256 `c7c98591…afd5e6a15`). Where a finding interacts with a sibling PR I say so explicitly.*

## #659 — reviewed at `74feb760`

The refactor is clean and the **fixed-key Module-SIS repair is genuinely sound.** Escapes are now
checked against `pp.outerMatrix` / `pp.dMatrix`, the actual key from the statement's public
parameters, so a break can no longer be smuggled in by supplying a convenient matrix alongside the
solution. I also went looking for the obvious hole and it isn't there: `ModuleSIS.relation` is
`decide (z ≠ 0) && isShort z && decide (A *ᵥ z = 0)`, so Definition 1's `0 < ‖z‖∞` really is
enforced.

Several paper checks came out right on independent re-derivation:

- `paperRelOut_subset_relOut` has the containment in the **sound** direction — the formalized
  relation is the weaker (larger) one, so soundness against it is at least as strong;
- the `ℓ₁` invertibility criterion implies paper Lemma 3's `ℓ₂` hypothesis, since `‖·‖₂ ≤ ‖·‖₁`. A
  sound generalisation, not a weakening;
- `SS(C, 2^r, 2)` really does have `ℓ(k−1)+1 = 2^r + 1` elements, as Lemma 8 states. (Pleasant
  contrast with Lemma 10, whose transcript count is wrong — see #656.)

The 14 Gadget Basic → Core declarations are body-identical moves, the five umbrella renames are
declaration-free, import integrity holds, and there is **no new `sorry`** (changed-file bodies
36 → 35). `quadEval_coordinateWiseSpecialSound`, `paperRelOut_subset_relOut`, `CWSSPackage.append`
and `EscapeCWSSPackage.append` are all axiom-clean.

Three things I'd like changed:

1. **Six deleted design documents with 21 live references left behind**, across 10 files. Several are
   load-bearing section pointers — `ZeroCheck/Constraints.lean:45,75,188` cite
   `HACHI_LEMMA10_GAP.md` **§3.K / §3.K.2** as the derivation of the Kronecker-seed design the code
   implements. `check-docs-integrity.py` doesn't see docstring paths, so CI stays green. **#654
   performs the identical deletion with zero dangling references** — adopt its cleanup on rebase, and
   migrate the surviving gap analyses into `docs/kb/audits/` first.
2. **`docs/skills/make-pr-ready.md` tells agents to delete untracked user files** after only a
   warning. Out of scope, and hazardous in a tree that routinely carries dozens of untracked planning
   documents. (#656 edits the same file with a genuinely useful canonical-base fix — worth keeping
   theirs and dropping this addition.)
3. **`QuadEvalEscapeMap` has no upper bound on `localEsc`** — and the obvious fix doesn't work,
   because `.self` sets `localEsc := quadEvalSISSet`, which is nonempty for every *compressing* key,
   and nonemptiness alone already collapses the certificate to a constant extractor. Compiled witness
   in that comment. Pre-existing framework behaviour, so not a blocker on its own, but the
   guarded-append and package-lattice claims should be scoped accordingly.

Two smaller items: the generated overview still advertises a superseded shape — "single line of nine
reductions" (:357), "9-link opening chain" (:365), "≈24 open sorries" (:367), and generation "from
the current working tree on the `hachi-skeleton` branch" (:527) — against an actual 12 rows and **33**
executable `sorry` bodies in the Hachi tree. Those lines fall outside this PR's diff hunks so I
couldn't anchor a comment; either regenerate or drop the numeric claims. And your HMZ25 bib entry is
the **correct** one of the two competing additions — see that comment, it's a heads-up rather than a
defect.

**On #615** (open, head `68e9a752`): it overlaps this PR on `ArkLib.lean`,
`blueprint/src/references.bib` and `docs/wiki/repo-map.md`. Worth noting it is *not* exposed to
item 3: #615 states security as an RBR knowledge-soundness probability bound and never uses
`withEscape`, so there's no constant-extractor escape hatch there. The fix I suggest above moves
these certificates toward the shape #615 already has.

**Ordering: #654 → #659 → #656.** Rebasing onto #654 gets you its proof of
`ScalarRound.coordinateWiseSpecialSound_of_mkWitness_scalar` (currently `sorry` here), its
`HACHI_*.md` reference cleanup, and a settled bib — three of the items above largely resolve
themselves that way.

docs/wiki/repo-map.md
```diff
@@ -101,40 +104,38 @@ home_page/            site assets and assembled website root
...
+    folder.
     (Distinct from the §3 packing reduction under `ProofSystem/RingSwitching/`, also a ring-switch.)
   - `ZeroCheck/` (§4.3, Figure 5 / **corrected** Lemma 10) — reduces the batched identities
     `H₀ ≡ 0 ∧ H_α ≡ 0` to random-point evaluations. `ZeroCheck/Constraints` is the **shared**
     encoding (Eqs. (21)–(23): the table `w̃`, `H₀`/`H_α`, the sumcheck polynomials, degree pins,
     the Kronecker curve `kroneckerPoint`, per-round seam `roundRel`), consumed by both this
     zero-check and `Sumcheck/`; `ZeroCheck/Batch` is the per-row/range ⇄ `H₀/H_α ≡ 0` batching
     bridge; `ZeroCheck/Reduction` is the corrected Lemma 10 (Kronecker seed pair, `(ℓ, k) = (2, D)`;
-    see `HACHI_LEMMA10_GAP.md`). `ZeroCheck.lean` re-exports the folder.
+    see `HACHI_LEMMA10_GAP.md`). `ZeroCheck/Basic.lean` re-exports the folder.
```
alexanderlhicks: Two more references to deleted documents here (lines 124 and 132) — see the comment on
`Hachi/Composition.lean:53` for the full list.

Worth prioritising this file specifically: `AGENTS.md` requires the matching `docs/wiki/` page to be
updated in the same PR when repo structure changes, and this PR changes structure substantially (five
umbrella renames, `Gadget.lean` → `Gadget/{Basic,Core}.lean`, `Escape.lean` removed). So this is the
page most likely to be trusted and most damaging to leave stale.

docs/skills/make-pr-ready.md
```diff
@@ -42,6 +42,24 @@ Work through these in order. Do not stop until every item is complete.
   revert, `git diff --cached --stat` shows huge index-vs-HEAD numbers for `_generated/` — that is
   the size of the revert, **not** new drift; do not panic. Confirm the post-commit guard view with
   `git diff --cached --quiet origin/main -- docs/kb/_generated/` (must be clean).
+- **Strip stray planning/scratch `.md` files.** A Markdown plan, design note, or working-notes
+  file that is **not already committed in the repo** (newly added by this branch, or still
+  untracked — e.g. `PLAN.md`, `NOTES.md`, `scratch/*.md`, an agent handoff/TODO dump) must
```
alexanderlhicks: This section authorises destroying untracked, user-owned work after only a warning:

> any … file that is **not already committed in the repo** (newly added by this branch, or still
> untracked — e.g. `PLAN.md`, `NOTES.md`, `scratch/*.md`, an agent handoff/TODO dump) must
> **never** make it into the PR … **warn the user** and remove it from the stage/commit state —
> `git restore --staged <file>` to unstage, and `git rm --cached <file>` (or **delete the file** /
> drop it from the commit)

Two problems. It's out of scope for a CWSS refactor. And more seriously it conflicts with ArkLib's
guardrails on preserving working state — which is not hypothetical here, since this repo's working
tree routinely carries dozens of untracked planning documents (`REWINDING_*.md`, `RING_SWITCH_*.md`,
`SORRY_DISPATCH.md`, `docs/kb/queries/*`) that an agent following this instruction could delete.
Untracked files are by definition not in the PR, so no PR-hygiene goal requires touching them at all.

Suggest dropping the section from this PR. If it returns as its own change it should (a) never delete
or `rm --cached` untracked files without explicit per-file approval, (b) restrict itself to files the
branch actually added *to the commit*, and (c) resolve the real PR base — which #656's change to this
same file already implements properly (fork-aware canonical base ref) and this version does not.
Since both PRs edit overlapping lines, that resolution is worth agreeing before either merges.

ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/Guarded.lean
```diff
@@ -192,7 +194,7 @@ guarded analogue of `CWSSPackage.append`/`▷`. The composed verdict is guarded
 of both checks (`Verifier.IsGuarded.append`), and the composed certificate is the guarded binary
 append theorem `Verifier.append_coordinateWiseSpecialSound_of_guardedLeft` (both currently
```
alexanderlhicks: This docstring is accurate — "both currently sorried B4 milestones" — and I'd like that candour to
reach the **package-level** presentation, because right now it doesn't.

`append_coordinateWiseSpecialSound_of_guardedLeft` (line 135) and its right-hand sibling are `sorry`,
and `#print axioms` confirms the blast radius: `GCWSSPackage.append`, every mixed guarded pairing, and
`hachi_iteration_coordinateWiseSpecialSound` all carry `sorryAx`, while `CWSSPackage.append` and
`EscapeCWSSPackage.append` are clean.

The elaborator test is good work and I re-ran it — all 16 ordered P/E/G/EG pairings elaborate to the
expected lattice join with monotone flags. But that demonstrates **type-level routing, not
semantics**, and a reader of the 2×2 package lattice would reasonably conclude the whole thing is
proved. One sentence — "the lattice composes at the type level; the guarded semantic append is
`sorry`" — fixes it.

Related and easy: `ScalarRound.coordinateWiseSpecialSound_of_mkWitness_scalar` is still `sorry` on
this branch and is **proved, axiom-clean, in #654**. Merging #654 first and preserving that proof
removes one of these.

ArkLib/Commitments/Functional/Hachi/QuadEval/Soundness.lean
```diff
@@ -28,11 +28,13 @@ import ArkLib.OracleReduction.Security.CoordinateWiseSpecialSoundness.Package
     `VerifiedBlock.scaled_short`.
```
alexanderlhicks, line 28: Worth stating explicitly here that `quadEvalBetaSq` (defined at line 95) is a **model** of Lemma 8's
parameter, not the parameter. The paper sets `(β̄, ω̄, γ̄) := (2b^τ, 2ω, b)` — an `ℓ₂` bound of `2b^τ`
on `c̄ⱼ · sⱼ` — whereas `quadEvalBetaSq = subL2NormSqBound (quadEvalZL2SqBound …) = 4·B_z` is a
*squared*-`ℓ₂` bound with dimensional factors.

I found nothing unsound about it and the module does describe it as a generalisation; the ask is only
that it never be described as exact parameter equality with Lemma 8. A line naming the modelled bound
and why it dominates the paper's would settle it.

Same paragraph could record the scope limit the code already respects: the escapes here are Lemma 8's
B/D divergent-transcript cases, **not** a formalization of Lemma 7 (weak binding, bound
`max(2ω̄β̄, 2γ̄)`) — the `LiftCom` / weak-binding integration is still a gap.

blueprint/src/references.bib
```diff
@@ -502,3 +502,20 @@ @inproceedings{NS24
   year         = {2024},
   publisher    = {Springer}
 }
+
+@article{HMZ25,
+  title        = {Sublinear Proofs over Polynomial Rings},
+  author       = {Huang, Mi-Ying Miryam and Mao, Xinyu and Zhang, Jiapeng},
```
alexanderlhicks: Cross-PR heads-up rather than a defect: **your author spelling is the correct one.**

`main` has no HMZ25 entry, so this PR and #654 both add one and they conflict:

- here — `@article{HMZ25, author = {Huang, Mi-Ying Miryam and Mao, Xinyu and Zhang, Jiapeng}, …}` ✅
  matches the paper's title page;
- #654 `references.bib:495-501` — `@misc{HMZ25, author = {Meng-Yuan Michael Huang and …}}` ❌.

Since #654 is likely to merge first, please make sure this entry survives the rebase rather than being
dropped as a duplicate — I've asked on #654 to adopt yours. Same for
`docs/kb/sources/HMZ25/metadata.yml`, where #654 has a `"Scaffolded from references.bib"` placeholder
and yours has the substantive note. (`docs/kb/papers/HMZ25.md` is byte-identical between the two, so
that one resolves trivially.)

Note this file is a **four-way** conflict point — #654, #656, #659 and the open #615 all edit it — so
it may be worth settling the bibliography in one PR and having the others adopt it.

Only nit: both entries say `year = {2025}` with no revision pin, and the revision this work depends on
is dated 2026-05-21.

ArkLib/Commitments/Functional/Hachi/QuadEval/Soundness.lean
tobias-rothmann: This is equivalent and intended; we're using the squared version of bounds/norms throughout because computing square roots is unsafe/makes things very complicated.

ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/Guarded.lean
tobias-rothmann: Since we don't do this for any protocol in Binius that uses OracleVerifier.append or for any VCV-io lemmas that are sorried-out, I wouldn't do this here as well. Might just create unnecessary doc bloat.

ArkLib/Commitments/Functional/Hachi/QuadEval/Reduction.lean
tobias-rothmann: Will fix this. 
There is some inherent limitation to this problem (the extractor technically needs to be poly-time, which we cannot ensure in Lean yet), but these simple false proofs can be mitigated by exposing the extractor. I'll need to make some changes for that, but that should be an improvement,

alexanderlhicks: yep, https://github.com/Verified-zkEVM/VCVio/pull/500 will help once it's merged

alexanderlhicks:
Thank you!

alexanderlhicks (approved):

PR #668, opened 2026-07-25
feat(oracle-reduction): worst-case round-by-round security + ProtocolSpec RBR game glue [ABF26 split 1/4]

alexanderlhicks:
First of **four stacked sub-PRs** splitting #505 (`feat/abf26-plan`) into reviewable buckets. This one
carries the framework layer only: the `OracleReduction` security definitions and the ArkLib-local
`ToVCVio` glue. Rebuilt directly on current `main` rather than merged from the branch, so the diff is
against today's tree.

Split order: **PR-0 (this)** → PR-1 `Data/CodingTheory` definitions → PR-2 ProximityGap/ListDecoding →
PR-3 `ProofSystem/ToyProblem`. #505 stays open as the tracking umbrella.

## What's new

**Worst-case-per-prefix round-by-round security** (`OracleReduction/Security/RoundByRound.lean`):

- `Verifier.rbrSoundnessWorstCase` and `Verifier.rbrKnowledgeSoundnessWorstCase` — the paper-shape
  definitions (cf. ABF26 Def A.5), bounding the bad-transition probability at **every fixed transcript
  prefix**, quantified *before* the challenge draw.
- `rbrSoundnessWorstCase_implies_rbrSoundness` and its knowledge analogue, discharging ArkLib's existing
  *averaged* forms from the worst-case ones **with the same error constants**. ArkLib's `rbrSoundness`
  samples the prefix inside the game and bounds the mixture, which is formally weaker; averaged ≤
  worst-case is the safe direction.

Both bad events are reused **verbatim** from the existing averaged definitions, so the implications are
a genuine strengthening rather than a restatement, and the proofs preserve the error function with no
epsilon inflation.

**`ArkLib/ToVCVio/OracleComp/RbrGame.lean`** (new) — ArkLib-specific `ProtocolSpec` glue: challenge-query
resolution, the master mixture bound the two implications rest on, the `OptionT` challenge-first master
bounds, and two `loggingOracle` lemmas.

**`Execution.lean`** — two prover-step theorems (`Prover.processRound_of_dir_eq_V_to_P`/`_P_to_V`),
resolving the internal dependent `match` on round direction once at the framework level.

## A long-standing admit, closed

`rbrKnowledgeSoundnessOneShot_implies_rbrKnowledgeSoundness` is now **proved**, together with the two
`sorry`s in `KnowledgeStateFunctionOneShot.toKnowledgeStateFunction.toFun_next` that blocked it. All
three predate this work. The whole one-shot cluster has no consumer outside its file, so the
supporting definitions were free to change.

It was not merely unfinished — it was unprovable as set up. A `KnowledgeStateFunction`'s round-0
obligation is `(stmtIn, extracted) ∈ relIn`, forced by `toFun_empty`; but `toRoundByRound.extractMid`
hands back its `witIn` argument at that step, and that argument is universally quantified, so it may
be an arbitrary *invalid* witness. Two changes fix it:

- new `Extractor.RoundByRoundOneShot.toRoundByRoundOfRel`, which selects a witness valid for `relIn`
  whenever one exists (classically — `extractMid` is a function, not an algorithm). It needs `relIn`,
  so it cannot live in the relation-agnostic `toRoundByRound`, which is left untouched;
- the induced state function now carries `∃ v, (stmtIn, v) ∈ relIn` as its disjunct at nonzero rounds,
  replacing one about the extractor's output at that round — unusable, since nothing relates the
  extractor at round `m+1` to the extractor at round `m`.

Together these force the general bad event to imply that `relIn` holds **no** witness for `stmtIn`.
That is the bridge to the one-shot event, whose extra conjunct says the extractor fails on the
prover's query log: the general event cannot mention a log at all, since `extractMid` never receives
one, but with no witness in existence that conjunct holds for free whatever the log. The comparison is
then pointwise — both notions score a syntactically identical game.

## Upstreaming cleanup

Most of the branch's `ToVCVio` work **is already in VCVio**. It was written against VCVio `5f7707fb`
(v4.30); `main` pins `cbd4144b` (v4.31, VCVio #477), which absorbed it — several of the local copies
carried "upstream VCV-io candidate" docstrings and were duly upstreamed. Confirmed by the compiler
(`… has already been declared`), not by name matching:

- `SimulateQ.lean` — **all 30** declarations already upstream; now a compatibility shell. 29 of them
  produced `already declared` errors outright. The 30th,
  `simulateQ_randomOracle_map_uniformFin`, is an exact duplicate too — identical statement *and*
  proof, down to already using `StateT.run'_map'` — but it never clashed, because both copies sit at
  root scope in *different* modules: this file imported `RandomOracle/Basic`, never
  `RandomOracle/Simulation`. Importing the latter lets the sole consumer
  (`KZG/Sampling.lean:56`) resolve upstream. A duplicate that is merely out of scope is still a
  duplicate.
- `RbrGame.lean` — 3 `probEvent_bind_le_*` lemmas dropped; the ~10 `ProtocolSpec`/`loggingOracle` glue
  lemmas are genuinely ArkLib-specific and stay.
- `SubSpec.lean` — both remaining lemmas upstream; now a compatibility shell, matching the convention
  already used for `EvalDist/Defs/Support.lean` and friends. `bind_liftComp_map` deleted with them: no
  call site anywhere in the tree, and it duplicates Mathlib's `bind_map_left`.
- `ToVCVio/ToMathlib/Control/StateT.lean` deleted; `SimulateQ` and the two KZG call sites use VCVio's
  `StateT.run'_map'`.

That is why this PR is 723/94 rather than the ~1214 lines the split plan projected: roughly 550 lines
would have been re-duplicated.

**To be precise about what that cleanup is worth, since it is easy to overstate:** those ~550 lines are
code *not added*, not code removed — they only ever existed on `feat/abf26-plan`. Measured against
`main`, this PR removes exactly **three** lemmas (~53 lines). Two of those are real wins:
`simulateQ_randomOracle_map_uniformFin` was defined at root scope under an identical name in both
ArkLib and VCVio and survived only because no module imported both — a latent ambiguity, now defused —
and `bind_liftComp_map` was unreferenced and duplicated Mathlib. The third is a rename.

## Declarations that land ahead of their consumers

By design of the split, most of `RbrGame.lean` has no in-tree consumer yet — only the master mixture
bound is used today (by the two implications). The `OptionT` bounds, the `loggingOracle` lemmas, and the
`$ᵗ`↔`$ᵖ` bridge are consumed by `ProofSystem/ToyProblem/**`, arriving in PR-3. This is stated in the
module docstring's staging note so a reviewer of this PR alone isn't left guessing.

Each staged declaration was checked to have a **live tactic call site** on `feat/abf26-plan`, not merely
a docstring mention. Two failed that check and were **deleted rather than staged**:
`…_getChallenge_first_bind_le_add` (superseded by its `_convex` sharpening, which is what the branch
actually calls) and `Prover.runToRound_succ` (no consumer on any ref).

One honest caveat on the headline pair: only the **knowledge** half
(`rbrKnowledgeSoundnessWorstCase` + its implication) has a consumer on the parent branch, at
`ToyProblem/Spec/General.lean` and `Spec/ErasureDecoder.lean`. The plain-soundness half is kept as the
definitional companion — it is the same paper definition specialised to soundness, and splitting the
pair would be worse than carrying it — but it is, today, unused.

## Verification

- `lake build` — **green, 4122 jobs, 0 errors**.
- `#print axioms` on every new and changed declaration — `{propext, Classical.choice, Quot.sound}`,
  **zero `sorryAx`**. Two `Execution.lean` theorems don't even need `Classical.choice`.
- **Three pre-existing `sorry`s removed, none added.** `RoundByRound.lean` is now sorry-free (see
  above). The one remaining in this PR's files is `Reduction.runWithLog_discard_logs_eq_run`
  (`Execution.lean`), unchanged from `main` and now documented as propagating `sorryAx` through
  `simp` — which `main` disclosed nowhere.
- **No new lint** in this PR's hunks.
- `scripts/validate.sh --lint --docs` has one failure, `Paper page without matching BibTeX key:
  docs/kb/papers/NOZ26.md` — **pre-existing on `main`** and untouched here (this PR changes neither
  `docs/kb/papers/` nor `references.bib`, so the check's inputs are identical to `main`'s).

## Documentation

`docs/kb/audits/open-problems-list-decoding-and-correlated-agreement.md` recorded ABF26 Definition A.5
as `present-but-different`, on the grounds that ArkLib had only the averaged forms. This PR makes that
false, so the row is promoted to `present` citing `rbrKnowledgeSoundnessWorstCase`.
`docs/wiki/repo-map.md`'s description of `ArkLib/ToVCVio/` ("helper lemmas before they are upstreamed")
no longer covers a directory containing never-upstreamable glue and five compatibility shells, and is
extended accordingly. `docs/kb/_generated/**` is deliberately untouched — CI blocks committing it from
feature PRs. No `\lean{}` reference in `blueprint/src/` points at any declaration removed here.

## Review

Two adversarial multi-agent review rounds were run over this diff before opening — 51 findings raised
in total, 17 confirmed, the rest refuted by independent verifiers. Every confirmed finding was
documentation accuracy or API hygiene; none touched a statement, a proof, or axiom hygiene. Acting on
them, two declarations were **deleted** rather than shipped, each verified to have no tactic call site
anywhere including the source branch: `…_getChallenge_first_bind_le_add` (its `_convex` sharpening is
what the eventual consumer actually calls) and `Prover.runToRound_succ`.

Non-vacuity was established constructively rather than argued: concrete protocols inhabiting both new
definitions were compiled at error `1/N` (instantiated at `N = 2^128`), the per-prefix game probability
was shown to be *exactly* `N⁻¹`, error `0` was proved **impossible** for that instance, and the two
implications were applied to those inhabitants to produce the averaged notions with the same constant.


alexanderlhicks:
**Note on the relationship to VCVio [#475](https://github.com/Verified-zkEVM/VCVio/pull/475)** (which adds a protocol-agnostic round-by-round layer), recorded here so reviewers of this PR know what is and isn't affected. **Short version: it does not block or change this PR.**

### What overlaps

VCVio #475's *generic* `KnowledgeTransitionFamily.IsBounded` packages, under the evident translation, exactly the inner obligation of this PR's `Verifier.rbrKnowledgeSoundnessWorstCase` — round ↦ challenge index, context ↦ statement-and-prefix, `preState`/`postState`/`extractBefore` ↦ the knowledge state function and `extractMid`. Its `badEvent` is our worst-case bad event definitionally.

### What does not overlap

- **#475 has no averaged, prover-sampled notion.** Its bounds quantify over contexts with the probability taken over the fresh challenge only. So the worst-case ⇒ averaged bridges proved in this PR are **complementary to it, not subsumed by it** — its `extractionCondition_iff_isBounded` relates two worst-case forms to each other.
- **`ExtractionCondition` is not our `rbrKnowledgeSoundnessOneShot`.** Ours quantifies over prover and input witness, samples the prefix by running the prover, and feeds the prover's query log to the extractor; theirs fixes context and message before sampling. Easy to assume equivalent from the names; it isn't.
- **Its generic transition family is law-free** (data only, no axioms), whereas our `KnowledgeStateFunction` carries three obligations.

An earlier revision of `RbrGame.lean`'s module docstring in this branch said #475 "may subsume" the bridge and implied the two one-shot notions matched. Both were wrong and are corrected in the branch; flagging it in case anyone read the intermediate state.

### Consequence for this PR: none, deliberately

We considered building an adapter from ArkLib's definitions onto #475's structures and decided against it. Because the transition family is law-free, an adapter would transfer the probability bound but none of our structure, and the theorems it unlocks are `simp`-level restatements. It would add a dependency from ArkLib's core security definitions onto an in-flight upstream API in exchange for re-packaging. Not worth it now; revisit if VCVio later grows theorems *on top of* that family that we'd actually consume.

What we do still intend, independently of #475: upstream the generic operational content of `ArkLib/ToVCVio/OracleComp/RbrGame.lean` to VCVio — the mixture bounds (after generalising the challenge oracle to an arbitrary uniform-answer `QueryImpl`; `QueryImpl.addLift` is already upstream), the two `loggingOracle` lemmas, and the `ℝ≥0∞` identity. ArkLib would then keep thin specialisations, or a compatibility shell, exactly as this PR already does for `SimulateQ.lean` and `SubSpec.lean`. That is orthogonal to #475 and judged on its own merits.

alexanderlhicks:
**Update: `RbrGame.lean` split along the `ToVCVio` boundary, and the upstreaming roadmap.**

### What changed in this PR

`ArkLib/ToVCVio/OracleComp/RbrGame.lean` was the only file in the entire `ToVCVio` tree that imported ArkLib core — an inverted layer, since `ToVCVio` is a shim that should sit *below* core. It got there because it mixed two different things and was filed by the destination of one half. Now split:

- **`ArkLib/ToVCVio/OracleComp/QueryTracking/LoggingOracle.lean`** (new) — the two `loggingOracle` lemmas, which mention nothing outside VCV-io. The path mirrors VCV-io's own `OracleComp/QueryTracking/LoggingOracle.lean`, so upstreaming is a straight file move.
- **`ArkLib/OracleReduction/Security/RbrGame.lean`** (moved) — the `ProtocolSpec` glue: challenge-query resolution, the master mixture bound, the three `OptionT` bounds, the `ℝ≥0∞` identity, and the `$ᵗ`/`$ᵖ` bridge. Now next to `Security/RoundByRound.lean`, its only consumer.

The `$ᵗ`/`$ᵖ` bridge is generic in content but needs ArkLib's `Pr_{…}` notation, so it moved to core too: depending on *any* ArkLib layer, not just core, breaks the property that makes a `ToVCVio` file movable.

This establishes an invariant, now stated in the directory README and `docs/wiki/repo-map.md` and verified to hold: **nothing under `ToVCVio` imports ArkLib outside `ToVCVio` itself.** It also retires the awkward "ArkLib-flavoured glue" category the README had grown to justify the misfiled file.

Build green (4123 jobs), docs integrity green, axiom probe unchanged across the move.

For the record: a reviewer raised this layering point during the first review gate, and it was refuted on the grounds that no rule, linter, or import cycle forbade it. That was procedurally right and design-wise wrong — worth recording so the reasoning isn't repeated.

### Upstreaming roadmap

Decided against combining this work with VCVio [#475](https://github.com/Verified-zkEVM/VCVio/pull/475). The two are independent, and neither shrinks the other:

1. **Upstream the generic content to VCVio, on its own merits.** `LoggingOracle.lean` moves as-is. The mixture bounds follow once the challenge oracle is generalised from `challengeQueryImpl` to an arbitrary uniform-answer `QueryImpl` (`QueryImpl.addLift` is already upstream); ArkLib then keeps thin specialisations, or a compatibility shell. Not blocked on #475 — verified no file overlap with it beyond a one-line import list.
2. **No adapter onto #475's structures.** Its generic transition family is law-free, so an adapter would carry the probability bound but none of ArkLib's structure, unlocking only `simp`-level restatements, in exchange for a dependency from ArkLib's core security definitions onto an in-flight upstream API. Revisit only if VCVio grows theorems *on top of* that family that ArkLib would consume.
3. **ArkLib's RBR definitions stay.** `rbrKnowledgeSoundness` mentions `Prover.runWithLogToRound` and cannot move; #475 has no averaged, prover-sampled notion, so the worst-case ⇒ averaged bridges here are complementary to it rather than redundant.

Findings behind this are in the note on [#475](https://github.com/Verified-zkEVM/VCVio/pull/475).

PR #689, opened 2026-08-05
Folding Context PoC

ElijahVlasov:
Folding context reduces the input complexity for the various folding lemmas. 

A folding context is a triple of natural numbers `k, d, n` with the property `1 ≤ k ≤ d ≤ n`. We're introducing grind/simp lemmas to simplify working with folding contexts.


ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+
+import Mathlib.Algebra.Order.Group.Nat
+import Mathlib.Algebra.Order.Monoid.Unbundled.Pow
+import Mathlib.Data.Nat.Cast.Order.Basic
+import Mathlib.Order.Lattice.Nat
+
+/-! # The folding context
+
+Folding arguments (as in FRI/STIR/WHIR) are parameterized by three natural numbers:
+z`
```
alexanderlhicks: assuming z here is a typo.

alexanderlhicks:
I'm curious about the likelihood that having everything as outParam gets annoying in some future use case where typeclass inference might get annoying?

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+  FoldingContextLeft k d, FoldingContextRight d n where
+
+/-- A full context yields the middle one by transitivity: `k ≤ d ≤ n`. -/
+instance {k d n : ℕ} [FoldingContext k d n] : FoldingContextMiddle k n where
+  k_ge_1 := FoldingContextLeft.k_ge_1
+  k_le_n :=
+    le_trans FoldingContextLeft.k_le_d FoldingContextRight.d_le_n
+
+/-- `k` is nonzero in a `FoldingContextLeft k d`. -/
+instance {k d : ℕ} [FoldingContextLeft k d] : NeZero k where
```
alexanderlhicks: claude is reporting NeZero here and in other places (with k being an outParam) now requires far more heartbeats as typeclass inference runs over NeZero ?x on a Nat.

> Measured: a failing `NeZero m` query costs 27 806 heartbeats without this import and 32 660
> with it. Demoting these four `instance`s to plain `theorem`s brings it back to *exactly*
> 27 806, so the ~17% is attributable to them and nothing else.

Given you said this PR was load bearing for follow up work I think we can agree to keep this setup but maybe the instances could be scoped?

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+    omega
+
+/-- `d` is nonzero, since `1 ≤ k ≤ d`, in a `FoldingContextLeft k d`. -/
+instance {k d : ℕ} [FoldingContextLeft k d] : NeZero d where
+  out := by
+    have := FoldingContextLeft.k_ge_1
+    have := FoldingContextLeft.k_le_d
+    omega
+
+/-- `n` is nonzero, since `1 ≤ k ≤ n`, in a `FoldingContextMiddle k d n`. -/
```
alexanderlhicks: I see no d in the instance

ElijahVlasov:
> I'm curious about the likelihood that having everything as outParam gets annoying in some future use case where typeclass inference might get annoying?

I don't think we'll ever rely much on inference when dealing with `FoldingContext`. It's pretty much the same thing as `NeZero` but for triples `1 \le k \le d \le  n`. Either you already have it in context or you bake up an instance manually with constructors. Usually it should be something like `FoldingContext.mk' (by grind) (by grind) (by grind)`. Pretty much analogous to the `NeZero` situation.

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+@[reducible]
+def oneStep {k d n : ℕ} [FoldingContext k d n] : FoldingContext 1 d n where
+  k_ge_1 := by rfl
+  k_le_d := by
+    have := FoldingContextLeft.k_ge_1
+    have := FoldingContextLeft.k_le_d
+    omega
+  d_le_n := FoldingContextRight.d_le_n
+
+attribute [simp, grind →] FoldingContextLeft.k_ge_1 FoldingContextLeft.k_le_d
```
alexanderlhicks: grind seems fine but this seems to slow down simp given the very generic inequalities between Nats in FoldingContextLeft. Where simp actually uses this grind can probably do the job.

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+lemma n_sub_1_sub_k_sub_1_eq_n_sub_k {k n : ℕ} [FoldingContextMiddle k n] :
+  n - 1 - (k - 1) = n - k := by
+  have := FoldingContextMiddle.k_ge_1
+  have := FoldingContextMiddle.k_le_n
+  omega
+
+/-- In a group, `2 ^ n / 2 ^ k = 2 ^ (n - k)`: the folded domain size is the quotient of
+  the original size by the block size. -/
+@[simp, grind =]
+lemma pow_2_n_sub_k_eq_n_sub_k
```
alexanderlhicks, line 191: uses FoldingContext k d n despite there being no d, maybe FoldingContextMiddle is fine (assuming the omission of d there is correct).

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+
+attribute [simp, grind →] FoldingContextLeft.k_ge_1 FoldingContextLeft.k_le_d
+                          FoldingContextRight.d_le_n FoldingContextMiddle.k_le_n
+
+attribute [grind cases] FoldingContext
+
+/-- Monotonicity of truncated subtraction on the context bounds: `k - 1 ≤ n - 1`.
+  Appears when comparing block indices after a single halving step. -/
+@[simp high, grind! →]
+lemma k_sub_one_le_n_sub_one {k d n : ℕ} [FoldingContext k d n] :
```
alexanderlhicks: uses `FoldingContext k d n` despite there being no d, maybe FoldingContextMiddle is fine.
comment also applies to several other lemmas below this (possible with a different FoldingContextXYZ that should replace `FoldingContext k d n`

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+  k - 1 ≤ n - 1 := by
+  grind
+
+/-- `2 ^ k ≤ 2 ^ n` in any ordered monoid where `1 ≤ 2`: the block size never exceeds
+  the domain size. -/
+@[simp, grind! →]
```
alexanderlhicks: similar simp comment as above. haven't rebuilt all the proofs without these, but maybe if simp is strictly needed pointing to the lemmas directly or having a local simp set would help avoid performance overall performance loss.

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+
+/-- Truncated subtraction cancels on the left of `k`, since `1 ≤ k`. -/
+@[simp, grind =]
+lemma one_add_sub_one {k d : ℕ} [FoldingContextLeft k d] :
+  1 + (k - 1) = k := by
+  rw [Nat.add_sub_cancel' (by simp)]
+
+/-- Truncated subtraction cancels on the left of `k`, since `1 ≤ k`. -/
+@[simp, grind =]
+lemma one_add_sub_one' {k n : ℕ} [FoldingContextMiddle k n] :
```
alexanderlhicks, line 174: wondering if it's possible to just keep one of the two one_add_sub_one lemmas given they are the exact same up to FoldingContextXYZ and these are just instances of the same class?

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+    (2 : A) ^ n / (2 : A) ^ k =
+      (2 : A) ^ (n - k) := by
+      rw [div_eq_mul_inv]
+      exact (pow_sub 2 (by simp)).symm
+    _ = (2 : A) ^ (n - k) := by simp
+
+/-- The shifted-by-one variant of `pow_2_n_sub_k_eq_n_sub_k`:
+  `2 ^ (n - 1) / 2 ^ (k - 1) = 2 ^ (n - k)`. -/
+@[simp, grind =]
+lemma pow_2_n_sub_1_sub_k_sub_1_eq_n_sub_k
```
alexanderlhicks: the name misses the division step :D

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+  calc
+    (2 : A) ^ (n - 1) / (2 : A) ^ (k - 1) =
+      (2 : A) ^ ((n - 1) - (k - 1)) := by
+      rw [div_eq_mul_inv]
+      exact (pow_sub 2 (by simp)).symm
+    _ = (2 : A) ^ (n - k) := by simp
+
+/-- `n - k + k` cancels, since `k ≤ n`. -/
+@[simp, grind =]
+lemma n_sub_k_add_k {k n : ℕ} [FoldingContextMiddle k n] :
```
alexanderlhicks, line 218: wondering if these are potential AI fluff that is not strictly required (i.e. any grind call in any proof that requires this would just deal with it without having this lemma) but the file isn't very big so not a critical issue / haven't tested this. probably saves some grind time. simp comment probably applies again here?

ArkLib/Data/CodingTheory/ProximityGap/Folding.lean
```diff
@@ -323,7 +324,7 @@ private lemma eval_comm {f : Polynomial (Polynomial F)} {a x : F} :
         Polynomial.eval₂_eq_sum, Polynomial.sum_def]
 
 private lemma interpolate_eq_folding_poly_eval
-  (hk : k ≤ n)
+  [FoldingContextMiddle k n]
```
alexanderlhicks, line 327: aside from the type class bring useful, this does seem like it makes the statement less readable

ArkLib/Data/CodingTheory/ProximityGap/Folding.lean
```diff
@@ -364,131 +365,104 @@ private lemma interpolate_eq_folding_poly_eval
 /-- Perfect completeness of folding: folding a codeword is the same as
   applying `polyFold` and then encoding.
 -/
-theorem foldWord_codeword {d : ℕ}
+theorem foldWord_codeword {d : ℕ} [FoldingContext k d n]
   {α : F}
-  (hk : k ≤ n)
-  {p : ReedSolomon.code (domain : Fin (2 ^ n) ↪ F) d} :
+  {p : ReedSolomon.code (domain : Fin (2 ^ n) ↪ F) (2 ^ d)} :
```
alexanderlhicks, line 374: I know the topic of d vs 2^d came up in a previous domain PR, would be useful to just make this explicit in a docstring or such to avoid misleading any reader

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
```diff
@@ -0,0 +1,290 @@
...
+    omega
+
+/-- `n` is nonzero, since `1 ≤ k ≤ n`, in a `FoldingContextMiddle k d n`. -/
+instance {k n : ℕ} [FoldingContextMiddle k n] : NeZero n where
+  out := by
+    have := FoldingContextMiddle.k_ge_1
+    have := FoldingContextMiddle.k_le_n
+    omega
+
+/-- `n` is nonzero, since `1 ≤ k ≤ n`, in a `FoldingContext k d n`. -/
```
alexanderlhicks: Is this redundant via` instance {k d n : ℕ} [FoldingContext k d n] : FoldingContextMiddle k n` ->  `instance {k n : ℕ} [FoldingContextMiddle k n] : NeZero n`

ArkLib/Data/CodingTheory/ProximityGap/Folding.lean
alexanderlhicks: this is also a different assumptions since FoldingContextMiddle requires k >= 1

alexanderlhicks: I guess k=0 is just the identity so it doesn't matter 👍

ArkLib/Data/CodingTheory/ProximityGap/Folding.lean
```diff
@@ -707,15 +681,10 @@ private lemma contradictory_hamming_dist_zero :
 
 @[simp]
 private lemma contradictory_hamming_dist_formula {s : Finset F}
-  {d : ℕ}
-  (h_s : s ⊆ (domain.subdomain k).toFinset)
-  (h_k_d : 2 ^ k ≤ d)
-  (h_d : d ≤ 2 ^ n) :
+  {d : ℕ} [FoldingContext k d n]
```
alexanderlhicks: a bit awkward introducing d here when it doesn't show up in the statement itself just for the sake of bringing in FoldingContext

ArkLib/Data/Domain/CosetFftDomain/Mem.lean
```diff
@@ -91,6 +91,10 @@ lemma mem_toFinset_iff_mem [Fintype ι] [DecidableEq F] :
 lemma mem_toFinset_self [Fintype ι] [DecidableEq F] {i : ι} :
   ω i ∈ toFinset ω := by simp
 
+@[simp high]
```
alexanderlhicks: first time I see simp high 👀 
How come this was needed? doesn't seem like a particularly tricky lemma

alexanderlhicks:
> > I'm curious about the likelihood that having everything as outParam gets annoying in some future use case where typeclass inference might get annoying?
> 
> I don't think we'll ever rely much on inference when dealing with `FoldingContext`. It's pretty much the same thing as `NeZero` but for triples `1 \le k \le d \le n`. Either you already have it in context or you bake up an instance manually with constructors. Usually it should be something like `FoldingContext.mk' (by grind) (by grind) (by grind)`. Pretty much analogous to the `NeZero` situation.

makes sense :+1:

ArkLib/Data/CodingTheory/ProximityGap/Folding/FoldingContext.lean
ElijahVlasov: Not sure. If you made it only for one class then at the "call-sites" you'd have to convert an instance of one class into the other which introduces some friction.

ArkLib/Data/CodingTheory/ProximityGap/Folding.lean
ElijahVlasov: Given it's a private lemma we don't care much about it

alexanderlhicks:
Since this branch lives on a fork with maintainer edits disabled, we couldn't resolve the conflicts with `main` in place — #825 carries these commits unchanged, merged with current main (v4.33.1), conflict-resolved and validated (validate.sh green, axiom profile unchanged). Follow-up there; this PR can be closed in favour of #825.

alexanderlhicks:
Merged via #825 (2e5ba2862) — all commits carried over with the main merge, validate-green. Thanks @ElijahVlasov!

PR #611, opened 2026-07-06
Tensor product of MCA generators

katyhr:
Lemma 4.4 of [BCGM25]. Weaker version (checking with authors)

PR #692, opened 2026-08-05
feat: module-alphabet MCA unification + BCGM25 Lemma 4.4 at the tight error

alexanderlhicks:
## Summary

Unifies ArkLib's two mutual-correlated-agreement (MCA) formalizations by generalising the BCGM25 generator framework from field alphabet `F` to module alphabets `ModuleCode ι F A`, and uses the gained expressiveness to state and prove **BCGM25 Lemma 4.4 at the paper's tight error `ε + ε′`** — previously unreachable (and, as stated with the same-code hypothesis, unprovable).

- **`Basic/LinearCode.lean`** — `projectedCodeSubmod` / `mem_projectedCodeSubmod_iff` / `projectedCode_linearCombination` generalised to `ModuleCode ι F A` (`[Field F]` weakened to `[Semiring F]`, `*` → `•`).
- **`ProximityGap/ProximityGenerators.lean`** — `IsMCA` / `IsMCAGenerator` generalised to module alphabets, matching BCGM25 Def 3.2's alphabet generality (`Matrix.vecMul` → `∑ j, G x j • U j`, with `vecMul_eq_smul_sum` recording agreement at `A := F`); distance argument renamed `γ → δ`. New value form `mcaError` + `isMCAGenerator_iff_mcaError_le`.
- **`InterleavedCode.lean`** — new `projectedCodeSubmod_moduleInterleavedCode_iff`: projection commutes with interleaving (row-wise), the bridge that lets MCA be applied to interleaved codes `C^⋈ℓ`.
- **`ProximityGap/TensorGenerator.lean` (new)** — the complete Lemma 4.4 story, all axiom-clean with no `sorry`:
  - `tensor_isMCAGenerator` (tight): if `G` has MCA error `ε` for `MC` and `G′` has MCA error `ε′` for the ℓ-fold interleaving `MC^⋈ℓ`, then `G ⊗ G′` has MCA error `ε + ε′` for `MC`;
  - `tensor_isMCAGenerator_of_base`: Lemma 4.4 **at its printed hypothesis** (`G′` has MCA for `MC` itself), at the weaker error `ε + ℓ·ε′` — the union bound over rows is forced;
  - `isMCAGenerator_of_moduleInterleavedCode`: interleaved MCA ⇒ base MCA at the same error, so the tight form's hypothesis is provably a strengthening of the printed one.
- **`ProximityGap/MCAGenerator.lean`, `AffineGenerator.lean`** — Lemmas 4.1/4.2 transported to module alphabets; Lemma 7.1 kept at `A = F` (its counting lives in the coefficient space; generalising `Prelims.lean`'s `affineComb`/`linComb` is deferred).

## On the interleaved hypothesis (deliberate deviation from the printed Lemma 4.4)

The printed Lemma 4.4 assumes `G, G′` have MCA for `C`, but its proof bounds Equation (5) by applying `G′`'s MCA to the ℓ-fold interleaving `C^ℓ` at the stacked family `w_j = (u_{(1,j)}, …, u_{(ℓ,j)})` — which is exactly the interleaved MCA event and is crucially independent of the outer seed. This PR hypothesises `G′`'s MCA **at the interleaving**, which is what the proof actually uses; with the same-code hypothesis one is forced into a per-row union bound and the weak error `ε + ℓ·ε′`. The strengthening is invisible in the paper's applications: Theorem 6.1's MDS error depends only on the relative distance, and `δ(C^ℓ) = δ(C)`. Details in the module docstring. (An adversarial review compiled the converse direction — interleaved MCA implies plain MCA at the same error — confirming the hypothesis is strictly stronger and the printed proof does not establish the printed statement; to be raised with the BCGM25 authors.)

## Verification

- `lake build ArkLib` green; `./scripts/validate.sh` green; lint kind-counts identical to base.
- Zero new `sorry`; all touched files remain `sorry`-free.
- `#print axioms` = `[propext, Classical.choice, Quot.sound]` on `tensor_isMCAGenerator`, `projectedCodeSubmod_moduleInterleavedCode_iff`, `mcaError`, `isMCAGenerator_iff_mcaError_le`, and Lemmas 4.1/4.2/7.1.
- Regression probe: `IsMCA` at `A := F` is `Iff.rfl` against the verbatim pre-PR definition (semantics-preserving for downstream PRs), plus a probe exhibiting the tight bound's mechanism (the case-(b) interleaved family is outer-seed-independent).
- Independently reviewed by four adversarial passes (soundness / #610-#611 fit / abf26 & proximity-prize compatibility / library integration); all soundness claims held, integration findings fixed in the last commit.

## Relation to other PRs

- **#610 / #611**: `main` previously had BCGM25 Lemmas 4.1/4.2 + Def 4.3 but no Lemma 4.4 in any form. This PR adds the tight Lemma 4.4 next to them; #610 can drop its unprovable-as-stated `tensor_of_MCA_is_MCA_tight` sorry after rebasing (its Theorem 6.1 and the interleaved-distance transfer `δᵣ(MC^⋈κ) = δᵣ(MC)` still need restating/proving — deferred here as they are not on the Lemma 4.4 path). #611's weak `ε + ℓ·ε′` has genuinely weaker hypotheses and remains valid; it is subsumed rather than superseded.
- **#505 (abf26)**: `epsMCA`/`mcaEvent` there is the (module-alphabet, affine-line) instance of the unified notion; a follow-up on that branch bridges `epsMCA C δ = mcaError (AffineLineGenerator F) MC δ` and retires the duplicate.
- Deliberately does **not** adopt #610's `ProximityGap/` → `ProximityGenerator/` directory rename; this PR stays on `main`'s layout and #610 rebases on top.


alexanderlhicks:
cc @katyhr

alexanderlhicks:
Independent adversarial review of this branch at `f04ee847`, done in a clean worktree from scratch (full build, `validate.sh`, `#print axioms`, and every paper claim re-derived against BCGM25/ABF26). **Verdict: merge — 0 soundness defects.** All three tensor theorems check out by hand, and the alphabet generalisation is *forced* rather than optional: BCGM25 Def 3.14 writes the combination as `G(x) · U` with `U ∈ Σ^{ℓ×n}`, so the `Matrix.vecMul` form could only ever express `Σ = F`, and neither Lemma 4.4's real hypothesis nor Lemma 4.5's `C ⊗ K ⊆ (Σ⊗K)ⁿ` is even *statable* without it.

Two commits pushed on top (`0bffd657`, `9cbb6135`) — build green, `validate.sh` green, 16/16 declarations axiom-clean, `isMCA_iff_vecMul_form` still `Iff.rfl`:

1. **`blueprint/lean_decls` was red on `main`, not here.** `checkdecls` has been failing since #722 moved `ListDecodable.*` into `namespace Code` without updating the blueprint. Folded the fix in here since it blocks this PR.
2. **Radius moved to `δ : ℝ`** on `IsMCA`/`mcaError` (matching `Code.Lambda`); `IsMCAGenerator` is unchanged in meaning and still quantifies `δ : I` with `ε_mca : I → ℝ≥0`, which is where BCGM25 puts `[0,1]`. No proof used the `I`-membership — it was only in binder types. Payoff, checked rather than argued: the abf26 bridge `epsMCA C δ = mcaError (AffineLineGenerator F) MC (δ : ℝ)` now compiles with **no hypothesis** on the whole domain, and `gridPt` can stay total (the recorded "`gridPt : ℕ → I` cannot be made total" constraint only existed because the value took `I`).
3. **Value-level API landed**: `mcaError_le_one`, `mcaError_ne_top`, `mcaError_mono` (BCGM25 Lemma 3.16), `size_clause_iff_floor`, `mcaError_eq_of_floor_eq`. Without these `epsMCA` could not be retired in favour of `mcaError`, so "one MCA definition" was not reachable from this PR as it stood.
4. **Docs**: both conventions pages claimed ABF26's `(0,1)` was "for `Λ`, a different slot". It isn't — Defs 4.1/4.3, Fact 4.5 and Lemmas 4.6/4.7 all write `δ ∈ (0,1)` for the ε-error slot while §1.2 and both Grand Challenges write `[0,1]`. The paper self-contradicts here exactly as it does for `Λ`; both pages now say so. Also unparked the `IsMDS → ModuleCode` pointers in `repo-map.md` and the list-decoding audit.

### Three claims in the PR description that the code no longer supports

The `TensorGenerator.lean` docstring is careful and correct; the description has drifted from it.

- *"confirming the hypothesis is strictly stronger"* — `isMCAGenerator_of_moduleInterleavedCode` shows the interleaved hypothesis implies the base one, i.e. **at least** as strong. No separation is exhibited. The docstring already says exactly this ("not claimed and not known").
- *"#610 can drop its unprovable-as-stated `tensor_of_MCA_is_MCA_tight` sorry"* — unsupported, and the accurate statement is more useful: the case-(b) event *is* the interleaved MCA event, so the printed error follows from the printed hypothesis as soon as `ε_mca(C^ℓ) ≤ ε_mca(C)` — which **ABF26 states as an open question** immediately after its Lemma 4.7. That `ℓ` is the open factor, not a formalisation artefact. Recorded in the module docstring in `9cbb6135`.
- *"#611's weak `ε + ℓ·ε′` ... is subsumed rather than superseded"* — understated. #611's `tensor_of_MCA_is_MCA` has the *same* hypotheses and the *same* conclusion as `tensor_isMCAGenerator_of_base` here (modulo `ε : I → ℝ` plus nonnegativity side conditions, which `main` has since improved away to `ℝ≥0`). It is fully superseded.

### Still open, and out of scope for this PR

- No unconditional inhabitant of `IsMCAGenerator` exists anywhere — all four results are transports. BCGM25 Theorem 6.1, the base case, is still a `sorry` elsewhere.
- `δᵣ(MC^⋈κ) = δᵣ(MC)`, which the "free at Theorem 8.2" reading needs, has no in-tree witness. Short proof: a nonzero interleaved codeword has a nonzero row, so `Δ(C^⋈κ) = Δ(C)`.
- MCA is unified; the wider CA/PG family is not (seven `δ_ε_*` definitions in `ProximityGap/Basic.lean`, two more in `DG25/`). The new conventions page documents the plan.

ArkLib/Data/CodingTheory/ProximityGap/AffineGenerator.lean
```diff
@@ -12,15 +12,30 @@ import ArkLib.Data.CodingTheory.Prelims
 import Mathlib.FieldTheory.Finiteness
 
 /-!
-## Main Results
+# Mutual correlated agreement for affine space generators
 
-- Lemma 7.1. [BCGM25]: Mutual correlated agreement (MCA) for the affine line generator implies
-MCA for the affine space generator.
+Mutual correlated agreement for the affine line generator `F → F²`, `x ↦ (1, x)`, implies it for
```
katyhr, line 17: Are we sure we want to remove the references to appropriate base results in the BCGM paper?

ArkLib/Data/CodingTheory/ProximityGap/AffineGenerator.lean
```diff
@@ -254,23 +276,23 @@ open Probability
 variable {ι : Type} [Fintype ι]
          {F : Type} [Field F] [Fintype F]
 
-/-- Lemma 7.1. [BCGM25].
-The affine line generator `F → F²`, `x ↦ (1, x)`, having MCA error `ε_mca` for `LC` implies that
+/-- The affine line generator `F → F²`, `x ↦ (1, x)`, having MCA error `ε_mca` for `LC` implies that
```
katyhr, line 279: I think we should keep the reference to Lemma 7.1 here as well as the comment that in fact only \ell \geq 1 is required

ArkLib/Data/CodingTheory/ProximityGap/AffineGenerator.lean
```diff
@@ -254,23 +276,23 @@ open Probability
...
-generator is the affine line generator, so the conclusion is immediate, and the proof below
-covers that case uniformly.
+Only `ℓ ≥ 1` is required: at `ℓ = 1` the affine space generator *is* the affine line generator, so
+the conclusion is immediate, and the proof below covers that case uniformly.
 
-Note: the error is valued in `ℝ≥0` rather than `I`, since `(1 - 1/|F|)⁻¹ • ε_mca` may exceed `1`. -/
-theorem AffineLine_MCA_AffineSpaceMCA {ℓ : ℕ} (hℓ : ℓ ≥ 1) (ε_mca : I → ℝ≥0) (LC : LinearCode ι F)
+The error is valued in `ℝ≥0` rather than `I`, since `(1 - 1/|F|)⁻¹ • ε_mca` may exceed `1`. -/
+theorem isMCAGenerator_affineSpaceGenerator_of_affineLineGenerator {ℓ : ℕ} (hℓ : ℓ ≥ 1)
+    (ε_mca : I → ℝ≥0) (LC : LinearCode ι F)
```
katyhr, line 288: Do we want linear code or module code here?

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
```diff
@@ -9,18 +9,34 @@ import ArkLib.Data.Matrix.Basic
 import ArkLib.Data.Probability.Instances
 
 /-!
-## Main Results
+# Preserving mutual correlated agreement under linear maps on the output
 
-- Lemma 4.1 [BCGM25] : Let `G : S → 𝔽^ℓ` be an MCA generator with error `ε_mca`, and `A` a matrix
```
katyhr, line 14: Same comment about the refs

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
```diff
@@ -31,69 +47,103 @@ open Probability
...
+    (MC : ModuleCode ι F A) (δ : ℝ) (Φ : (ℓ' → (ι → A)) → (ℓ → (ι → A)))
+    (h : ∀ (U : ℓ' → (ι → A)) (x : S), IsMCA G' MC x U δ → IsMCA G MC x (Φ U) δ) :
+    mcaError G' MC δ ≤ mcaError G MC δ := by
+  unfold mcaError
+  refine iSup_le fun U => le_trans (Pr_le_Pr_of_implies ($ᵖ S) _ _ (fun x hx => h U x hx)) ?_
+  exact le_iSup (fun V => Pr_{let x ←$ᵖ S}[IsMCA G MC x V δ]) (Φ U)
+
+/-- Let `G : S → 𝔽^ℓ` be a generator and let `M` be an `ℓ × ℓ'` matrix. Then `G' : S → 𝔽^ℓ'` is a
+generator defined by `x ↦ G(x) · M`.
+This is the generator whose error is bounded by `mcaError_generatorByRightMul_le`. -/
```
katyhr, line 73: Same comment about removing the ref

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
```diff
@@ -31,69 +47,103 @@ open Probability
...
+def generatorByRightMul (G : Generator S ℓ F) (M : Matrix ℓ ℓ' F) : Generator S ℓ' F :=
```
katyhr, line 74: The only difference between this and the original `generatorByRightMul` def is that the matrix name is changed from `A` to `M`.  If we want to be close to the paper, we should keep `A`. Also, it's more customary to have a module named `M`

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
```diff
@@ -31,69 +47,103 @@ open Probability
...
+    fun x ↦ Matrix.vecMul (G x) M
 
 /-- Let `G : S → 𝔽^ℓ` be a generator and `κ` a subset of `ℓ`. Define a new generator
 `G' : S → 𝔽^κ`, which we call a projected generator, by restricting the output of `G` to the indices
 given by `κ`.
-This is the generator `G'` inside Corollary 4.2 [BCGM25] -/
+This is the generator whose error is bounded by `mcaError_projectedGenerator_le`. -/
```
katyhr, line 80: same comment about the ref

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
```diff
@@ -31,69 +47,103 @@ open Probability
...
+    (MC : ModuleCode ι F A) (M : Matrix ℓ ℓ' F) (hM : HasLeftPseudoInverse M) (γ : ℝ) :
+    mcaError (generatorByRightMul G M) MC γ ≤ mcaError G MC γ :=
+  mcaError_le_of_forall_isMCA_imp G (generatorByRightMul G M) MC γ (matrixMulCodewords M)
+    (fun U x h => isMCA_generatorByRightMul_of_isMCA G MC M hM U γ x h)
+
+/-- Let `G : S → 𝔽^ℓ` be an MCA generator with error `ε_mca`, and `M` a matrix
+with a left pseudoinverse. Then the generator `G'` obtained from `G` by right multiplication by `M`
 is an MCA generator with the same error `ε_mca` as `G`.
-Lemma 4.1 [BCGM25]. -/
+The `IsMCAGenerator` form of `mcaError_generatorByRightMul_le`. -/
```
katyhr, line 127: same comment about the ref removal

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
```diff
@@ -102,18 +152,24 @@ lemma isMCA_projectedGenerator_of_isMCA (LC : LinearCode ι F) [Nonempty S] (G :
...
 subset of `ℓ`. Then the projected generator over `κ` is an MCA generator with the same error as `G`.
-Corollary 4.2 [BCGM25]. -/
-lemma generatorSubset [Nonempty S] (G : Generator S ℓ F) (ε_mca : I → ℝ≥0) (LC : LinearCode ι F)
-(hGMCA : IsMCAGenerator G ε_mca LC) (κ : Set ℓ) [Fintype κ] :
-  IsMCAGenerator (projectedGenerator G κ) ε_mca LC := by
-  intro U γ
-  exact le_trans (Pr_le_Pr_of_implies ($ᵖ S) _ _
-          fun x h => isMCA_projectedGenerator_of_isMCA LC G κ U γ x h)
-    (hGMCA (zeroExtend κ U) γ)
+The `IsMCAGenerator` form of `mcaError_projectedGenerator_le`. -/
```
katyhr, line 168: same comment about the reference

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
```diff
@@ -102,18 +152,24 @@ lemma isMCA_projectedGenerator_of_isMCA (LC : LinearCode ι F) [Nonempty S] (G :
...
+lemma generatorSubset [Nonempty S] (G : Generator S ℓ F) (ε_mca : I → ℝ≥0)
+  (MC : ModuleCode ι F A)
+  (hGMCA : IsMCAGenerator G ε_mca MC) (κ : Set ℓ) [Fintype κ] :
+  IsMCAGenerator (projectedGenerator G κ) ε_mca MC :=
+  fun γ => le_trans (mcaError_projectedGenerator_le G MC κ (γ : ℝ)) (hGMCA γ)
```
katyhr, line 173: I don't think we need the explicit `(γ : ℝ)` type annotation

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
```diff
@@ -87,21 +92,152 @@ Definition 3.12 [BCGM25]. -/
...
+The body is the `mcaError` bound, so `isMCAGenerator_iff_mcaError_le` is `Iff.rfl`: `exact`,
+`refine` and `apply` see through to the value inequality, and only `rw`/`simp` need the bridge.
+The pointwise reading — the bound at one individual family `U` — is `IsMCAGenerator.prob_le`.
+
+The radius is quantified over `I`, the closed unit interval: this is the bound, and the bound is
+where `[0,1]` belongs, while the value underneath is total in the radius. See
+`docs/wiki/proximity-error-conventions.md`. -/
+def IsMCAGenerator {S : Type} [Nonempty S] [Fintype S] {A : Type} [AddCommMonoid A] [Module F A]
+  (G : Generator S ℓ F) (ε_mca : I → ℝ≥0) (MC : ModuleCode ι F A) : Prop :=
+  ∀ δ : I, mcaError G MC (δ : ℝ) ≤ (ε_mca δ : ENNReal)
```
katyhr, line 169: ` ∀ δ : ℝ`?

katyhr: Also this deviates a bit from the proximity gens paper. Is this the phrasing in abf?

ArkLib/Data/CodingTheory/ProximityGap/AffineGenerator.lean
alexanderlhicks: I lean in favour of keeping the docstrings minimal and focused on the underlying lean declaration and leaving any specific reference to a paper/lemma at the top level file documentation or in the relevant paper entry of the knowledge. Partly this is for readability (someone might look at the lemma without caring/context about BCGM25), the lemma might appear in different forms elsewhere (we wouldn't cite all references in that case) and to guard against any drift/churn (paper numbering might change, definitions might change...). 
See https://github.com/Verified-zkEVM/ArkLib/blob/main/docs/kb/audits/bcgm25-mca-generators.md for example.

ArkLib/Data/CodingTheory/ProximityGap/AffineGenerator.lean
alexanderlhicks: see comment above.

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
```diff
@@ -87,21 +92,152 @@ Definition 3.12 [BCGM25]. -/
...
+
+/-- **Unfolding lemma for `IsMCAGenerator`.** It *is* the `mcaError` bound, by definition; this is
+the entry point for `rw` and `simp only`, which do not see through a semireducible `def`. -/
+lemma isMCAGenerator_iff_mcaError_le {S : Type} [Nonempty S] [Fintype S] {A : Type}
+    [AddCommMonoid A] [Module F A] (G : Generator S ℓ F) (ε_mca : I → ℝ≥0)
+    (MC : ModuleCode ι F A) :
+    IsMCAGenerator G ε_mca MC ↔ ∀ δ : I, mcaError G MC (δ : ℝ) ≤ (ε_mca δ : ENNReal) := Iff.rfl
```
katyhr, line 176: ∀ δ : ℝ?

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
alexanderlhicks: see comment above.

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
```diff
@@ -200,11 +336,10 @@ noncomputable local instance {F : Type} [Fintype F] {S : Set F} : Fintype S := F
 
 /-- If `G` is a polynomial generator, then `G` is zero-evading with error the maximum of the total
 degrees of the individual polynomials divided by the size of the smallest evaluation sets `S i`.
-Remark 3.20, the version of the statement in the brackets [BCGM25].
-Note: Remark 3.20 provides two ways of viewing a polynomial generator as a zero-evading generator.
-one in terms of individual degrees, and one in terms of total degrees. We choose the total degree
-approach. Ultimately, the reasoning is the same. The difference is the version of Schwartz-Zippel
-used to obtain the upper bound. -/
+
```
katyhr, line 339: I don't understand the need for this docstring to be changed :D

alexanderlhicks: same as above, removing paper specific details which can be maintained in the kb if desired.

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
alexanderlhicks: yeah we have some unfortunate lack of nice choices because we also use A a lot as in ModuleCode ι F A, maybe we just pick something else entirely :sweat_smile:

ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean
alexanderlhicks: good point!

ArkLib/Data/CodingTheory/ProximityGap/TensorGenerator.lean
```diff
@@ -0,0 +1,243 @@
...
+    rwa [hrow] at h
+
+/-- If `G` and `G'` both have mutual correlated agreement for `MC` itself, the tensor generator has
+it for `MC` with the inner error scaled by `Fintype.card ℓ`.
+
+The factor is forced by this route: without the interleaved hypothesis of
+`isMCAGenerator_tensorGenerator_of_moduleInterleavedCode`, the row witnessing non-membership
+depends on the outer seed `x'`,
+so the family fed to `G'` is not fixed and a union bound over the `ℓ` rows is required. -/
+theorem isMCAGenerator_tensorGenerator (G : Generator S ℓ F) (G' : Generator S' ℓ' F)
```
katyhr, line 187: This is what I had here: https://github.com/Verified-zkEVM/ArkLib/blob/300730d3bfb9ea19f62be130e4261092cbf2d595/ArkLib/Data/CodingTheory/ProximityGap/MCAGenerator.lean#L129

This is the loose bound (it changes the error by adding a factor of the size of the field to the second component), which is a looser bound than what is presented in the paper. Ideally, we want to remove this factor

ArkLib/Data/CodingTheory/ProximityGap/TensorGenerator.lean
```diff
@@ -0,0 +1,243 @@
...
+Writing `W x' i := ∑ j, G' x' j • U (i, j)`, the tensor combination is the `G`-combination of
+`W x'`. Case-split on whether some `W x' i` fails to project into the code on the witness set `T`:
+if so, the tensor event implies `G`'s event at the family `W x'`; if not, it implies `G'`'s event
+at the interleaved family `w j := (k, i) ↦ U (i, j) k`, which is independent of `x'`. A union
+bound over the two cases gives the sum. -/
+theorem isMCAGenerator_tensorGenerator_of_moduleInterleavedCode
+    (G : Generator S ℓ F) (G' : Generator S' ℓ' F)
+    (ε_mca ε'_mca : I → ℝ≥0) (MC : ModuleCode ι F A)
+    (hG : IsMCAGenerator G ε_mca MC)
+    (hG' : IsMCAGenerator G' ε'_mca (ModuleCode.moduleInterleavedCode F A ℓ ι MC)) :
```
katyhr, line 81: According to Ignacio, we don't have a condition of the interleaved code.. But I also got errors trying to prove this without it

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
alexanderlhicks: one improvement would be to formalize both, which then removes the need for the details in the original docstring.

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
alexanderlhicks: See https://github.com/Verified-zkEVM/ArkLib/blob/main/docs/wiki/proximity-error-conventions.md
If you had I you would need to carry around of proof of membership of I `k/n` or the usual bounds, which you can still do but just when you need it rather than all the time. Also helped bridge to epsMca which is R>=0 (that is a bit more artificial, and we could get rid of epsMCA in due course if we wanted now that it's redefined as a wrapper around mcaError).
IsMCAGenerator still has I as you can see.

I went back and forth a bit, so I'm not opposed to tweaking things.

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
alexanderlhicks: see above

ArkLib/Data/CodingTheory/ProximityGap/AffineGenerator.lean
alexanderlhicks: we should have module in due course, will just require fixing the proof.

ArkLib/Data/CodingTheory/ProximityGap/TensorGenerator.lean
alexanderlhicks: hmm I should catch up with that conversation and see

ArkLib/Data/CodingTheory/ProximityGap/TensorGenerator.lean
alexanderlhicks: yeah right now there is either this version or the tight version but with interleaved codes, I can see if I can prove the tight version without I guess? but the proof didn't seem to go through.

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
alexanderlhicks: > Is this the phrasing in abf?

There is some alignment with abf26 since this PR came out of laying ground work to be able to merge the abf26 content.

ArkLib/Data/CodingTheory/ProximityGap/TensorGenerator.lean
katyhr: I had the same issue. I think the blocker is at their equation 4.5, it's not exactly clear how to apply the MCA bound there to obtain the needed bound without going to interleaved codes

ArkLib/Data/CodingTheory/ProximityGap/ProximityGenerators.lean
katyhr: Oh I meant, did you want the delta to be in I (after the :=) or do you want it in \R? Because it's currently in \R almost everywhere else

PR #616, opened 2026-07-08
Folding preserves block relative distance balls (Claim 4.22 WHIR)

ElijahVlasov:

PR #696, opened 2026-08-06
chore(BlockRelDistance): naming and docs follow-up to #616

alexanderlhicks:
Follow-up to #616, which was merged as-is because none of these were blocking. Hygiene plus one
hypothesis weakening — **no proof term changes, and nothing here alters what is proved except to
make one statement slightly stronger.**

### Weaker hypothesis

`folding_block_rel_ball` and `folding_preserves_block_balls` took `hd : k ≤ d`, but the only use is
discharging `2 ^ 1 ∣ 2 ^ d`, which needs just `1 ≤ d`. Weakened both. (`1 ≤ d` follows from
`1 ≤ k ≤ d` in WHIR's setting, so this is a strict generalisation of the statement, not a change of
meaning.)

### `private` definitions with public API

`complDisagreementSet` and `agreementBlockUnion` were `private def`, while #616 made **13 lemmas
about them public** — three of those global `@[simp]`. That left rewrite rules in the global simp
set whose left-hand sides name constants no downstream file can write, and it means
`ListDecodability.lean` reasons about definitions it cannot mention. Promoted both definitions to
public with docstrings.

While there: added docstrings to the 13 lemmas #616 newly exposes. (Six lemmas earlier in the file
were already public and undocumented before #616; left alone as out of scope.)

### Names

- `agreement_sub_agreementBlockUnion` → **`agreementBlockUnion_subset_agreement`**. The name was
  the wrong way round: the statement is `agreementBlockUnion k φ f g ⊆ {i | f i = g i}`, but the
  name read as `agreement ⊆ agreementBlockUnion`.
- `complDisagreementSet_def'` → **`complDisagreementSet_def`** and `card_disagreementSet'` →
  **`card_disagreementSet`**. Neither prime had an unprimed sibling, and the unprimed
  `card_disagreementSet` was free — it now pairs with `card_complDisagreementSet`.
  (The primes on `blockRelDistance_eq_one_sub'` and `relHammingDist_le_sub_agreementBlockUnion'`
  are kept: those *do* have `ℚ≥0` siblings.)
- `complDisagreementSet_sub_subdomain` → **`complDisagreementSet_subset_subdomain`**, using
  `subset` for `⊆`.

I left `complDisagreementSet` itself alone. `agreementSet` would read better and pair with
`agreementBlockUnion`, but the current name is systematically derived from the existing
`disagreementSet`, which keeps the relationship explicit and greppable; happy to change it if
preferred.

### Docstring corrections

- `ListDecodability.lean`'s module docstring sat *inside* `namespace ProximityGap` after the
  `open`/`variable` block, had no title, and said it "contains a proof of Theorem 4.20" when the
  content is Claim 4.22. Moved above the namespace, gave it a title, a `## Main results` section,
  and an accurate description — Theorem 4.20 is noted as not yet formalised.
- `subdomain_comp`'s docstring described only the `j = 1` case ("one more folding step … the
  `(k+1)`th subdomain … `n - k - 1 = n - (k + 1)`"), but the lemma is general in `j`. This matters
  because that generality is exactly what lets it subsume `subdomain_one_comp` in #534.

### Deliberately not included

`neg_mem_block_of_mem`, `neg_mem_block_iff_mem` (`Block.lean`) and `foldWord_k_1'` (`Folding.lean`)
have no consumers — verified by deleting them and rebuilding green, and none is `@[simp]`, so there
is no implicit use either. They are left in place: the two `Block.lean` lemmas encode the
"if `x ∈ Block` then `−x ∈ Block`" step of WHIR Claim 4.22 and are reasonable API for that file,
and `foldWord_k_1'` is a fine `funext` companion to `foldWord_k_1`. The only live issue is that
#534 adds an identical `foldWord_k_1'`, which that PR will need to drop.

### Validation

`./scripts/validate.sh` green.

### DIFF
--- a/ArkLib/Data/CodingTheory/Basic/BlockRelDistance.lean
+++ b/ArkLib/Data/CodingTheory/Basic/BlockRelDistance.lean
@@ -147,24 +147,31 @@ noncomputable def blockRelDistanceBall
 scoped notation "Λ𞁒("C", "k", "φ'", "f", "δ")" =>
   blockRelDistanceBall k φ' f δ C
 
-private def complDisagreementSet
+/-- The complement of `disagreementSet` inside the `k`th subdomain: the points `z` of
+  `φ.subdomain k` on which `f` and `g` agree across the whole block indexed by `z`. -/
+def complDisagreementSet
   (k : ℕ) (φ : SmoothCosetFftDomain n F) (f g : Fin (2 ^ n) → F) : Finset F :=
   (φ.subdomain k).toFinset \ disagreementSet k φ f g
 
-lemma complDisagreementSet_def' :
+/-- `complDisagreementSet` as a filter: the subdomain points `z` such that `f` and `g` agree at
+  every index of the block indexed by `z`. -/
+lemma complDisagreementSet_def :
   complDisagreementSet k φ f g =
     { z ∈ (φ.subdomain k).toFinset | ∀ i ∈ blockIdx φ k z, f i = g i  } := by
   aesop (add simp [complDisagreementSet, disagreementSet])
 
+/-- The agreement set sits inside the `k`th subdomain, so has at most `2 ^ (n - k)` points. -/
 @[simp]
 lemma card_complDisagreementSet_le :
   (complDisagreementSet k φ f g).card ≤ 2 ^ (n - k) := by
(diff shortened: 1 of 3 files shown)


alexanderlhicks:
actually no

PR #697, opened 2026-08-06
feat(Hachi): make extractors computable 

tobias-rothmann:
make extractors computable


ArkLib/OracleReduction/Security/CoordinateWiseSpecialSoundness/Guarded.lean
```diff
@@ -100,6 +100,67 @@ theorem IsGuarded.of_isPure (V : Verifier oSpec StmtIn StmtOut pSpec) (h : V.IsP
...
+**`verify_eq` is sorried**, and it is the *only* remaining obligation of the guarded seam:
+`Verifier.IsGuarded.append` is proved from it by forgetting the data.
+Proof plan: normalize `Verifier.append`'s bind with `failure_bind`/`pure_bind`
+under the two `if`-splits, mirroring `Verifier.PureForm.append`. -/
+def GuardedForm.append {V₁ : Verifier oSpec Stmt₁ Stmt₂ pSpec₁}
+    {V₂ : Verifier oSpec Stmt₂ Stmt₃ pSpec₂} (G₁ : V₁.GuardedForm) (G₂ : V₂.GuardedForm) :
+    (V₁.append V₂).GuardedForm where
+  check := fun stmt tr => G₁.check stmt tr.fst && G₂.check (G₁.out stmt tr.fst) tr.snd
+  out := fun stmt tr => G₂.out (G₁.out stmt tr.fst) tr.snd
+  verify_eq := by sorry
```
alexanderlhicks: essentially fixed by #729

alexanderlhicks:
## Coordination note: merge order and required changes

Reviewed alongside #730 and #729. To say the main thing first: **the interface change here is the right one**, and it fixes a defect I had not appreciated until I looked at the composition. On `main` the composed extractor is

```lean
extractor := fun stmt tree => L₁.extractor stmt tree.appendSplit.fst
```

which discards the right factor entirely, and the append theorems take the right factor's soundness as a bare existential — so the "chain of named certificates exposes a runnable end-to-end extractor" claim is not true today: the composite for the whole Hachi chain is row 1's extractor on the prefix tree, with every downstream link's extraction hidden inside its own `Exists.choose`. `Extractor.TreeBased.append out₁ E₁ E₂` makes it real. `ReduceClaim.treeExtractor` going from "classically pick any `witOut`; junk if none exists" to `fun stmtIn tree o => (o tree.onlyPath).map (mapWitInv stmtIn)` is the clearest illustration.

### Merge order

**#730 → #697 → #754.** #730 first for two reasons: it carries the `axiom_baseline.json` refresh that `main` needs after #729, and — more relevant here — its `endPiece` is what closes the chain into a *proof*. Your `NoChallenge` docstring stakes the design on closing factors ("what lets the whole composed extractor run as a function of `(stmtIn, tree)` alone"), and the zero-round components do have that shape. But `openingChain` ends in `handoffPackage`, which produces a `QuadEvalWitness Φ'` — the iteration is a *reduction*, so its composed extractor legitimately consumes a witnessing. Only `evaluation = iteration ▷ endPiece` closes. Landing #730 first means this PR arrives into a topology where its own story is realizable; it might be worth saying so in the `NoChallenge` docstring either way.

### Required

**1. `GuardedForm.append.verify_eq` is still `sorry` (`Guarded.lean:153`), and `IsGuarded.append` is now derived from it.** So a theorem that was a direct `sorry` on the old `main` is still tainted, and it propagates: `GCWSSPackage` carries `isGuarded : verifier.GuardedForm` as data, and its `append` builds the composite field with `L₁.isGuarded.append L₂.isGuarded`.

I measured this on this branch, before and after applying the fix:

| declaration | at HEAD | after |
| --- | --- | --- |
| `Verifier.GuardedForm.append` | `propext, sorryAx, Classical.choice, Quot.sound` | clean |
| `Verifier.IsGuarded.append` | `propext, sorryAx, …` | clean |
| `CoordinateWise.GCWSSPackage.append` | `propext, sorryAx, …` | clean |
| `CoordinateWise.EscapeGCWSSPackage.append` | `propext, sorryAx, …` | clean |
| `CoordinateWise.CWSSPackage.appendGuarded` | `propext, sorryAx, …` | clean |

Build stays green and the `Guarded.lean` sorry warning goes 1 → 0.

To be precise about what does *not* break: `verify_eq` is `Prop`-valued and erased at codegen — I confirmed `GuardedForm.append`, `TreeBased.append` and `GCWSSPackage.append` all still have IR at HEAD, so nothing stops computing. What it costs is the soundness trail: `sorryAx` under the `isCWSS` field of every guarded package.

**The proof is now on `main`** — #729 landed `IsGuarded.append` outright, and its tactic body transfers verbatim; only `check₁/out₁/hV₁` become `G₁.check/G₁.out/G₁.verify_eq`, and drop the `omit … SampleableType` line (this section does not bind that instance):

```lean
  verify_eq := fun stmt tr => by
    simp only [Verifier.append]
    rw [G₁.verify_eq stmt tr.fst]
    by_cases hc₁ : G₁.check stmt tr.fst = true
    · rw [if_pos hc₁, pure_bind, G₂.verify_eq (G₁.out stmt tr.fst) tr.snd]
      by_cases hc₂ : G₂.check (G₁.out stmt tr.fst) tr.snd = true <;> simp [hc₁, hc₂]
    · rw [if_neg hc₁]
      simp [hc₁]
```

**2. The extractor port is now forced by the rebase.** #729 landed three more old-shape `Extractor.TreeBased` occurrences (54 total on `main`) plus one live `onlyTranscript` consumer at `Sumcheck/FinalEval.lean:169`, which this PR's deletion breaks. The good news is that the port is natural: #729's `resp : Fin k → WitOut` family in `ScalarRound.treeExtractorScalarOfValid` — currently conjured with `Exists.choose` — **is** your leaf witnessing, and the `hmk` premise is unchanged (its `∀ j, (out s v (fam j), resp j) ∈ relOut` is `IsValid` at a guarded verifier, via `isValid_iff_pure`).

### Should

**3. No escape-threaded closer.** `treeSpecialSoundWith.mem_relIn_of_isAccepting` and `specialSound.exists_total_extractor` are the two theorems carrying the argument that nothing is lost, and both are stated for the **plain** notion — which has no Hachi consumers. Every Hachi certificate is escape-threaded. The missing statement compiles verbatim here, axiom-clean, six lines of proof (the two CWSS wrappers are one-line delegations, since `coordinateWiseSpecialSoundWith{,Escape}` are `def … := treeSpecialSoundWith{,Escape} … (toShape D)`):

```lean
theorem treeSpecialSoundWithEscape.escape_or_mem_relIn_of_isAccepting [Inhabited WitIn] … :
    esc stmtIn tree ∨
      (stmtIn, (Ext stmtIn tree (canonWitnesses init impl verifier relOut stmtIn)).getD default)
        ∈ relIn := by
  rcases h stmtIn tree hstr hacc with hesc | hgood
  · exact Or.inl hesc
  · obtain ⟨w, hw, hrel⟩ := hgood _ (canonWitnesses_isValid hacc)
    exact Or.inr (by simpa [hw] using hrel)
```

One framing suggestion while you are there: I would soften "the textbook statement survives". `canonWitnesses` is `if h : ∃ w, … then some h.choose else none`, so the closer plugs the choice function back in — it recovers `main`'s *non-algorithmic* reading, which is exactly the right migration receipt, but for a reduction the `∀ o valid` form is the better statement, not a wart. Worth saying, because it is the argument for the premise rather than an apology for it.

**4. Document the quantifier order.** The disjunction sits outside the witnessing quantifier (`esc ∨ ∀ o valid, …`, not `∀ o valid, esc ∨ …`). That is strictly stronger, it stops the escape decision depending on the supplied witnessing, and your composition proof depends on it — `append_treeSpecialSoundWithEscape_guardedLeft` resolves both escape disjuncts before `refine Or.inr fun o hvalid => ?_`. Nothing records that it is deliberate; a later "simplification" commuting them would weaken every certificate in the tree.

**5. The existential escape closures are now dead.** `treeSpecialSoundEscape` / `coordinateWiseSpecialSoundEscape` / the `OracleVerifier` variant had five consumers on the old `main` (right-factor hypotheses of every escape append) and have **zero** here, since every append now takes a named `E₂`. That is a good consequence of the redesign, but their docstrings still say "keep this form for plumbing" and there is no plumbing left that takes them. Delete or re-docstring.

**6. Off-scope deletion in `f71ce7e4 "clean-up"`.** It removes the ~65-line "Superseded: the one-round Kronecker-seed repair" section from `docs/kb/audits/noz26-zero-check-lemma10.md`, plus the pointer in `NOZ26.md`. That section is the quantitative record of why the route was rejected — `≈2^-102` vs `≈2^-123` at the Figure-9 parameters, `k | d/2 = 512` forcing `k = 8`, sumcheck cost `7.3 KB → 14.6 KB` — and it is explicitly labelled a record of a rejected design, so nothing in this PR makes it stale. Could it come back, trimmed if you like?

### Not a comment on this PR

`docs/skills/make-computable.md` is good, and I would promote one line from the pitfalls to the Goal section: *"ArkLib's `Extractor.TreeBased` is a plain function type with no complexity bound, so a computable extractor is no stronger a security statement than a classical one."* That is the honest framing, and it is easy for a reader to miss at the bottom.

alexanderlhicks (dismissed):

ErVinuelas (approved):

PR #715, opened 2026-08-11
fix(binius): encode witnesses over the Boolean cube, not the diagonal

alexanderlhicks:
## The bug

Three sites in the Binius formalization built the novel-polynomial-basis coefficient vector of a multilinear witness `t` with the spelling

```lean
fun ω => t.val.eval ω
```

where `ω : Fin (2^ℓ)`. This does **not** read `t`'s Boolean-hypercube table: Lean silently coerces `ω` to a single ring element and evaluates `t` at the *constant* point `fun _ => ↑ω` — the diagonal. Distinct witnesses collapse to identical coefficient tables (e.g. `X 0` and `X 1` become indistinguishable), so the encoded codeword did not determine the witness.

Affected sites (all on `main`):
- `Binius.BinaryBasefold.getMidCodewords`
- `Binius.BinaryBasefold.firstOracleWitnessConsistencyProp`
- `Binius.FRIBinius.CoreInteractionPhase.sumcheckFoldCtxLens` (honest-prover lens)

## The fix

New shared encoder in `BinaryBasefold/Basic.lean`:

```lean
def witnessNovelCoeffs (t : L⦃≤ 1⦄[X Fin ℓ]) : Fin (2^ℓ) → L :=
  fun ω => t.val.eval (fun j => (Nat.getBit j.val ω.val : L))
```

Coefficient `ω` is `t` evaluated at the Boolean cube point whose bits are the LSB-first binary digits of `ω` — the inverse of the index order used by the existing decoder (`hypercube_evals` via `Nat.binaryFinMapToNat`). All three consumers now use it.

A compiled regression probe (`example`) checks that the encoder distinguishes `X 0` from `X 1` at `ω = 1`, which the old diagonal spelling provably could not — so the bug cannot silently regress.

## Validation

- `lake build` of both affected modules on top of current `main` (`a4ac38e0e`): green, 3,709 jobs, only pre-existing baseline warnings.
- `#print axioms` on `witnessNovelCoeffs` and all three updated consumers: `[propext, Classical.choice, Quot.sound]` — no `sorryAx`.
- Full `./scripts/validate.sh` (build, imports, docs integrity, KB lint): passed.
- `lint-style` on the two changed files: identical diagnostic set to `main` (line-number shifts only); no new style findings.

This is the isolated extraction of the cube-table correction that was previously bundled inside draft #615; it is independent of the generic ring-switching work and is a prerequisite for rebasing that PR.

### DIFF
--- a/ArkLib/ProofSystem/Binius/BinaryBasefold/Basic.lean
+++ b/ArkLib/ProofSystem/Binius/BinaryBasefold/Basic.lean
@@ -792,10 +792,39 @@ def BBF_SumcheckMultiplierParam : SumcheckMultiplierParam L ℓ (SumcheckBaseCon
     degCombinator := 1
     combinator_natDegree_le := by intro _; exact Polynomial.natDegree_X_le }
 
+/-- The [DP24] coefficient vector of a multilinear witness `t`: its Boolean-hypercube table,
+read in the LSB-first bit order of `Nat.binaryFinMapToNat` (`ω = ∑ⱼ 2^j·wⱼ`, so coefficient `ω`
+is `t` at the point `w j = getBit j ω`). This is the inverse of the index mapping used by the
+novel-coefficients decoder above (`hypercube_evals`, via `Nat.binaryFinMapToNat`).
+
+This explicit map avoids the silent coercion in the old spelling `fun ω => t.val.eval ω`, which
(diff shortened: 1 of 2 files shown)


chung-thai-nguyen (approved):
Thank you @alexanderlhicks. This LGTM.

PR #717, opened 2026-08-12
RS-codes are list decodable

ElijahVlasov:
The Reed–Solomon code `RS[F, domain, m]` of rate `ρ` is `(1 - √ρ - η, 1/(2η√ρ))`-list decodable, for every `η > 0`.


ArkLib/Data/CodingTheory/ListDecodability.lean
```diff
@@ -80,10 +81,46 @@ The finiteness conjunct is necessary because `Set.ncard` assigns cardinality zer
...
+    have := Set.ncard_le_ncard hT hfin
+    rw [Set.ncard_coe_finset] at this
+    exact le_trans (by exact_mod_cast this) hcard
+  · intro h y
+    have hfin : (closeCodewordsRel C y r).Finite := finite_of_forall_finset_card_le (h y)
+    refine ⟨hfin, ?_⟩
+    have hsub : (hfin.toFinset : Set (ι → F)) ⊆ closeCodewordsRel C y r := by
+      simp [Set.Finite.coe_toFinset]
+    have hcard := h y hfin.toFinset hsub
+    rwa [Set.ncard_eq_toFinset_card _ hfin]
```
alexanderlhicks: **Delete this whole block — #722 already landed all of it on `main`, in generalised form.**

* `listDecodable'` → `Code.isListDecodable_iff_forall_finset_card_le` (`ListDecodability.lean:378`),
  with `Code.isListDecodable_of_forall_finset_card_le` (`:357`) as the `mpr` direction you actually
  use, and `Code.IsListDecodable.finset_card_le` (`:366`) as the `mp`.
  Note these take the hypothesis in the `∀ c ∈ T, c ∈ closeCodewordsRel C y r` spelling rather than
  `↑T ⊆ …`, which is exactly the shape `card_le_of_subset_closeCodewords` already wants — so no
  `simp only [Set.subset_def, Finset.mem_coe]` bridging is needed.
* `finite_of_forall_finset_card_le` → `Set.finite_of_forall_finset_card_le` in the new
  `ArkLib/ToMathlib/Set/Finite.lean`. Same proof, but stated over an arbitrary `FloorSemiring`
  rather than `ℝ`, and it needs no `0 ≤ ℓ` step (at a negative bound the `T = ∅` instance of the
  hypothesis is already unsatisfiable) — which removes the `have hℓ0 : 0 ≤ ℓ := by grind` here,
  whose provenance is not obvious.
* `listDecodable_iff_listDecodable'` → the same
  `Code.isListDecodable_iff_forall_finset_card_le`.

Also note `def listDecodable` on line 84 no longer exists at all: it is now
`Code.IsListDecodable (C) (r : ℝ) (ℓ : ℝ≥0)`, defined as `Code.Lambda C r ≤ ⌊ℓ⌋₊`, with
`Code.Lambda : ℕ∞` the primitive. So this file needs no changes from this PR whatsoever — dropping
it takes the PR from 6 files to 5.

ArkLib/Data/CodingTheory/ReedSolomon/ListDecodability.lean
```diff
@@ -0,0 +1,138 @@
...
+  {η : ℝ} (hη : 0 < η) :
+  listDecodable (ReedSolomon.code domain m : Set (ι → F))
+    (1 - (ReedSolomon.sqrtRate m domain : ℝ) - η)
+    (1 / (2 * η * (ReedSolomon.sqrtRate m domain : ℝ))) := by
+  rw [listDecodable_iff_listDecodable']
+  intro y T hT
+  exact card_le_of_subset_closeCodewords domain hm hη y T fun c hc ↦ hT hc
+
+end ReedSolomon
+
```
alexanderlhicks, line 138: **This is the part that will not compile post-#722.** Here is the version I built and verified
against current `main`:

```lean
/-- **Theorem 4.3.** The Reed–Solomon code `RS[F, domain, m]` of rate `ρ` is
`(1 - √ρ - η, 1/(2η√ρ))`-list decodable, for every `η > 0`.

[ACFY24] states this for `η ∈ (0, 1 - √ρ)`; the upper restriction is unnecessary, since above it
the radius `1 - √ρ - η` is negative and the point list is empty. -/
theorem isListDecodable_code [Nonempty ι] (domain : ι ↪ F) {m : ℕ} (hm : 0 < m)
    {η : ℝ≥0} (hη : 0 < η) :
    Code.IsListDecodable (ReedSolomon.code domain m : Set (ι → F))
      (1 - (ReedSolomon.sqrtRate m domain : ℝ) - (η : ℝ))
      (1 / (2 * η * ReedSolomon.sqrtRate m domain)) := by
  refine Code.isListDecodable_of_forall_finset_card_le fun y T hT => ?_
  rw [NNReal.coe_div, NNReal.coe_one, NNReal.coe_mul, NNReal.coe_mul, NNReal.coe_ofNat]
  exact card_le_of_subset_closeCodewords domain hm (by exact_mod_cast hη) y T hT
```

Four things changed, and one trap:

1. `listDecodable` → `Code.IsListDecodable`.
2. **The list-size argument is now `ℝ≥0`, not `ℝ`.** The cleanest fix is to take `η : ℝ≥0` too, so
   that `1 / (2 * η * sqrtRate m domain)` is an `ℝ≥0` term directly (`sqrtRate` is already `ℝ≥0`)
   and only the *radius* needs a coercion into `ℝ`. Taking `η : ℝ` instead forces a
   `Real.toNNReal` or an anonymous-constructor subtype term at every call site.
3. `rw [listDecodable_iff_listDecodable']; intro y T hT` → the single
   `Code.isListDecodable_of_forall_finset_card_le`, which hands you `∀ c ∈ T, c ∈ …` directly. The
   trailing `fun c hc ↦ hT hc` adapter disappears with it.
4. Name: `listDecodable_reedSolomon` stutters inside `namespace ReedSolomon`, and the predicate is
   now `IsListDecodable`, so `isListDecodable_code` fits the post-#722 convention (cf.
   `Code.isListDecodable_natCast_iff`, `CodingTheory.johnson_isListDecodable`).

⚠️ **Trap:** do not close this with `push_cast`. `ReedSolomon.sqrtRate` is a `noncomputable abbrev`,
so `push_cast` unfolds it and pushes the coercion through `NNReal.coe_sqrt`, leaving
`√↑↑(LinearCode.rate (code domain m))` on the right and a type mismatch against
`card_le_of_subset_closeCodewords`. The explicit `NNReal.coe_*` rewrite chain above stops at the
`sqrtRate` boundary and works.

`card_le_of_subset_closeCodewords` itself needs no proof changes at all — only
`closeCodewordsRel` → `Code.closeCodewordsRel` (see the `open` line above). I confirmed
`simp only [Code.mem_relHammingBall_iff, Code.relHammingDist_coe]` on line 53 still fires post-#722
despite the `open Classical` instance mismatch that `Code.mem_closeCodewordsRel_iff` warns about, so
that line can stay as-is.

ArkLib/Data/CodingTheory/ReedSolomon/ListDecodability.lean
```diff
@@ -0,0 +1,138 @@
...
+## References
+
+  * [Arnon, G., Chiesa, A., Fenzi, G., and Yogev, E., *WHIR: Reed–Solomon Proximity Testing
+      with Super-Fast Verification*][ACFY24]
+
+-/
+
+namespace ReedSolomon
+
+open Finset ListDecodable ReedSolomon
```
alexanderlhicks: `ListDecodable` is not a namespace any more — #722 moved everything into `Code`. And `ReedSolomon`
is redundant, since we are inside `namespace ReedSolomon` already.

I used:

```lean
open Finset
open scoped NNReal
```

and then qualified the four `Code.`-namespace uses explicitly (`Code.closeCodewordsRel`,
`Code.IsListDecodable`, `Code.isListDecodable_of_forall_finset_card_le`, plus the `Code.agree*`
lemmas which the proof body already qualifies). `open scoped NNReal` is needed for the `ℝ≥0`
notation in the theorem statement.

Prefer this over `open Code`: #722 flagged an `NNReal`-notation clash whose blast radius was
deliberately kept to `open`s *inside* `namespace Code`, and this file is not in that namespace.

ArkLib/Data/CodingTheory/ReedSolomon/ListDecodability.lean
```diff
@@ -0,0 +1,138 @@
...
+import ArkLib.Data.CodingTheory.ListDecodability
+import ArkLib.Data.CodingTheory.ReedSolomon
+
+/-!
+# The Johnson-type list-decoding bound for Reed–Solomon codes
+ .
+
+  This file proves that the Reed–Solomon code
+  `RS[F, L, m]` of rate `ρ` is `(1 - √ρ - η, 1/(2η√ρ))`-list decodable for every `η > 0`
+  (`RSListDecoding.listDecodable_reedSolomon`) which appears as theorem 4.3
```
alexanderlhicks, line 15: `RSListDecoding` is not a namespace in this file — the declaration lives in `ReedSolomon`. With the
rename suggested below this line should read:

```
  (`ReedSolomon.isListDecodable_code`) which appears as theorem 4.3
```

Worth getting right: `checkdecls` and the blueprint `lean_decls` list resolve these by full name.

Related, and *not* this PR's fault: #722's rename left `blueprint/lean_decls` stale on `main` —
lines 155/164/165 still read `ListDecodable.Code`, `ListDecodable.listDecodable`,
`ListDecodable.closeCodewordsRel`, none of which resolve any more. So expect a red `checkdecls`
inherited from `main` when you rebase; it is not something this PR introduced.

ArkLib/Data/CodingTheory/ReedSolomon/ListDecodability.lean
```diff
@@ -0,0 +1,138 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Ilia Vlasov, Aristotle (Harmonic)
+-/
+
+import ArkLib.Data.CodingTheory.ListDecodability
+import ArkLib.Data.CodingTheory.ReedSolomon
+
+/-!
+# The Johnson-type list-decoding bound for Reed–Solomon codes
+ .
```
alexanderlhicks: Stray ` .` left over on its own line in the module docstring — drop it.

ArkLib/Data/CodingTheory/Basic/Distance.lean
```diff
@@ -210,6 +211,78 @@ theorem mem_hammingBall_iff (y x : n → R) (r : ℕ) :
   x ∈ hammingBall y r ↔ Δ₀(y, x) ≤ r := by
   aesop (add simp [hammingBall, hammingDist])
 
+section Agreement
+
+variable {u v : n → R}
+
+/-- The number of positions at which the two words `u` and `v` agree. -/
+def agree (u v : n → R) : ℕ := ({i | u i = v i} : Finset _).card
```
alexanderlhicks, line 219: `Code.agree` is exactly the complement of the existing `Code.disagreementCols` (line 153 of this
file, ~66 lines above):

```lean
def disagreementCols (u v : n → R) : Finset n := Finset.filter (fun i => u i ≠ v i) Finset.univ

lemma hammingDist_eq_disagreementCols_card (u v : n → R) :
    hammingDist u v = (disagreementCols u v).card
```

Per ArkLib's no-duplication convention these two should be tied together rather than left as
independent filters over `univ`. Minimally, add a bridge next to `agree_add_hammingDist` (verified
to compile):

```lean
/-- `agree` is the complement of `disagreementCols`. -/
lemma agree_add_card_disagreementCols (u v : n → R) :
    agree u v + (disagreementCols u v).card = Fintype.card n := by
  rw [← hammingDist_eq_disagreementCols_card]; exact agree_add_hammingDist
```

(Stated additively rather than as `agree u v = (disagreementCols u v)ᶜ.card` because `n` carries
only `[Fintype n]` here, not `[DecidableEq n]`, so `Finset.compl` is not available without
strengthening the section variables.)

Two more notes on this block:

* `sq_sum_agree_le` is the same Cauchy–Schwarz-on-fibers argument as the `private`
  `cauchy_schwarz_fiber` in `ArkLib/Data/Probability/Combinatorial.lean:126` (collision counts of a
  map rather than agreement of codewords). Not reusable as-is because it is `private`, so a fresh
  proof here is fine — but a `See also` note in the docstring would save the next person the
  search.
* Heads-up on ordering: #723 (`ABF26 §3 list-decoding bounds`) is stacked on #722 and re-derives
  `agree` + `agree_add_hammingDist` inline inside the proof of
  `agreementWeight_ge_of_hammingDist_le` (`ListDecodability/Bounds/AgreementHypergraph.lean:90-97`).
  No top-level name clash, but whichever of the two lands second should switch to `Code.agree`.

Adding `import Mathlib.Algebra.Order.Chebyshev` for `sq_sum_le_card_mul_sum_sq` is fine —
`Distance.lean` goes from 981 to ~1054 lines, comfortably under the 1500-line long-file cap.

ArkLib/Data/CodingTheory/ReedSolomon.lean
```diff
@@ -243,6 +243,43 @@ lemma evalOnPoints_mem_code_of_natDegree_lt {α : ι ↪ F} {p : F[X]} (h_deg :
   evalOnPoints α p ∈ code α n :=
   mem_code_of_polynomial_of_natDegree_lt_of_eval p h_deg (by simp [evalOnPoints])
 
+/-- Two distinct Reed–Solomon codewords of degree `< m` agree in fewer than `m` positions. -/
+lemma agree_lt_of_mem_code {F : Type*} [Fintype ι] [Field F] [DecidableEq F]
```
alexanderlhicks: This is correct as written and I verified it compiles unchanged, but it re-does root counting that
`main` already has. `ReedSolomon.minDist_of_le` (line 431 of this file) gives
`minDist (code α n) = |ι| - n + 1` for `n ≤ |ι|` — and `minDist_eq_card_sub_min_add_1` (line 486)
gives the unconditional form. Combined with `Code.agree_add_hammingDist`, `agree c c' < m` for
distinct codewords is a corollary rather than a fresh induction on `(p - q).roots`.

If the derivation from `minDist` turns out to need side conditions this one doesn't (`[Inhabited ι]`
/ `[NeZero n]` appear in those statements), keeping the direct proof is defensible — but then a
docstring line saying *why* it isn't derived from `minDist_of_le` would be worth having, since the
duplication is otherwise going to look accidental.

ArkLib/Data/CodingTheory/ReedSolomon.lean
```diff
@@ -384,6 +421,48 @@ lemma dist_le_length [DecidableEq F] (inj : Function.Injective α) :
...
+  · rw [div_le_one (by exact_mod_cast h)]
+    exact_mod_cast min_le_right _ _
+
+private lemma le_one_of_sq_le_one {x : ℝ} (hx : 0 ≤ x) (h : x ^ 2 ≤ 1) : x ≤ 1 := by
+  nlinarith
+
+@[simp]
+lemma sqrtRate_le_one [Fintype ι] (m : ℕ) (domain : ι ↪ F) :
+  (ReedSolomon.sqrtRate m domain : ℝ) ≤ 1 :=
+  le_one_of_sq_le_one (sqrtRate_nonneg m domain) (sqrtRate_sq_le_one m domain)
```
alexanderlhicks: These five `sqrtRate` lemmas are a genuinely useful addition — `main` has none of them, and
`sqrtRate` is currently either unfolded via `simp [ReedSolomon.sqrtRate]` (many sites in
`DivergenceOfSets.lean`) or re-proved inline. Two inline proofs this PR could retire directly:

* `ProximityGap/BCIKS20/AffineSpaces.lean:1445` proves `sqrtRate deg domain ≤ 1` inline →
  `sqrtRate_le_one`
* `ProximityGap/BCIKS20/AffineSpaces.lean:1749` proves `0 < sqrtRate 1 domain` inline →
  `sqrtRate_pos`

Not blocking for this PR, but worth either doing here or filing, so the new lemmas actually become
the single source of truth rather than a sixth spelling.

Minor: `le_one_of_sq_le_one` duplicates `abs_le_one_iff_mul_self_le_one` / `pow_le_one_iff_of_nonneg`
from Mathlib — `sqrtRate_le_one` can be
`pow_le_one_iff_of_nonneg (sqrtRate_nonneg m domain) two_ne_zero |>.mp (sqrtRate_sq_le_one m domain)`,
dropping the private helper.

alexanderlhicks (commented):
## Review: statement is correct; the proof needs a rebase onto #722

I checked the statement against [ACFY24] Thm 4.3 and then rebased the whole PR onto current `main`
(which now contains #722) locally. **Everything builds, and it is axiom-clean** —
`ReedSolomon.isListDecodable_code`, `card_le_of_subset_closeCodewords`, `agree_lt_of_mem_code` and
`Code.sq_sum_agree_le` all report only `[propext, Classical.choice, Quot.sound]`. The inline
comments below are the exact deltas I needed; I have deliberately *not* pushed them.

### 1. Statement review — correct, and slightly stronger than the paper

The counting argument is right. Writing `s = √ρ`, `L = |T|`, `n = |ι|`:
`agree(c,y) ≥ (s+η)n`, `agree(c,c') ≤ s²n` for distinct codewords, and Cauchy–Schwarz
`(Σ_c agree(c,y))² ≤ n · Σ_{c,c'} agree(c,c')` give `L(s+η)² ≤ 1+(L-1)s²`, i.e.
`L(2sη+η²) ≤ 1-s²`, hence `L ≤ 1/(2ηs)`. That is exactly Thm 4.3.

Two deviations from the paper, both fine:

* **The paper requires `η ∈ (0, 1-√ρ)`; this PR only requires `0 < η`.** That is a genuine
  strengthening and it is true: for `η ≥ 1-√ρ` the radius `1-√ρ-η` is `≤ 0`, so the point list is
  empty or `{y}`. Worth saying in the docstring, since a reader diffing against the paper will
  stop on it.
* **After #722, `IsListDecodable C r ℓ` is *by definition* `Lambda C r ≤ ⌊ℓ⌋₊`.** So the rebased
  statement claims the *floored* bound, which is stronger than the paper's real-valued one. It
  still holds: `⌊1/(2η√ρ)⌋₊ = 0` forces `η > 1/(2√ρ)`, and `1 - s - 1/(2s) < 0` for every
  `s ∈ (0,1]` (discriminant of `2s²-2s+1` is `-4`), so the radius is already negative there and the
  point list is empty. No strength is lost by flooring.

`min m |ι|` in `sqrtRate_sq` also handles `m > |ι|` correctly (`s = 1`, radius negative).

### 2. This supersedes an existing theorem on `main`

`main` already has `CodingTheory.rs_lambda_le_johnson_mds` (`JohnsonBound/Family.lean:924`):

```
Lambda (ReedSolomon.code dom n) (1 - √ρ - η) ≤ ENNReal.ofReal (1 / (2 * η * ρ))
```

That is the same radius with the **weaker** list size `1/(2ηρ)` instead of `1/(2η√ρ)` (weaker since
`ρ ≤ 1 ⟹ ρ ≤ √ρ`). So this PR is a real improvement rather than a duplicate — and it additionally
drops the `[Finite F]` hypothesis, matching the paper's remark that the bound is field-size
independent.

I verified the implication compiles:

```lean
theorem supersedes {ι : Type*} [Fintype ι] [Nonempty ι] {F : Type*} [Field F]
    {n : ℕ} [NeZero n] (dom : ι ↪ F) (η : ℝ) (hη_pos : 0 < η) :
    let ρ : ℝ := (Module.finrank F (ReedSolomon.code dom n) : ℝ) / Fintype.card ι
    (Lambda ((ReedSolomon.code dom n : Set (ι → F))) (1 - Real.sqrt ρ - η) : ENNReal) ≤
      ENNReal.ofReal (1 / (2 * η * ρ)) := by
  classical
  intro ρ
  have hn : 0 < n := Nat.pos_of_ne_zero (NeZero.ne n)
  have hρdef : ρ = ((ReedSolomon.sqrtRate n dom : ℝ)) ^ 2 := by
    rw [ReedSolomon.sqrtRate_sq]
    show (Module.finrank F (ReedSolomon.code dom n) : ℝ) / (Fintype.card ι : ℝ) = _
    congr 1
    exact_mod_cast ReedSolomon.dim_eq_min_deg_card (α := dom)
  have hspos : 0 < (ReedSolomon.sqrtRate n dom : ℝ) := ReedSolomon.sqrtRate_pos hn (domain := dom)
  have hs1 : (ReedSolomon.sqrtRate n dom : ℝ) ≤ 1 := ReedSolomon.sqrtRate_le_one n dom
  have hrate : Real.sqrt ρ = ((ReedSolomon.sqrtRate n dom : ℝ)) := by
    rw [hρdef, Real.sqrt_sq hspos.le]
  have hηc : ((η.toNNReal : ℝ≥0) : ℝ) = η := Real.coe_toNNReal η hη_pos.le
  have h717 := ReedSolomon.isListDecodable_code (ι := ι) (F := F) dom hn
    (η := η.toNNReal) (by simpa using hη_pos)
  rw [Code.isListDecodable_iff_toENNReal_le_ofReal, hηc] at h717
  rw [hrate]
  refine h717.trans (ENNReal.ofReal_le_ofReal ?_)
  rw [NNReal.coe_div, NNReal.coe_one, NNReal.coe_mul, NNReal.coe_mul, NNReal.coe_ofNat, hηc, hρdef]
  apply one_div_le_one_div_of_le (by positivity)
  nlinarith [mul_nonneg (mul_pos hη_pos hspos).le (sub_nonneg.mpr hs1)]
```

Suggestion (non-blocking, could be a follow-up): re-derive `rs_lambda_le_johnson_mds` from this
theorem rather than from `mds_johnson_lambda_le`, and cross-reference the two in the docstrings so
nobody reaches for the weaker one by accident.

### 3. The rebase

`main` no longer has `def listDecodable`. #722 replaced it with `Code.Lambda : ℕ∞` as the primitive
and `Code.IsListDecodable (C) (r : ℝ) (ℓ : ℝ≥0) : Prop := Lambda C r ≤ ⌊ℓ⌋₊`. The good news is that
**everything you added to `ListDecodability.lean` already exists on `main` in a better form**, so
that file drops out of the diff entirely (6 files → 5, −39 lines). Details inline.

Net rebased diff, all green:

```
 ArkLib.lean                                        |   1 +
 ArkLib/Data/CodingTheory/Basic/Distance.lean       |  73 +++++++++++
 ArkLib/Data/CodingTheory/Basic/RelativeDistance.lean |   5 +
 ArkLib/Data/CodingTheory/ReedSolomon.lean          |  81 +++++++++++-
 ArkLib/Data/CodingTheory/ReedSolomon/ListDecodability.lean | 140 ++++++++++++++
```

alexanderlhicks (approved):

PR #701, opened 2026-08-07
feat(coding-theory): ABF26 foundations and code families [split 2/4]

alexanderlhicks:
# feat(coding-theory): ABF26 foundations and code families [split 2/4]

Second of four PRs superseding #505. This is the reusable coding-theory foundation for the later
proximity-gap and toy-problem splits: probability/combinatorial support, distance/list primitives,
Johnson bounds, folded/interleaved/multiplicity Reed–Solomon codes, subspace designs, and
extension codes.

No proximity-bound theorem, toy protocol, or prize/leaderboard claim is in scope.

### Probability, distance, and list primitives

- Probability helpers are consolidated under `Probability`; the Schwartz–Zippel result is derived
  from the pre-existing arbitrary-sampling-set theorem rather than duplicated.
- `Probability/Combinatorial.lean` proves ABF26 Claim B.1. Its reusable core is
  universe-polymorphic, assumes only finite `S`, permits arbitrary `T`, and does not expose an
  unnecessary `DecidableEq T`; the `Pr_` spelling is a compatibility wrapper.
- `qEntropy`, Hamming-ball volume, disagreement-coordinate bridges, relative minimum-distance
  bridges, alphabet-normalized linear/module rate, and Singleton/MDS bridges cover the relevant
  §2 foundations.
- `ListDecodable.Lambda` uses `Set.encard`: infinite point lists contribute `top`, not the old
  accidental `ncard = 0`. Finite and numeric bridges are explicit.
- The former erasure-correction existential API was removed after review found it tautological.
  The meaningful generic metric-uniqueness theorem remains; no arbitrary function is labeled a
  decoder.

### Johnson-bound family

- `Jcap`/`Jqell`, Plotkin support, and the alphabet-generic ABF26 T3.2
  `johnson_bound_lambda_le_ell` are proved, including the radicand-negative corner internally.
- `mds_johnson_lambda_le_of_rate_distance` proves the exact C3.3 conclusion for an arbitrary
  finite alphabet and arbitrary code set from an explicit normalized rate and MDS rate-distance
  equation. Field-linear and Reed–Solomon statements are thin wrappers.
- Accidental universe-0 constraints in Johnson/MDS/RS APIs are removed. The RS wrapper exposes
  only `Nonempty`; `Inhabited` is constructed locally.

### ABF26 §2 code families

- Folded RS: admissible domains, evaluation/code, RS transport, exact saturated dimension,
  block-metric MDS distance, and the `s = 1` bridge.
[...]

PR #722, opened 2026-08-12
refactor(coding-theory): make the list size primitive, list-decodability notation

alexanderlhicks:
## Why

`ListDecodable.listDecodable` was a second definition whose entire content is the inequality `Lambda C r ≤ ℓ`. Having two notions of *"the point list is small"* is what let one of them drift: the `∀ y, ncard ≤ ℓ` body is satisfied by an **infinite** point list, since `Set.ncard` reports `0` there. Closing that took an explicit `Finite` conjunct plus **five bridge lemmas** to move between the two spellings at four numeric types.

This makes `Lambda : ℕ∞` the primitive quantity and `listDecodable` a `def` whose body *is* `Lambda C r ≤ ⌊ℓ⌋₊`.

Three consequences:

- **Finiteness becomes a theorem, not an obligation** (`finite_closeCodewordsRel_of_Lambda_le`). `encard` sends an infinite list to `⊤`, so `Lambda ≤ n` cannot be established by an infinite list at all — the bug class is unrepresentable rather than caught by review.
- **The bound moves `ℝ → ℝ≥0`.** A negative list-size bound is meaningless, and admitting one split the readings: `⌊ℓ⌋₊ = 0` says every point list is empty, while `(ncard : ℝ) ≤ ℓ < 0` is unsatisfiable. That disagreement is why the removed bridges carried `0 ≤ ℓ` side conditions; on `ℝ≥0` both vanish. Measured, not assumed: **all nine call sites already use `ℝ≥0`** — three STIR here, three WHIR and three more arriving with later splits — so the carrier change edits nothing anywhere.
- **The `∀`/`ncard` view survives as a lemma**, so it cannot drift. Five bridges *between two definitions* are replaced by facts about one object.

## Scope

**In:** the list-size layer and its only two Lean consumers.

| File | Change |
|---|---|
| `Data/CodingTheory/ListDecodability.lean` | the refactor; 6 lemmas removed, 9 added |
| `Data/CodingTheory/JohnsonBound/Family.lean` | `johnson_listDecodable` / `_of_le` gain `ℝ≥0` conclusions; `_of_le` becomes a one-liner via `.mono` |
| `docs/wiki/coding-theory-conventions.md` | durable rationale, plus two wrong turns recorded |
| `docs/kb/papers/Joh62.md`, `docs/kb/audits/open-problems-…md` | drop references to the deleted bridges and the "two shapes" framing |

**Out:** STIR and WHIR are untouched — their six hypotheses elaborate unchanged. No `sorry` filled, added, or removed. No other coding-theory file touched. Generalising the radius `r` to `ℝ≥0` is *not* attempted: `coding-theory-conventions.md` sanctions `ℝ` for the radius in existing API, and changing it would touch `Lambda` and every radius call site.

## `def`, not `abbrev` — the one non-obvious choice

As an `abbrev` this is reducible, and then Mathlib's `@[simp] ge_iff_le` — an `Iff.rfl` on a bare `GE.ge` — unifies with the whole `≤`-shaped body, so **`simp` silently unfolds it all the way to `WithBot.LE` on `ℕ∞`**. That mangles goals *and* makes `listDecodable.mono` unreachable through dot notation. This is specific to the shape: an `abbrev` whose body is a `∀` is unaffected. Since the proofs that will consume these hypotheses are `simp`-heavy, semireducibility is load-bearing. `exact`/`refine`/`apply` still unify through it at default transparency, so no bridge lemma is needed; `listDecodable_iff_Lambda_le` and `listDecodable_natCast_iff` exist for `rw`/`simp only`, and both have consumers.

## Verification

- `lake build ArkLib` green at 4219 jobs; `./scripts/validate.sh` green **including the `ArkLib/Data` zero-warning gate**; no lint hit on any line this PR touches; all new declarations axiom-clean (`[propext, Classical.choice, Quot.sound]`).
- **The new predicate is provably the same proposition as the one it replaces** — checked against a verbatim copy of the old body, including at the degenerate bound `0`, over an empty word space, and in the infinite-list direction. So no existing hypothesis changed meaning.
- Flooring at the definition is sharp; no generality lost by `r : ℝ≥0` in `johnson_listDecodable_of_le`; `uniqueDecodable` still means "at most one close codeword".
- **The shape is exercised at its first real consumer, not merely instantiated.** STIR Lemma 4.5.1 (`out_of_dom_smpl_1`) needs the point list as a `Finset` with a card bound; extracting that is three lines, and the only delta from the superseded definition is one application of `listDecodable_iff_forall_ncard_le`. Its union-bound step also settles a pre-existing question about STIR's own statement favourably: the `ℝ≥0` truncated `l - 1` is not a weakening, because `l < 1` forces the point list empty.
- Two adversarial passes, one independent. The independent pass found the `abbrev`/`simp` interaction above — a regression in the first draft — plus that a helper duplicated Mathlib's `@[norm_cast] Nonneg.nat_floor_coe`, and two docstring inaccuracies. All applied.

## Known costs, recorded in the docstring

`rw` cannot see through a `def` (nor can `simp`, which is the point), and consuming proofs pay one extra lemma application. `omega` is no help on `ℕ∞` either way.

## Sequencing

Please land this **before** #717 (*RS-codes are list decodable*). #717's bound is a real expression (`1/(2η√ρ)` with `η : ℝ`), so this change touches its **statement**, not just its proof; landing it in between means adapting it twice. Its other required adaptation — its proof currently discharges the infinite-point-list branch with `Set.Infinite.ncard`, which #701 already closed — is one line via `listDecodable_of_forall_finset_card_le`, at its own generality with no finiteness on `F`.


ArkLib/Data/CodingTheory/ListDecodability.lean
```diff
@@ -59,218 +72,379 @@ section
...
+The bound is `ℝ≥0`-valued because the theorems that consume such a hypothesis reuse the same bound
+as a number in their conclusions. Flooring loses nothing, `Lambda` being integer-valued
+(`isListDecodable_iff_forall_ncard_le`), and point-list finiteness is implied rather than asserted.
+
+Kept a `def` rather than an `abbrev`: as an `abbrev` it is reducible, and `simp` then unfolds the
+`≤`-shaped body through `ge_iff_le` down to `WithBot.LE`, which mangles goals and makes
+`IsListDecodable.mono` unreachable by dot notation. `exact`, `refine` and `apply` still see through
+to the `Lambda` inequality, so no bridge lemma is needed to prove one; `simp` and `rw` need
+`isListDecodable_iff_Lambda_le`. -/
+def IsListDecodable (C : Set (ι → F)) (r : ℝ) (ℓ : ℝ≥0) : Prop :=
```
ElijahVlasov, line 260: Love the new name!

PR #730, opened 2026-08-13
feat(Hachi): update overview file (and some GuardedComposition lemmas)

tobias-rothmann:


ErVinuelas (dismissed):
We talked about it on Thursday, I think it is a good summary of what is left to be done. :)

alexanderlhicks (approved):

PR #778, opened 2026-08-19
Make Constrained RS codes not depend on `Smooth`

ElijahVlasov:


ArkLib/Data/CodingTheory/ReedSolomon.lean
```diff
@@ -803,60 +802,5 @@ class Smooth
...
 
-variable {F : Type*} [Field F] [DecidableEq F]
-        {ι : Type*} [Fintype ι] [DecidableEq ι]
-        {domain : ι ↪ F} [Smooth domain]
-        {m : ℕ}
-
-/-- Definition 4.2, WHIR[ACFY24]
-Smooth Reed-Solomon codes are Reed-Solomon codes defined over smooth domains, such that their
-decoded univariate polynomials are of degree less than `2ᵐ` for some `m ∈ ℕ`. -/
-noncomputable def smoothCode
```
quangvdao, line 814: **[P2] Preserve or deliberately migrate this public declaration.** Removing `smoothCode` is broader than decoupling `constrainedCode` from the `Smooth` class, and it immediately breaks the blueprint's `\lean{ReedSolomon.smoothCode}` declaration check. The smallest repair is to retain/move it (and consider a deprecated compatibility alias if its replacement changes type); otherwise update the blueprint/docs and describe the migration explicitly.

quangvdao (changes requested):
Thanks, Ilia — moving the constrained-code definitions out of the generic `ReedSolomon` module and onto the concrete smooth FFT domain is a sensible architectural direction.

I reviewed exact head `4089b8b46e6f0298bfb6391e85e99c52ba06102f` against its recorded base `d57c8e11855c2fd528b8521f4e3b90dd2d7a64c9`, and checked it against current upstream `main` at `5a9626d331d713c7e74efa7e3d93b7d0ec9c4dc2`.

## Verdict

Requesting changes. The definitions themselves correspond sensibly to WHIR Definitions 4.5 and 4.6, but the PR currently breaks a documented public declaration and fails required checks.

## Merge blockers

1. **[P2] Public API / blueprint regression.** The move deletes `ReedSolomon.smoothCode` (and also `mVdecode`) while `blueprint/src/coding_theory/defs.tex` still declares `ReedSolomon.smoothCode`. The blueprint checker therefore fails with `ReedSolomon.smoothCode is missing.` Please either:
   - retain/move these declarations, using deprecated compatibility aliases if the new API should replace them; or
   - make the removal explicit, update the blueprint and documentation, and provide a migration path for the old public API.

   Since the stated goal is only to remove the `Smooth` dependency from constrained codes, preserving the unrelated public names is the smallest repair.

2. **Required hygiene.** The whitespace job identifies trailing spaces on lines 61–62 of `Constrained.lean` (and `git diff --check` also reports the blank line at EOF). The new module also needs the repository-required module docstring with a `## References` section and a properly formatted `[ACFY24]` citation.

3. **API transition.** The old constrained-code API accepted an arbitrary embedding with `[Smooth domain]`; the replacement accepts only `SmoothCosetFftDomain n F`. That may be the intended boundary, but it is a public narrowing. Please either supply a compatibility bridge or explain the deliberate break and update affected documentation/consumers.

## Nonblocking source/coverage request

`rs_code_is_multi_constrained` uses (t=0), so the result follows from an empty family of constraints. If this is intended to capture the WHIR observation that an ordinary smooth RS code is constrained by the zero weight/target, please add the explicit one-zero-constraint theorem (or explain why the empty-family formulation is the desired library statement). That would be a more falsifiable source bridge than the vacuous `Fin 0` case.

## Validation

- The exact head completes the Lean build and normal validation path; imports and docs-integrity checks pass.
- The combined build workflow fails at blueprint declaration checking, and whitespace fails as above.
- No new `sorry`, `admit`, unsafe declaration, or axiom was found.
- The branch is behind current `main`, but a synthetic merge-tree check is currently clean.

Please also use a conventional title such as `refactor(coding-theory): decouple constrained RS codes from Smooth` and expand the PR body with motivation, compatibility impact, and source correspondence. Thanks again — this should be a compact repair once the API decision is made.

PR #808, opened 2026-08-27
fix(Hachi): commit the ring-switch quotient as base-b digits - #807

ErVinuelas:
**Problem.** Hachi's security rests on an Ajtai (lattice) commitment, whose binding is only meaningful for short openings. One piece of the committed witness — the quotient polynomial ρ produced by the ring-switching step — is not short for honest provers: the only provable bound on it is the trivial q/2. Since the protocol range-checks the whole committed vector against a single base, that one large block dragged every norm parameter up to q/2. At those values the range check and the norm ball accept everything in ZMod q, so a "short collision" of the commitment carried no norm information and the Module-SIS reduction target was trivially solvable: the composed soundness statement was vacuous at every parameter set the correctness theorem accepted.

**Fix.** Follow the paper's actual protocol ([NOZ26] §4.3): don't commit ρ itself, commit its balanced base-b digit decomposition (new module RingSwitch/RhoDigits.lean). Digits of any value are ⌊b/2⌋-bounded by construction, so the committed vector is short unconditionally, at the cost of widening it from μ + n to μ + n·δ entries (δ = clog_b q). The verifier recombines the digits publicly via ρ(α) = ∑ᵤ bᵘ·ρᵤ(α). No new rounds or messages; the shortness predicate collapses from two bounds to one.


scripts/HachiRuntime.lean
```diff
@@ -117,9 +121,10 @@ def pp : Hachi.PublicParamsD 𝓜(Q, A) IR (2 ^ Mm) Dg OR (2 ^ Rr) Dg DR where
   outerMatrix := fun _ j => rr (j.val + 2) 1
   dMatrix := fun _ j => rr 1 (j.val + 1)
 
-/-- The lift commitment's own key: `dRows × (μ₀ + n₀)`, the width a whole lifted witness needs
-(see the note above `hachiLiftCom`). -/
-def dMat : Ajtai.Simple.PublicParams 𝓜(Q, A) DR (Mu + Nn) :=
+/-- The lift commitment's own key: `dRows × (μ₀ + n₀·δ)`, the width a whole lifted witness needs
+once the quotient block is committed as its base-`bZero` digits (see the note above
+`hachiLiftCom`). `δ = clog_{bZero} Q`. -/
+def dMat : Ajtai.Simple.PublicParams 𝓜(Q, A) DR (Mu + Nn * rhoDigitCount Q P.bZero) :=
```
dhsorens, line 133: **[bug]** After the gadget encoding, the committed table is `(μ₀ + n₀·δ)·d` wide. At the advertised toy parameters (`q = 7`, `bZero = 3`, `δ = clog_3 7 = 2`, `μ₀ = 8`, `n₀ = 5`, `d = 1`) that is `8 + 5·2 = 18`. `MM` is still `3` (unchanged in this PR), so the sumcheck/zero-check cube has `2^{M+1} = 16` points. The last two table rows (both digits of the last quotient row) sit at indices 16 and 17 and are silently dropped: `wTable` returns `0` off-cube, `H₀` never range-checks them, and the public `M̃_α` contraction misses `ρ_4`. This `dMat` width is the new arity; the module header still describes the *old* pin (`bZero = 4`, `(μ₀ + n₀)·d = 13 ≤ 16`). Default `validate.sh` only runs the millisecond checks, which evaluate `wTableMleEval` against itself on the truncated cube, so CI can stay green while the only executable instance is not a model of the theorems (`hμn` is `18 ≤ 16`).

**Suggestion:** Bump `MM` so `2^{M+1} ≥ 18` (e.g. `MM = 4` → cube 32), recompute the `--full` cost, and rewrite the module docstring (the `bZero = ⌊q/2⌋ + 1` / `13 ≤ 16` paragraph) to match `P` (`γ = 2`, `bZero = 3`) and the widened arity. Confirm `--full` still accepts after the bump.

ArkLib/Commitments/Functional/Hachi/Composition.lean
```diff
@@ -281,15 +281,16 @@ The certificate `iteration.isCWSS` is the one-iteration CWSS statement — every
...
+    (K : LiftCom (LiftedWitness 𝓜(q, α) μ₀ n₀) (liftShort 𝓜(q, α) γ bDig))
     (pp : Hachi.PublicParamsD 𝓜(q, α) innerRows (2 ^ m) messageDigits outerRows (2 ^ r)
       innerDigits dRows)
     (φF : ZMod q →+* F)
     (hd : 0 < (𝓜(q, α)).φ.natDegree) (hb : b - 1 ≤ γ)
-    (hbpos : 0 < b)
-    (hρ : b - 1 ≤ ρBound) (hcov : (μ₀ + n₀) * (𝓜(q, α)).φ.natDegree ≤ 2 ^ m₀)
+    (hbpos : 0 < b) (hb1 : 1 < b)
+    (hdig : DigitBaseOk q γ bDig)
+    (hcov : (μ₀ + n₀ * rhoDigitCount q b) * (𝓜(q, α)).φ.natDegree ≤ 2 ^ m₀)
```
dhsorens, line 292: **[suggestion]** `iteration` takes independent `b` (QuadEval / range / `wTable` digit base) and `bDig` (`liftShort` / `hachiLiftCom` digit base), then sizes the cube with `rhoDigitCount q b`. `liftMessage` and `Concrete.nonrecursiveLiftCom` are `μ + n·clog_{bDig} q` wide. `HonestRangeParams` also keeps `b` and `bZero` as separate fields (the default witnesses set them equal). If a caller instantiates `iteration` with `b ≠ bDig`, the range-checked table is a *different* digit encoding than the committed vector, `hcov` no longer matches the Ajtai key width, and recombination uses weights `b^u` for the wrong base. Before this PR `ρBound` was only a numeric bound, so the split was harmless; now both parameters are encodings.

**Suggestion:** Index `hcov`, `batchPackage`, `wTable`, and recombination by `bDig` (or add a hypothesis `b = bDig` / share one base through the chain). The completeness assembly already uses `P.bZero` consistently for the table and the lift; the soundness `iteration` should match.

ArkLib/Commitments/Functional/Hachi/Correctness.lean
```diff
@@ -615,10 +616,11 @@ theorem hachiNonrecursive_perfectCorrectness
...
         (dRows := dRows) (m := m) (r := r) P.b)).run s))
     (hclog : 0 < Nat.clog P.b q) (hd : 0 < 𝓜(q, α).φ.natDegree) (hbZero : 0 < P.bZero)
-    (K : LiftCom (LiftedWitness 𝓜(q, α) (μ₀ P) n₀) (liftShort 𝓜(q, α) P.γ (q / 2)))
+    (K : LiftCom (LiftedWitness 𝓜(q, α) (μ₀ P) n₀) (liftShort 𝓜(q, α) P.γ P.bZero))
     [DecidableEq K.TCom]
-    (φF : ZMod q →+* F) (hμn : ((μ₀ P) + n₀) * 𝓜(q, α).φ.natDegree ≤ 2 ^ (M + 1))
-    (hZeroγ : P.bZero - 1 ≤ P.γ) (hZeroρ : P.bZero - 1 ≤ q / 2) :
+    (φF : ZMod q →+* F)
+    (hμn : ((μ₀ P) + n₀ * rhoDigitCount q P.bZero) * 𝓜(q, α).φ.natDegree ≤ 2 ^ (M + 1))
+    (hZeroγ : P.bZero - 1 ≤ P.γ) :
```
dhsorens, line 625: **[suggestion]** The theorem docstring above still says the reverse range orientations “pin `P.γ = q/2 = P.bZero − 1`”. That was the vacuity this PR exists to remove. The theorem now takes `hZeroγ : P.bZero - 1 ≤ P.γ` together with `DigitBaseOk`, which `pinned_of_soundness_orientations` proves lands at `γ = bZero − 1 < q/2`. Leaving the old pin in the correctness theorem header undercuts the PR’s claim.

**Suggestion:** Replace the pin with `γ = P.bZero − 1 < q/2` (and mention `ofPinnedDigitBase`).

ArkLib/Commitments/Functional/Hachi/RingSwitch/Completeness.lean
```diff
@@ -242,7 +243,8 @@ theorem rhoShort_honestLiftWitness_half (hd : 0 < Φ.φ.natDegree) (s : RlinStat
   rhoShort_half _
 
 omit [NeZero q] in
-/-- **Perfect completeness of Hachi's lift from an honest seam, at `ρBound := q/2`** — error `0`,
+/-- **Perfect completeness of Hachi's lift from an honest seam, at the digit base `bDig`** —
+error `0`,
 and with **no admissibility hypothesis**: the `ρ`-half of `liftShort` is discharged outright
 (`rhoShort_honestLiftWitness_half`), so the only premises are the three facts the seam relation
```
dhsorens: **[suggestion]** `liftReduction_perfectCompleteness_of_zShort` is documented as having “no admissibility hypothesis” with the `ρ`-half discharged by `rhoShort_honestLiftWitness_half`. The theorem actually takes `1 < bDig`, `bDig ≤ q/2`, `bDig / 2 ≤ bound` and discharges the digit half by `rhoDigitsShort_of_half_le`. The old `q/2` lemma is still in the file for the *raw* quotient and is not what completeness uses.

**Suggestion:** Cite `rhoDigitsShort_of_half_le` / `DigitBaseOk` and drop “no admissibility hypothesis”.

ArkLib/Commitments/Functional/Hachi/HonestChain.lean
```diff
@@ -184,21 +240,23 @@ theorem rlinReduction_perfectCompleteness_params (P : HonestRangeParams q)
...
-    (liftReduction (oSpec := oSpec) (F := F) Φ P.γ (q / 2) K hd).perfectCompleteness init impl
-      (relRlinImage (zDigits := zDigits) Φ pp base ω P.γ) (relLift Φ P.γ (q / 2) K φF) :=
-  liftReduction_perfectCompleteness_image Φ K φF init impl hd pp base ω
+    (liftReduction (oSpec := oSpec) (F := F) Φ P.γ P.bZero K hd).perfectCompleteness init impl
+      (relRlinImage (zDigits := zDigits) Φ pp base ω P.γ) (relLift Φ P.γ P.bZero K φF) :=
+  liftReduction_perfectCompleteness_image Φ P.hbZero P.hbZeroq P.hbZeroγ K φF init impl hd
+    pp base ω
 
 omit [NeZero q] in
 /-- **Seam 3 — the lift's output feeds the batching bridge**, at the bundled `bZero`. The bridge's
```
dhsorens, line 262: **[suggestion]** Seam 3’s docstring still says `relLift γ (q/2)` is *literally* the batching bridge’s input relation. The theorem immediately below is `relLift Φ P.γ P.bZero` on both sides. Harmless for Lean, but it is exactly the degeneracy the gadget encoding is supposed to have deleted.

**Suggestion:** Write `relLift γ bZero`.

dhsorens (commented):
here are some initial notes we can go through in a minute

dhsorens (dismissed):
Merging, thanks @ErVinuelas !

dhsorens (approved):
Re-approving

PR #847, opened 2026-09-03
refactor(hachi): decouple τ from δ and drop the unsigned-digit path

ErVinuelas:
Two changes to Hachi:

**Balanced digits only.** The unsigned digit convention and its norm theory were unreachable from
the committer, which was already balanced. They are removed; norm bounds are proved once for an
arbitrary digit map and instantiated at the balanced digits. `commitBalanced` becomes `commit`.

**τ decoupled from δ, and set to 5.** The folded witness `z = Σᵢ cᵢ sᵢ` is deterministically short
but was decomposed at the message width `δ`, because the only decomposition interface demanded
reconstruction of every residue. A new interface makes reconstruction conditional on a shortness
bound, so `τ` is sized from that bound rather than from `q`. A new module pins [NOZ26]'s Figure 9
profile at `τ = 5`, proves the arithmetic hypotheses used by the chain, and instantiates the
QuadEval completeness and soundness results at the same radius `γ = 15`. `τ = 5` follows the
paper's §4.4 deterministic rule. Figure 9 instead lists `τ = 4` and `z = 30583`; the latter equals
four balanced base-16 digits' capacity, but the paper does not derive it as a deterministic norm
bound or analyze the corresponding abort probability. Docstrings and `docs/kb/papers/NOZ26.md`
record this discrepancy without inferring how the table entry was obtained.

The recursive tail and the existing generic `Reduction.append_completeness` admission remain
outside this PR. The new bounded-decomposition and profile completeness results are axiom-clean.

Validation at `694b47afa`: `./scripts/validate.sh --axioms` passed, including the compiled fast
runtime checks, source policy, documentation integrity, and axiom-regression gate.


quangvdao (approved):
Thanks @ErVinuelas for this work! Separating the folded-witness width from the message width is a useful step toward a concrete Hachi instantiation, and the balanced-digit cleanup makes the development easier to follow.

Reviewed at 694b47aface43154ee513423c5191c1e94362a62. We found and fixed three issues in that follow-up commit:

- The concrete completeness theorem used radius 16 while the soundness package used 15. Completeness now uses `params.γ = 15`, so the profile relations match.
- The documentation attributed a digit-range restriction to `DigitDecomposition` that its abstract type does not impose. The impossibility claim now explicitly concerns digits restricted to the balanced box.
- The explanation of Figure 9 inferred too much from `30583` equaling four digits’ capacity. It now distinguishes that numerical identity from a proved norm bound, while retaining the verified observation that §4.4’s deterministic rule yields `τ = 5`.

No remaining mathematical blocker found in this PR. `./scripts/validate.sh --axioms` passed locally on the reviewed revision, including the default compiled runtime checks and the axiom-regression gate; the new bounded-decomposition and profile results use only the standard foundational axioms. The optional full toy opening was stopped after ten minutes without a verdict, so that run remains unverified. Hosted build CI is still running.

The existing generic append-completeness admission, recursive tail, and final scheme-level security bridge remain separate follow-up work. Approving this parameter and API repair, subject to the remaining CI and normal merge queue checks.

PR #848, opened 2026-09-03
feat(dsfs): prove Theorems 6.1 and 6.2

chung-thai-nguyen:
## Summary

This PR formalizes the security layer needed for the duplex-sponge Fiat--Shamir transformation and proves the paper's Theorems 6.1 and 6.2.

- Define adaptive, query-bounded NARG soundness and straightline knowledge soundness, including the corresponding state-restoration notions and coin-bearing experiments.
- Prove the canonical single-salt Fiat--Shamir soundness and straightline knowledge-soundness results (Theorems 3.18 and 3.19).
- Prove DSFS soundness (Theorem 6.1).
- Define the DSFS straightline extractor (Construction 6.3) and prove DSFS straightline knowledge soundness (Theorem 6.2).
- Document the definitions, reductions, and explicit Section 5/Section 6 proof boundary in the blueprint.

## Authors

- Chung Thai Nguyen
- Michele Orrù
- Yuxi Zheng

## Proof boundary

The Section 6 reductions take an explicit `KeyLemmaSecurityWitness`, which packages the Section 5 endpoint statistical-distance and query-bound guarantees. Consequently, the Section 6 theorem bodies and the canonical Fiat--Shamir security proofs are complete; proving that the concrete Section 5 transforms produce this witness remains the separate Key Lemma obligation.

## Stacked PR

This PR is intentionally based on `dsfs-section5` while #469 is open. Once #469 merges, this PR will be retargeted to `main` automatically due to Github Stack; its content is limited to the security definitions and Section 6 results built on that foundation.

## Verification

- Ran `lake build` successfully.
- Checked every new blueprint `\lean{...}` declaration against Lean.
- Built the printable blueprint with LuaLaTeX and resolved the CO25 citation.


ArkLib/OracleReduction/Security/Basic.lean
```diff
@@ -405,6 +405,279 @@ class Extractor.Straightline.IsMonotone
...
+    (init : ProbComp σ) (impl : QueryImpl oSpec (StateT σ ProbComp))
+    (auxImplE : QueryImpl auxSpecE ProbComp)
+    (verifier : NonInteractiveVerifier Proof oSpec StmtIn StmtOut)
+    (relIn : Set (StmtIn × WitIn)) (langOut : Set StmtOut)
+    (bound : OracleComp oSpec (StmtIn × Proof) → Prop) (error : ENNReal) : Prop :=
+  ∃ extractor : StmtIn → Proof → QueryLog oSpec → QueryLog oSpec →
+      OptionT (OracleComp auxSpecE) WitIn,
+  ∀ P : OracleComp oSpec (StmtIn × Proof), bound P →
+    Pr[ nargKSFailEvent relIn langOut
+      | adaptiveNARGKnowledgeSoundnessExp init impl auxImplE verifier extractor P ] ≤ error
```
quangvdao: **[P1] The advertised straightline KS contract differs materially from CO25**

Owner: #848. `ArkLib/OracleReduction/Security/Basic.lean:510–572`, coin counterpart 662–677; `DuplexSponge/Security/Soundness.lean:254–297`; `KnowledgeSoundness.lean:696–721`.

The definition labels itself CO25 Definition 3.6, but its extractor takes both prover and verifier logs and returns `OptionT (OracleComp auxSpecE) WitIn`. The final DSFS theorem provides a live uniform sampler, and its concrete trace transform invokes LookAhead, which samples missing rate blocks. The predicate also chooses an extractor after fixing each bound/error, so its theorem family does not package one extractor uniformly across query budgets. The SR premise has the same existential order.

The formalized claim is a pointwise, randomized, verifier-trace-assisted knowledge-soundness variant. It does not establish the paper's stated deterministic, uniformly chosen extractor contract. This is a claim/API mismatch, not a failed Lean proof. CO25 itself needs reconciliation: Definition 3.6 and the straightline discussion require determinism and prover-trace inputs, while Construction 6.3 supplies the verifier trace and D2STrace uses randomized LookAhead. Do not silently equate those interfaces.

Repair: expose fixed-extractor NARG/SR predicates, transport a supplied SR extractor through `fsSRDelegatingExtractor` and the DSFS wrapper, and quantify that value before the budget family. Then either prove the required determinism/input-reconstruction bridge or explicitly name/document the weaker notion and narrow the paper-theorem claim. Running-time and rewinding guarantees are separately outside scope.

ArkLib/ToVCVio/Tactic/VCVNorm.lean
```diff
@@ -0,0 +1,255 @@
...
+  unless e.isAppOf ``Bind.bind && e.getAppNumArgs == 6 do return none
+  let args := e.getAppArgs
+  let lhs := args[4]!
+  let cont := args[5]!
+  let some sq := lhs.find? (fun s =>
+    s.isAppOf ``simulateQ &&
+      s.getAppArgs.any (fun a => a.getAppFn.constName? == some ``OracleSpec.loggingOracle)) |
+    return none
+  let sqArgs := sq.getAppArgs
+  return some (sqArgs[sqArgs.size - 1]!, cont)
```
quangvdao: **[P2] `vcv_strip_log` overmatches nested logged computations**

Owner: #848. `ArkLib/ToVCVio/Tactic/VCVNorm.lean:142–169` searches recursively inside the bind's left operand with `lhs.find?`, then constructs a proof as if that operand were exactly a logged run. A surrounding map therefore receives the wrong equality proof.

Reproduced on exact head:

```lean
import ArkLib.ToVCVio.Tactic.VCVNorm
open OracleComp OracleSpec
example {ι : Type} {spec : OracleSpec ι} {α : Type}
    (oa : OracleComp spec α) :
    ((id <$> (simulateQ loggingOracle oa).run) >>= fun p => pure p.1) = oa := by
  vcv_strip_log
  simp
```

Lean rejects the generated term with a kernel application type mismatch. Control proof passes: `simp only [id_map]; rw [loggingOracle.run_simulateQ_bind_fst]; simp`. The kernel catches the malformed proof; this does not enlarge the trust boundary. An additional higher-universe probe reaches a separate application mismatch because the supporting lemma is Type-0-only.

Repair: recognize the exact supported logged-run shape, normalize only justified wrappers, and decline unsupported shapes/universes. Add focused positive, wrapped-input, used-log refusal and universe canaries.

quangvdao (changes requested):
Thanks @chung-thai-nguyen for the substantial work on this stack and for addressing the earlier review. The concrete transforms, query-bound plumbing, and explicit Section 5/6 proof boundary are valuable progress.

I reviewed #469 at 7613419e6a662eca4a2f277ba56a147ee9b26d3c together with #848 at d4b27f0b41aafa3b7dcdd933a73a991b6b9ccbe7. I am requesting changes for the knowledge-soundness contract mismatch, the hybrid state handoff, and the reproducible logging-tactic failure. The issues introduced here are inline; the inherited Section 5 findings are included below so the stack review stays together.

**[P2, inherited from #469] The hybrid game discards the prover's simulator state**

Owner: #469, inherited by #848. [KeyLemma.lean:344–355](https://github.com/Verified-zkEVM/ArkLib/blob/d4b27f0b41aafa3b7dcdd933a73a991b6b9ccbe7/ArkLib/OracleReduction/FiatShamir/DuplexSponge/Security/KeyLemma.lean#L344) discards the outer `D2SQueryState`, retains only inner `memo₁`, and invokes `d2fRaw` for the verifier. `ProverTransform.lean:931` initializes that outer state with `default` on each invocation.

This resets hash/permutation tables, trace, and synthesis cache between prover and verifier, whereas CO25's hybrids use one D2SQuery simulator over their combined execution. The existing bad-event experiment correctly carries `s₁` into the verifier (`BadEvents/Lemma5_8.lean:149–160`). Consequently the currently defined hybrids are not the shared-state experiment used by the named paper proof or the existing bad-event bridge.

Repair: let the raw runner accept an initial outer simulator state; retain both the outer state and inner memo returned by the prover and seed the verifier with both. Verify cache continuity and the precise experiment correspondence. Confidence limit: no final-output TV-distance counterexample was established; raw oracle inconsistency alone does not establish one because the hybrid outputs project/transform logs. The probability claims remain admitted.


**[P2, nonblocking API hardening, inherited from #469] The sampled-oracle abstraction does not enforce fixed deterministic answers**

Owner: #469. [OracleDistribution.lean:86–94](https://github.com/Verified-zkEVM/ArkLib/blob/d4b27f0b41aafa3b7dcdd933a73a991b6b9ccbe7/ArkLib/OracleReduction/Security/OracleDistribution.lean#L86) calls itself a distribution over deterministic interpretations, but its field is `Carrier → QueryImpl spec ProbComp`. A checked constructor with `Carrier := Unit`, `sample := pure ()`, and `toImpl := fun _ _ => uniformSample Bool` inhabits this type and resamples on every identical query. Thus sampling a carrier does not establish the documented fixed-oracle invariant.

The inspected concrete DSFS/function-table/permutation constructors do use fixed tables; no current security theorem was shown invalid by this gap. Treat this as API hardening before generic clients rely on the documented invariant. Repair by storing deterministic answer functions/`QueryImpl spec Id`, adding an explicit fixedness law, or accurately documenting a broader distribution over randomized stateless handlers and stating deterministic assumptions separately.


Smaller nonblocking follow-ups: correct the single-pass/linear-work claim at SingleSalt.lean:563–565 (challengesOfLog scans afresh per requested round); qualify the claim at DeriveTranscript.lean:20 about arbitrary logs containing a canonical log, since an earlier conflicting answer wins; and update the relevant docs/wiki module-routing documentation.

Validation: full ./scripts/validate.sh --axioms passed on the exact upper head. Direct axiom checks of the final canonical-FS and DSFS security theorems show only propext, Classical.choice and Quot.sound. Both frozen PR heads have green hosted CI. The Section 5 hybrid claims/key lemma and single-salt completeness remain admitted; KeyLemmaSecurityWitness explicitly exposes that conditional boundary. These disclosed obligations are not additional hidden-sorry findings.

The earlier automated WitOut warning does not apply to the current proof-only NARG game: there is no output witness to discard, and relation-valued/SR interfaces retain theirs.

Source checked: https://eprint.iacr.org/2025/536.pdf, March 27, 2026 version.

ArkLib/ToVCVio/Tactic/VCVNorm.lean
chung-thai-nguyen: [DONE] I removed vcv_strip_log

ArkLib/OracleReduction/Security/Basic.lean
chung-thai-nguyen: [DONE] The extractor (of `Verifier.adaptiveNARGKnowledgeSoundness(WithCoins)`) now is fixed before the selection of the prover bound, so it matches Definition 3.6 of CO25 more than before. Caveat: we don't claim 100% match with Definition 3.6, because the Construction 6.3 requires the extractor to sample values inside D2STrace - which is not deterministic, and our `adaptiveNARGKnowledgeSoundness(WithCoins)` extractor has the 3 properties of randomized, non-rewinding, and accepting two logs. We consider the current definition is likely reasonably compatible with the paper, and will be iterated upon later.

Issue #854, opened 2026-09-04
feat(coding-theory): formalize low-rate capacity-achieving Reed–Solomon list decoding


quangvdao:
# Coordination issue: formalize low-rate capacity-achieving Reed–Solomon list decoding

> **This is the source-of-truth issue for the project.** Please claim a work package in a comment before starting, link every child issue and PR back here, and keep the dashboard below current. The intended unit of contribution is one narrow package with a frozen public interface, not one long-lived mega-branch.

## Summary

Formalize Brakensiek–Chen–Putterman–Zhang–Zheng, *Algorithmic List Decoding of Reed–Solomon Codes up to Capacity in the Low-Rate Regime* ([ECCC TR26-164](https://eccc.weizmann.ac.il/report/2026/164/), [PDF](https://eccc.weizmann.ac.il/report/2026/164/download/)).

The paper gives a deterministic polynomial-time list decoder for every Reed–Solomon evaluation set over a prime field in the low constant-rate regime. Its new ingredient is a multivariate interpolant constrained by the values of hidden Hasse derivatives. A local rank bound makes the interpolation system small enough; [Kopparty's differential-equation root finder](https://theoryofcomputing.org/articles/v011a005/) turns the interpolant into a bounded candidate list.

This is a large formalization with four genuinely independent technical trunks:

1. Hasse–Taylor algebra and structured multivariate substitution;
2. interpolation spaces and finite-dimensional linear algebra;
3. anisotropic simplex/lattice estimates and the local-rank argument;
4. Kopparty's differential-equation solver and its exact list bound.

The project should be parallel by design. The common theorem signatures and variable encodings will be frozen first; after that, contributors should own disjoint modules and merge through small, axiom-clean PRs.

## Source audit blockers — resolve before freezing the headline theorem

The source is not ready to transcribe literally. Phase P0 must produce an auditable repaired statement.

### 1. Proposition 3.13 is not proved under its stated rate hypothesis

The proposition assumes only

\[
  k/n \le (1-\theta)\varepsilon,
\]

but its proof says “Since we set \(k=\lfloor(1-\theta)\varepsilon n\rfloor\)” and uses the resulting **lower** bound on \((k-1)/n\). Theorem 4.1 then invokes the proposition for arbitrary smaller \(k\). An upper bound on the rate cannot justify that step.

Proposed repair: introduce an ambient design dimension

\[
  K := \lfloor(1-\theta)\varepsilon n\rfloor,
\]

[...]

PR #855, opened 2026-09-04
feat(coding-theory): define the target theorem for low-rate Reed–Solomon list decoding

quangvdao:
**Superseded scaffold.** The proved reusable exact-list and candidate-filter specification is now proposed in #908, tracked by #907. That replacement proves the two filtering lemmas that were admitted here and omits the obsolete abstract geometric contracts. It does **not** prove this PR's BCPZZ26 low-rate headline theorem: that historical mathematical scope and its source audit remain tracked in #854. This PR is being closed without merging its admissions or axiom-baseline changes. The original proposal is preserved below.

---

## Summary

This PR defines—but does not yet prove—the Lean target for formalizing Brakensiek, Chen, Putterman, Zhang, and Zheng's result on list decoding low-rate Reed–Solomon codes up to capacity ([BCPZZ26](https://eccc.weizmann.ac.il/report/2026/164/)).

Its purpose is to give a large, parallel formalization effort one agreed statement and one set of interfaces before contributors begin proving separate parts of the argument. In particular, it establishes:

- the exact list-decoder behavior that the finished formalization must provide;
- a stable boundary between interpolation, polynomial root finding, and final decoder construction;
- the paper's main theorem as a visible Lean guidepost with all quantitative hypotheses stated explicitly; and
- a source audit and blueprint recording ambiguities that must not be copied into downstream proofs.

The full project plan and work-package dashboard live in #854.

## Why this PR must come first

The paper's proof naturally splits into several large, mostly independent developments: Hasse-derivative algebra, multivariate interpolation, lattice and rank estimates, and a differential-equation root solver. Those parts can be developed by different contributors only if they agree in advance on what data crosses each boundary.

There is also a concrete source-level risk. The main theorem allows every message dimension `k` satisfying

```text
k / n ≤ (1 - θ) ε,
```

while one interpolation proof uses equality with the largest permitted dimension. A literal transcription would therefore prove the wrong statement for smaller `k`. This PR makes the repair structural: perform interpolation and root finding in a larger design space of degree `< K`, then filter the result back to the requested message space of degree `< k`, with an explicit assumption `k ≤ K`.

Keeping that repair—and all other algorithmic parameters—in the shared interface lets later work improve constants, list bounds, or parameter regimes without rewriting the decoder theorem.

## The target result, in plain language

For a prime field `ZMod q`, `n` distinct evaluation points, and parameters satisfying the paper's low-rate hypotheses, the final theorem says that every received word has a finite list containing **exactly** the degree-`< k` polynomials that agree with it in at least `⌈εn⌉` positions. The list has size at most

```text
q ^ (4 * ⌈ε ^ (-3 / θ)⌉ + 6).
```

The theorem also exports this result through ArkLib's existing `Code.IsListDecodable` definition at relative radius `1 - ε`. The source's runtime claim is deliberately not included: ArkLib does not yet have a cost model capable of expressing and checking it.

The source-facing Lean statement is [`exists_decoderCertificate_of_low_rate`](ArkLib/Data/CodingTheory/ReedSolomon/LowRateListDecoding/Main.lean).

## How the pieces fit together

```text
received word
    │
    ▼
interpolation produces a valid equation
    │
    ▼
root solver lists all degree-<K solutions
    │
    ▼
filter to degree < k and check actual agreement
    │
    ▼
exact decoder list, with the solver's cardinality bound
```

The three reusable interfaces are:

1. `InterpolationContract`: sufficiently agreeing degree-`< K` polynomials satisfy the equation produced from the received word.
2. `RootSolverContract`: for every valid equation, the solver returns exactly its degree-`< K` polynomial solutions and proves a caller-supplied list bound.
3. `DecoderCertificate`: after filtering to degree `< k`, membership is equivalent to meeting the agreement threshold, and the output list respects the same bound.

The central conditional theorem, `HiddenDerivative.exists_decoderCertificate_of_contracts`, composes these interfaces. It does not hard-code the paper's real-valued parameter choices or the particular bound `q^(4d+6)`. Consequently, a better root solver can supply a better bound without changing interpolation or decoder integration.

## What lands in this PR

| Area | Before | After this PR |
| --- | --- | --- |
| Public target | The result existed only in the paper and coordination issue | A checked Lean theorem signature states the exact paper-facing hypotheses and conclusion |
| Decoder semantics | ArkLib had its general list-decodability predicate | `DecoderCertificate` names the actual polynomial list and requires exact membership plus a uniform size bound |
| Message dimensions | The paper's `k`/`K` mismatch was unresolved in code | `messageDim` and `designDim` are separate, with an explicit embedding and final filter |
| Proof decomposition | Interpolation and root solving had no shared boundary | Ordinary Lean structures specify what each independent proof must provide |
| Parameters | The paper's asymptotic choices risked leaking into every proof | The core consumes exact natural parameters; rates and named choices are later specializations |
| Auditability | Source repairs were discussion items | The blueprint and knowledge-base audit record the repairs, trust boundary, and downstream obligations |

The implementation is organized as follows:

- [`ListDecoding/Specification.lean`](ArkLib/Data/CodingTheory/ReedSolomon/ListDecoding/Specification.lean) defines exact decoder and ambient-candidate certificates, the `k ≤ K` embedding, and the concrete final filter.
- [`HiddenDerivative/Parameters.lean`](ArkLib/Data/CodingTheory/ReedSolomon/HiddenDerivative/Parameters.lean) keeps `K`, agreement, derivative order, multiplicity, and support caps as independent natural numbers.
- [`HiddenDerivative/Contracts.lean`](ArkLib/Data/CodingTheory/ReedSolomon/HiddenDerivative/Contracts.lean) defines the interpolation/root-solver boundary and their conditional composition.
- [`LowRateListDecoding/Main.lean`](ArkLib/Data/CodingTheory/ReedSolomon/LowRateListDecoding/Main.lean) states the paper's explicit low-rate corollary.
- [`low_rate_rs_list_decoding.tex`](blueprint/src/coding_theory/low_rate_rs_list_decoding.tex) is the normative formalization blueprint; `content.tex` and `references.bib` register it and its sources.
- [`bcpzz26-main-theorem.md`](docs/kb/audits/bcpzz26-main-theorem.md), the BCPZZ26/Kopparty paper notes, their source metadata, and `docs/kb/index.md` preserve the clause-by-clause source audit.
- `ArkLib.lean` exposes the new modules, and `scripts/axiom_baseline.json` makes the intentional proof gaps visible to the repository's axiom regression check.

## Source issues recorded by the blueprint

This PR does not silently strengthen or repair the paper. It records the points that later proofs must discharge:

- **Message dimension:** decode in the ambient dimension `K` used by interpolation, then filter to every requested `k ≤ K`.
- **Lattice constant:** retain the exact additive `5/4` term hidden by an `O_θ(1)` step in the paper.
- **Ceiling arithmetic:** replace the false inequality `⌈x⌉ ≤ x` used in a support estimate with the corrected inequality documented in the blueprint.
- **Possible stronger regimes:** keep parameters general enough to express improved results, but do not advertise the candidate fixed-rate strengthening as proved.

## Current proof status and trust boundary

This is a statement and interface PR, not the completed formalization. Three declarations intentionally contain `sorry`:

- `CandidateCertificate.filteredDecoder_isExact`;
- `CandidateCertificate.filteredDecoder_card_le`; and
- `exists_decoderCertificate_of_low_rate`.

The first two are the generic correctness and cardinality obligations for the concrete final filter. The third is the paper's complete headline theorem. `CandidateCertificate.toDecoderCertificate` and the central conditional decoder theorem inherit the first two obligations.

The composition from interpolation plus root solving to an ambient candidate certificate is implemented without a new axiom. The axiom baseline records every intentional dependency so that later PRs must preserve or reduce this ledger rather than introduce hidden assumptions.

Still deferred are the interpolation proof, the differential-equation solver, the paper-parameter arithmetic, the final agreement/distance bridge proof, an executable decoder, and any machine-checked runtime theorem.

## Compatibility

This PR is additive. It does not change existing Reed–Solomon code semantics, serialization, or runtime behavior. The final theorem deliberately reuses ArkLib's existing `Code.IsListDecodable` API rather than introducing a competing public definition.

## Diff metadata

- Base: `22dbd4e836c15a21f68889afa69b7130da04abbb` (`main`)
- Head: `c7e9e01edf95b99adc30b76cb6bfcd2b59367c30`
- Commits: 1
- Files changed: 15
- Diff: 955 insertions, 2 deletions
- Normative blueprint: [`blueprint/src/coding_theory/low_rate_rs_list_decoding.tex`](blueprint/src/coding_theory/low_rate_rs_list_decoding.tex)

## Validation at `c7e9e01e`

Passed locally:

- `lake build ArkLib` (4,220 jobs);
- theorem-type and axiom audit over 11,174 declarations in 435 modules, with no non-standard axioms;
- knowledge-base lint;
- axiom-sweep fixture matrix and baseline regression check; and
- whitespace/error-marker check with `git diff --check`.

At the time of this description update, GitHub's library-import, documentation-integrity, and PR-summary checks are green. The main CI build is still running.

## Reviewer guide

The main question is whether these are the right statements to let the formalization safely fan out—not whether the paper has already been proved.

Suggested reading order:

1. Review the public theorem in `LowRateListDecoding/Main.lean` against BCPZZ26.
2. Check that `ListDecoding/Specification.lean` describes the actual output list we ultimately want, including both soundness and completeness.
3. Check that the `k`/`K` separation and the contracts in `HiddenDerivative/` permit independent and improved interpolation/root-solving results.
4. Review the blueprint's source repairs and deferred runtime boundary.
5. Confirm that the three intentional `sorry`s—and no stronger claim—are accurately represented in the axiom ledger.

Refs #854


ArkLib/Data/CodingTheory/ReedSolomon/ListDecoding/Specification.lean
```diff
@@ -0,0 +1,154 @@
...
+This is a deliberately exposed formalization target: downstream decoder work should refine this
+theorem rather than restating the `messageDim`/`designDim` repair. -/
+theorem CandidateCertificate.filteredDecoder_isExact {F ι : Type*} [Semiring F]
+    [DecidableEq F] [Fintype ι]
+    {domain : ι ↪ F} {designDim minAgreement listBound messageDim : ℕ}
+    (certificate : CandidateCertificate domain designDim minAgreement listBound)
+    (h : messageDim ≤ designDim) :
+    IsExactDecoder domain messageDim minAgreement
+      (certificate.filteredDecoder messageDim) := by
+  sorry
```
alexanderlhicks, line 130: [Blocking: compiled replacement proof] This generic filtering lemma can be proved now. Together with the cardinality proof below, it removes `sorryAx` from these two lemmas, `toDecoderCertificate`, and `exists_decoderCertificate_of_contracts`. I checked the replacement at `c7e9e01e` under normal project limits and rebuilt all three affected modules.

Replace this `sorry` with:

```suggestion
  intro received p
  constructor
  · intro hp
    exact (Finset.mem_filter.mp hp).2
  · intro hp
    apply Finset.mem_filter.mpr
    refine ⟨Finset.mem_preimage.mpr ?_, hp⟩
    let pK : MessagePolynomial F designDim :=
      ⟨p, Polynomial.degreeLT_mono h p.2⟩
    have hpK_agreement :
        minAgreement ≤ Code.agree (ReedSolomon.evalOnPoints domain pK) received := by
      change minAgreement ≤ Code.agree (ReedSolomon.evalOnPoints domain p) received
      exact hp
    have hpK_mem := certificate.complete received pK hpK_agreement
    change (p : Polynomial F) ∈ certificate.candidates received
    change (p : Polynomial F) ∈ certificate.candidates received at hpK_mem
    exact hpK_mem
```

Constructing the ambient subtype `pK` explicitly avoids a deterministic kernel timeout encountered in an alternative proof that relied on reduction through `messagePolynomialEmbedding`. This proof uses the existing `degreeLT_mono` API and works over the current general semiring assumptions.

ArkLib/Data/CodingTheory/ReedSolomon/ListDecoding/Specification.lean
```diff
@@ -0,0 +1,154 @@
...
+
+/-- Filtering a candidate list does not increase its cardinality. -/
+theorem CandidateCertificate.filteredDecoder_card_le {F ι : Type*} [Semiring F]
+    [DecidableEq F] [Fintype ι]
+    {domain : ι ↪ F} {designDim minAgreement listBound : ℕ}
+    (certificate : CandidateCertificate domain designDim minAgreement listBound)
+    (messageDim : ℕ) :
+    ∀ received, (certificate.filteredDecoder messageDim received).card ≤ listBound := by
+  sorry
```
alexanderlhicks, line 139: [Blocking: compiled replacement proof] The bound follows by filtering, taking the preimage under an embedding, and applying the candidate certificate's bound. This also compiles at the reviewed head under normal limits:

```suggestion
  intro received
  classical
  calc
    (certificate.filteredDecoder messageDim received).card ≤
        ((certificate.candidates received).preimage (messagePolynomialValue messageDim)
          (messagePolynomialValue messageDim).injective.injOn).card :=
      Finset.card_filter_le _ _
    _ = ((certificate.candidates received).filter fun p =>
        p ∈ Set.range (messagePolynomialValue messageDim)).card :=
      Finset.card_preimage _ _ _
    _ ≤ (certificate.candidates received).card := Finset.card_filter_le _ _
    _ ≤ listBound := certificate.card_le received
```

After both filtering proofs are applied, `#print axioms` reports no `sorryAx` for either lemma, `toDecoderCertificate`, or `exists_decoderCertificate_of_contracts`; only standard `propext`, `Classical.choice`, and `Quot.sound` remain. The paper-level theorem still needs its separate proof.

ArkLib/Data/CodingTheory/ReedSolomon/HiddenDerivative/Parameters.lean
```diff
@@ -0,0 +1,43 @@
...
+  /-- Ambient polynomial degree bound used for interpolation and root finding. -/
+  designDim : Nat
+  /-- Absolute number of required agreement positions. -/
+  minAgreement : Nat
+  /-- Highest Hasse-derivative order appearing in the interpolant. -/
+  derivOrder : Nat
+  /-- Root multiplicity imposed at every agreement point. -/
+  multiplicity : Nat
+  /-- Degree cap for the distinguished low-derivative variable. -/
+  yDegreeCap : Nat
```
alexanderlhicks, line 35: [P0 interface correction] The blueprint identifies this field with the paper's scalar `B`, but `B` bounds the **joint** exponent `b₀+b₁+|c|`, rather than one distinguished low-derivative variable. The support constraints are:

```text
b₀ + b₁ + |c| ≤ B
b₁ ≤ m
ω(c) ≤ W
|c| ≤ C
```

Please rename this to something like `totalJetDegreeCap`, document its joint meaning, and update the blueprint mapping. The separate strict interpolation degree budget is `a + (K-1)(b₀+b₁+|c|) < mA`. The issue's formal-target/P14 parameter lists also need `C` (already represented here by `l1Cap`). This distinction matters when P6 derives both coordinate and weighted degree bounds.

ArkLib/Data/CodingTheory/ReedSolomon/HiddenDerivative/Contracts.lean
```diff
@@ -0,0 +1,113 @@
...
+
+/-- The exact interface supplied by the interpolation proof.
+
+`Valid` should contain the structural facts needed by the chosen root solver, such as nonzeroness,
+weighted-degree bounds, and characteristic side conditions. Those facts are intentionally not
+encoded as rates or asymptotic estimates here. -/
+structure InterpolationContract {F ι Equation : Type*} [Semiring F] [DecidableEq F]
+    [Fintype ι] (domain : ι ↪ F) (params : Parameters)
+    (Valid : Equation → Prop)
+    (Satisfies : Equation → MessagePolynomial F params.designDim → Prop) where
```
alexanderlhicks, line 42: [P0 integration boundary] These generic contracts are useful, but P0 also needs a shared concrete equation type, Hasse-specialization relation, and validity predicate for P11/P12. Otherwise independently proved instances can choose incompatible representations and fail to compose.

For illustration, this canary compiled in `ReedSolomon.HiddenDerivative` against the current imports:

<details>
<summary>A tautological interpolation instance admitted by the generic interface</summary>

```lean
def reviewToyParams : Parameters where
  designDim := 1
  minAgreement := 0
  derivOrder := 0
  multiplicity := 0
  yDegreeCap := 0
  weightCap := 0
  l1Cap := 0

def reviewToyDomain : Fin 1 ↪ ℕ where
  toFun _ := 0
  inj' _ _ _ := Subsingleton.elim _ _

def reviewTautologicalInterpolation :
    InterpolationContract reviewToyDomain reviewToyParams
      (fun _ : Unit => True) (fun _ _ => True) where
  equation _ := ()
  valid _ := trivial
  satisfies_of_agreement _ _ _ := trivial
```

</details>

This is not a logical soundness bug: decoder composition still requires a matching bounded root solver. It shows that the differential-equation representation remains unfrozen. Please retain this generic adapter and add the concrete paper-facing layer, coordinating with the [completed P4b variable/substitution work](https://github.com/Verified-zkEVM/ArkLib/issues/854#issuecomment-5544374017). It can be supplied by that dependency rather than duplicated here; P0 should explicitly track the remaining join.

ArkLib/Data/CodingTheory/ReedSolomon/HiddenDerivative/Contracts.lean
```diff
@@ -0,0 +1,113 @@
...
+Concrete instances may expose further parameters—candidate degree, characteristic, weighted-degree
+cap, or hitting-extension degree—through `Equation` and `Valid`, without changing this contract. -/
+structure RootSolverContract {F Equation : Type*} [Semiring F] [DecidableEq F]
+    (designDim listBound : ℕ) (Valid : Equation → Prop)
+    (Satisfies : Equation → MessagePolynomial F designDim → Prop) where
+  /-- Enumerate candidate solutions of an equation. -/
+  solve : Equation → Finset (MessagePolynomial F designDim)
+  /-- On valid equations, solver membership is exactly equation satisfaction. -/
+  isExact : ∀ (equation : Equation), Valid equation → ∀ p,
+    p ∈ solve equation ↔ Satisfies equation p
```
alexanderlhicks, line 65: [API suggestion] `toCandidateCertificate` consumes only `.mpr` here: `Satisfies equation p → p ∈ solve equation`. It also consumes `card_le`, while the final decoder filters by actual agreement. Therefore a complete bounded enumerator that returns some extra candidates is enough for the integration theorem.

Consider a minimal contract with `complete` and `card_le`, plus an optional stronger exact-solver contract adding soundness. A concrete Kopparty solver can still prove the full iff. This is a reuse improvement, not a correctness objection to the current iff.

ArkLib/Data/CodingTheory/ReedSolomon/HiddenDerivative/Contracts.lean
```diff
@@ -0,0 +1,113 @@
...
+namespace ReedSolomon
+namespace HiddenDerivative
+
+open ListDecoding
+
+noncomputable section
+
+/-- The exact interface supplied by the interpolation proof.
+
+`Valid` should contain the structural facts needed by the chosen root solver, such as nonzeroness,
```
alexanderlhicks, line 36: [P0/P12/P14: make the concrete validity obligations explicit] For the Kopparty instance, the validity record should expose `Q ≠ 0`, derivative order `r ≤ D`, an integer `t > 0` with every `deg_{Y_j} Q ≤ t < q`, and weighted degree `< q²` with weights `(1,D,D-1,…,D-r)`.

Nonzeroness is essential: `Q=0` makes every degree-bounded polynomial a solution. The displayed statement of [Kopparty Theorem 4.3](https://theoryofcomputing.org/articles/v011a005/) omits it, although the algorithm/proof uses it; the BCPZZ restatement includes it.

The exact solution bound is

```text
t * (r+1) * q^(2*r*floor(D/q) + 4*r + 4).
```

For `r=d`, `D=K-1`, and `d<K≤q`, `floor(D/q)=0` and `d+1≤q`; `t<q` gives `t≤q`. Hence

```text
t * (d+1) * q^(4*d+4) ≤ q^(4*d+6).
```

The public list bound is correct once these premises are discharged. Please record this specialization as a named P12/P14 obligation rather than leaving the mathematical content solely in prose behind arbitrary `Valid`.

ArkLib/Data/CodingTheory/ReedSolomon/LowRateListDecoding/Main.lean
```diff
@@ -0,0 +1,106 @@
...
+    (hεSmall : ε < smallEpsilonBound θ)
+    (hnq : n ≤ q)
+    (hq : fieldSizeLowerBound n k ε θ ≤ q)
+    (domain : Fin n ↪ ZMod q) :
+    Nonempty
+        (DecoderCertificate domain k (agreementThreshold n ε) (listSizeBound q ε θ)) ∧
+      Code.IsListDecodable
+        (ReedSolomon.code domain k : Set (Fin n → ZMod q)) (1 - ε)
+        (listSizeBound q ε θ : NNReal) := by
+  sorry
```
alexanderlhicks, line 103: [Blocking: theorem status] The signature substantially matches the paper's extensional/list-size claim. The division edge cases are covered: `hk` and `hkn` imply `n>0`, and `ε,θ∈(0,1)` exclude the relevant zero denominators. `ceil(εn)` is also the correct closed-ball threshold: for `n=10`, `ε=1/4`, radius `3/4` permits at most 7 errors and therefore requires 3 agreements.

However, declaring the unfinished result with `sorry` conflicts with #854's explicit no-new-admissions rule. Please freeze the target as a `def LowRateStatement : Prop` (with these quantifiers/hypotheses) until the proof is available. That keeps the target inspectable without a theorem depending on `sorryAx`.

I also suggest separating certificate construction from a named corollary deriving `Code.IsListDecodable`. The latter should use the reusable agreement/distance bridge; the runtime omission is appropriate and already documented.

docs/kb/audits/bcpzz26-main-theorem.md
```diff
@@ -0,0 +1,137 @@
...
+  dimension, and the decoder rejects candidates of excessive degree or insufficient agreement.
+- The solver list bound is a parameter. The integration theorem does not hard-code Kopparty's
+  quadratic hitting extension or `4*d+6` exponent.
+- The degree bound is strict `< k`, not `≤ k`, and is carried by `Polynomial.degreeLT` rather than
+  `natDegree`; this handles the zero polynomial without a special convention.
+- The field-size hypothesis retains both independent parts. Although the embedding implies
+  `n ≤ q`, keeping `hnq` in the corollary mirrors the source arithmetic.
+- `Nonempty DecoderCertificate` is extensional existence, not executability or a complexity claim.
+
+## Confirmed source repairs
```
alexanderlhicks, line 76: [Source audit / P8 / P14] The ambient `K` repair and the proposed arithmetic repairs check out mathematically. Please add per-repair status labels (e.g. independently derived / Lean-checked / author-confirmed) and record the separate ceiling obligation for `B<q`. The page already says it is a statement-fidelity ledger; this is a request to make each repair's status explicit.

Here are derivations and the remaining obligations, so downstream contributors can use them. Throughout, `0<θ<1`; the lattice estimate assumes `d≥3`. These arithmetic derivations have been checked mathematically, but **have not been formalized in Lean in this review**.

<details>
<summary>Lattice bound, support-cap ceiling, and the separate field-bound ceiling</summary>

The paper hides a term in `O_θ(1)`. The PR proposes, for `m=d^3` and

```text
W = floor ((1+θ/2) * d*m / log(e*d)),
```

the bound

```text
(d-1) * (m + binom d 2) / W
  ≤ 2/(2+θ) * log d + 5/4.
```

This particular repair is mathematically sound, and in fact a slightly stronger constant follows.
Let

```text
L = log(e*d) = 1 + log d
a = (1+θ/2) * d^4 / L
N = (d-1) * (d^3 + d(d-1)/2).
```

For `d ≥ 3`, `L ≤ d`, hence `a ≥ d^3`. Since `W = floor a`,

```text
W > a - 1 ≥ a * (1 - 1/d^3).
```

Also

```text
((1 - 1/d) * (1 + (d-1)/(2*d^2))) / (1 - 1/d^3) ≤ 1,
```

because, after multiplying by `2*d^3`, this reduces to `d^2 + 2*d ≥ 3`.
Consequently,

```text
N / W < L / (1+θ/2)
      = 2/(2+θ) * log d + 2/(2+θ)
      < 2/(2+θ) * log d + 1.
```

Thus `5/4` is safe. P8 can prove the stronger `+1` form and weaken it.

The paper's small-`ε` condition also supplies much more than the remaining displayed margin. Put
`u = log(1/θ) ≥ 0`. From the stated bound on `ε` and `d ≥ ε^(-3/θ)`,

```text
θ(1-θ)/18 * log d
  > (5+θ)/6 * log (768 / (θ^3(1-θ))).
```

Using `log 768 > 6`, `log 7 < 2`, and `4+3θ < 7`, the right side is greater than
`5 + (5/2)u`, while

```text
log ((4+3θ)/θ) + 5/4 < 13/4 + u.
```

The desired margin follows. These elementary inequalities are good candidates for small named P14
lemmas rather than one large nonlinear tactic proof.

Although the mathematics checks out, explicit labels such as “independently derived; Lean proof
pending” would make the status easier to track. The audit already says it is a statement-fidelity
ledger, so its `Confirmed source repairs` heading alone does not claim a completed Lean proof.

### Support ceiling: The support-cap ceiling repair is correct but unfinished in Lean

The typeset proof of Lemma 3.2 uses the false inequality

```text
ceil ((1+3θ/4)m) ≤ (1+3θ/4)m.
```

The corrected route is

```text
|c| < (1+3θ/4)m + 1,
mA/(K-1) - |c|
  > θm/4 + θ^2*m/(1-θ) - 1.
```

The final term is positive under the paper's small-`ε` hypothesis. A coarse derivation is enough:
that hypothesis implies `ε < θ^3`; hence

```text
d ≥ ε^(-3/θ) > θ^(-9/θ),
m = d^3 > θ^(-27/θ) ≥ (1-θ)/θ^2.
```

Therefore `θ^2*m/(1-θ) > 1`. This validates the proposed repair, but it remains a P14 Lean
obligation.

### Field-bound ceiling: Record the second ceiling-sensitive bound

In the proof of Theorem 4.1, the paper also needs

```text
B = ceil (mA/(K-1)) < q.
```

The typeset page asserts the intermediate estimate

```text
ceil (mA/(k-1)) ≤ 2*d^3*ε*n/k
```

without showing how the ceiling is absorbed. There is ample slack under the displayed small-`ε`
condition, but it is a distinct proof obligation from Lemma 3.2's cap `C`. In particular, P14 should
use `ceil x < x+1`, derive a coarse lower bound on `εn` from `d < k ≤ K`, and explicitly absorb the
extra `1` before invoking the field-size hypothesis.

This second rounding step should be added to the source-audit ledger rather than disappearing inside
the eventual field-bound proof.


</details>

scripts/axiom_baseline.json
```diff
@@ -250,6 +248,11 @@
   "Reduction.liftContext_runWithLog",
   "Reduction.seqCompose_completeness",
   "Reduction.seqCompose_perfectCompleteness",
+  "ReedSolomon.HiddenDerivative.exists_decoderCertificate_of_contracts",
```
alexanderlhicks, line 251: [Blocking: avoid new admission debt] These five additions can be avoided in this statement/interface PR. The two inline filtering proofs remove four entries (the two lemmas and their two dependents), and representing the unfinished public target as a proposition removes the fifth. This follows the no-new-`sorry`/`axiom` rule in #854.

Please regenerate the baseline after those changes. The two unrelated `OutOfDomSmpl` removals should also be explained or separated from the feature diff, even if they are legitimate cleanup.

ArkLib/Data/CodingTheory/ReedSolomon/ListDecoding/Specification.lean
```diff
@@ -0,0 +1,154 @@
...
+This file defines reusable extensional contracts for Reed–Solomon list decoding. A final decoder
+returns exactly the degree-bounded polynomials meeting an absolute agreement threshold. A candidate
+generator may return false positives, but it must include every polynomial meeting that threshold;
+`CandidateCertificate.toDecoderCertificate` records the target obtained by filtering those
+candidates by message degree and actual agreement.
+
+The separation between `messageDim` and `designDim` is intentional. An interpolation or root-finding
+algorithm may work in the larger ambient space of polynomials of degree `< designDim`, then filter
+back to the target code of degree `< messageDim`. This is the interface needed to repair the
+`k`/`K` gap in [BCPZZ26], and it is useful independently of that paper.
```
alexanderlhicks, line 22: [Documentation] Please add the module-level `## References` section for `[BCPZZ26]`, following `CONTRIBUTING.md`. `Contracts.lean` should similarly give a formal Kopparty reference if it keeps that discussion.

Minor API naming suggestion: `minAgreement_le_agree_of_mem` and `mem_of_minAgreement_le_agree` would state the direction more explicitly than `agreement_le_of_mem` and `mem_of_agreement_le`.

alexanderlhicks (changes requested):
Reviewed commit `c7e9e01edf95b99adc30b76cb6bfcd2b59367c30` against [BCPZZ26 / ECCC TR26-164](https://eccc.weizmann.ac.il/report/2026/164/) and [Kopparty Theorem 4.3](https://theoryofcomputing.org/articles/v011a005/), including the existing ArkLib interfaces and the separate `kz99/rs-ld-mca` development.

The public statement is substantially faithful. I checked the strict degree convention (including zero), `ceil(εn)` agreement threshold, positive denominators, ambient `k≤K` filtering, and the intended `q^(4d+6)` specialization. The natural-parameter core and the separation of mathematical correctness from executability/runtime are appropriate.

I am requesting changes before treating this as the completed P0 freeze:

1. Close the two generic filtering gaps using the **compiled replacement proofs posted inline**, and represent the unfinished headline target as a proposition until proved. The current three admissions propagate to five baseline declarations and conflict with #854's merge discipline.
2. Correct the meaning/name of `yDegreeCap` (`B` bounds joint jet degree), and include `C` in the issue's parameter inventory.
3. Freeze the concrete equation/satisfaction/validity layer shared by P11/P12, coordinating with the P4b work already reported on the issue. The generic adapter alone cannot fix those representations.
4. Record Kopparty's precise preconditions and the exact exponent specialization, plus both ceiling-sensitive parameter obligations. An inline audit comment includes the mathematical derivations, including a stronger lattice constant.

The weaker bounded-enumerator interface and splitting the public theorem into certificate and `IsListDecodable` layers are reuse suggestions. Citation/naming observations are minor.

Validation: the three affected modules built at the original head; after applying the two filtering proofs in a disposable checkout, they built again under normal project limits. `#print axioms` then showed no `sorryAx` for either filtering lemma, `toDecoderCertificate`, or the conditional decoder theorem; the unfinished public theorem remained dependent on `sorryAx`. The arithmetic derivations are mathematical review evidence, not newly Lean-checked lemmas. Broader architecture/provenance feedback is being posted on #854.

docs/kb/audits/bcpzz26-main-theorem.md
```diff
@@ -0,0 +1,137 @@
...
+```text
+θ(1-θ)/18 * log d ≥ log ((4+3θ)/θ) + 5/4.
+```
+
+The parameter-discharge theorem must prove that condition (20) supplies this margin. The core
+contracts do not depend on `768`.
+
+### Ceiling in the support cap
+
+Lemma 3.2 uses the false inequality `ceil x ≤ x`. The repaired calculation uses
```
alexanderlhicks, line 103: Please describe the printed gap precisely here. In the [pinned PDF](https://eccc.weizmann.ac.il/report/2026/164/download#page=10), page 9 defines the support using |c| ≤ ceil(x), where x=(1+3θ/4)m; the Lemma 3.2 proof on page 10 instead writes |c| ≤ floor(x) ≤ x immediately before (19). The unjustified switch to the floor cap loses the rounding slack; the paper does not print ceil(x) ≤ x. Please update this sentence and the matching blueprint/paper-note wording while retaining the proposed repair accounting for that slack. This also corrects my corresponding wording in the [earlier review](https://github.com/Verified-zkEVM/ArkLib/pull/855#discussion_r3938520125).

alexanderlhicks (changes requested):
Re-reviewed `c7e9e01edf95` against the source, issue #854, and the actual CI log; the existing changes request still applies. The two filtering lemmas and headline theorem remain admitted. CI reports 317 sorry-tainted declarations and passes the baseline extended by this PR, so that green result does not discharge the five added tainted declarations. Issue #854 explicitly requires ordinary hypotheses/contracts for conditional work and prohibits new admissions.

Please complete the filtering proofs, represent the unfinished headline target without an admitted theorem, and remove the corresponding baseline additions. The existing [parameter feedback](https://github.com/Verified-zkEVM/ArkLib/pull/855#discussion_r3938520116) also remains applicable: the paper's B bounds the joint exponent sum, so the public parameter description needs to match that meaning. The generic decoder contracts themselves are coherent; this review does not claim a contradiction in them. Existing inline threads contain the admission repairs; the additional source wording correction is below.

[Full CI](https://github.com/Verified-zkEVM/ArkLib/actions/runs/33873726334) tested a tree identical to reviewed head `c7e9e01edf95`. I inspected the build, warning/source-policy, axiom-fixture/baseline and blueprint/docs steps and logs. The baseline result is against this PR’s edited baseline, as discussed above.

PR #879, opened 2026-09-08
feat(sumcheck): derive round relations through multivariate projection

quangvdao:
## Purpose

Derive the univariate round premise from the existing multivariate Sumcheck relation. After earlier challenges are fixed, the round polynomial sums the original polynomial over its remaining coordinates. Summing that polynomial gives the current target; evaluating it at the next challenge establishes the successor round relation.

#872 exposed this projected oracle and proved completeness under a separate sum premise. This PR discharges that premise from the mathematical relation and supplies the invariant transport used by #883 and #892.

## Content and limits

`sum_cube_cons` splits a finite Cartesian-power sum into its first coordinate and the remaining coordinates. The coordinate insertion/cast lemmas reconcile the dependent round indices. `projectedRoundPolynomial_eval` identifies projected evaluation with the residual multivariate sum; `projected_sum_of_relationRound` derives the current sum condition. `relationRound_projected_output` proves the next relation for the honest projected target. The final theorem applies these results to the actual virtual-input executor.

The algebra works over a commutative semiring. It makes no challenge-distribution, adversarial, or soundness claim. The existing names express their mathematical operations and premises accurately, so the repair retains them.

## Repair, review, and validation

Reconciled with repaired #874. The only repair to the unique production module, `ProjectionTransport.lean`, is the required public module format; its theorem statements/proofs and the acceptance client remain unchanged. The generated root import is retained. Parent review checked the finite-sum decomposition and both relation directions; Sol High prepared the module migration.

Full `LAKE_ARTIFACT_CACHE=false LAKE_NO_CACHE=true ./scripts/validate.sh --axioms` passed, including all builds/tests, warning/style and repository/docs checks, and axiom regression. Census: 11,985 declarations across 475 modules; 291 inherited sorry-tainted declarations; zero nonstandard-axiom-tainted declarations; no new taint. `git diff --check` passed.

PR #881, opened 2026-09-08
feat(sumcheck): prove one-round reduction soundness

quangvdao:
## Purpose

Prove one-round Sumcheck reduction soundness for the actual typed executor: a false input sum claim produces an accepted, true closed output claim with probability at most `deg / |F|`, when the sent degree-bounded polynomial is fixed before a fresh uniform challenge over a finite field.

The bound is on acceptance together with a true output relation. It is not a bound on local verifier acceptance alone. The verifier can accept a dishonest message's sum while the retained input oracle disagrees with its next target.

## Argument and hypotheses

The input polynomial `p` and sent polynomial `q` are both degree bounded. The executor's output oracle retains `p`, while its target is `q(r)`. Exact execution and closing equations reduce the successful true-output event to the sent sum check and `p(r) = q(r)`.

If the input sum is false and the sent sum passes, `p - q` is nonzero. Its degree is at most `deg`, so agreement challenges are contained in at most `deg` roots. Uniform sampling gives the stated probability bound. When the degree exceeds the field size the bound may be trivial; no extra nontriviality hypothesis is hidden.

`executeRandomCommitment_soundness` permits arbitrary randomized, possibly failing message selection, provided it occurs before the fresh challenge. Failure and explicit rejection contribute no mass to the successful event. Measure-valued forms restate the same results without a countability assumption on dependent run values.

This covers fixed or randomized pre-challenge messages, not challenge-dependent messages, arbitrary adaptive/persistent-world adversaries, multi-round error accumulation, or knowledge soundness. “Commitment” here means choosing the message first, not a cryptographic commitment scheme.

## Repair and review map

Reconciled with repaired #879. `Soundness.lean` uses current public module syntax and `Verifier.liftAccessImpl`; the two `*_measure_soundness` names become `*_measureSoundness`. `committedRun` remains a useful algebraic normal form whose provenance is supplied by `executeCommitted_eq`. The acceptance client and axiom canaries use the new names; theorem hypotheses, root argument, probability operations, and effects are unchanged. The generated import and naming documentation accompany the source.

Sol High prepared the migration; parent review inspected the executor equality, retained-oracle event, root bound, and quantifier order.

## Validation

Full `LAKE_ARTIFACT_CACHE=false LAKE_NO_CACHE=true ./scripts/validate.sh --axioms` passed, including all builds/tests, warning/style and repository/docs checks, and axiom regression. Census: 11,995 declarations across 476 modules; 291 inherited sorry-tainted declarations; zero nonstandard-axiom-tainted declarations; no new taint. `git diff --check` passed.


alexanderlhicks (dismissed):
Reviewed the committed one-round soundness bound against the actual executeCore semantics. Acceptance from a false input gives a nonzero difference polynomial; both degree bounds control its roots. The message is chosen before the fresh uniform challenge, with potentially failing message computations handled explicitly. CI compiles the soundness clients and axiom checks. No blocker survived the independent second review. This is the stated single-round committed-message bound; later multi-round soundness needs its own composition argument.

[Full CI](https://github.com/Verified-zkEVM/ArkLib/actions/runs/34301741412) passed on the source tree of reviewed head `229bbf99b05c`. I inspected the actual build, acceptance coverage, warning/source-policy gates, axiom-fixture/baseline results and blueprint/docs checks. A restacked revision needs fresh validation.

PR #894, opened 2026-09-08
refactor(ring-switching): use the explicit tensor basis API

alexanderlhicks:
Advance CompPoly from the `v4.33.1` release pin (`a09455a`) through current main to the reviewed [tensor-basis cleanup](https://github.com/Verified-zkEVM/CompPoly/pull/325) (`9b14b21`) and import its neutral `LinearAlgebra.TensorProduct.Basis` module from ring-switching packing. This incorporates the 16 already-merged upstream commits since the release pin as well as the tensor change. Remove a redundant import and correct the carrier/profile docstrings.

The packing definitions, the two embeddings, and the row/column reconstruction proofs are unchanged. Tests exercise the actual coordinate functions, preserve the default left algebra after importing the packing prelude, and distinguish the two coordinate views over `ℂ/ℝ`.

Validation: full `./scripts/validate.sh --axioms`, independent statement/polish review, and additional checks of both profile reconstruction laws. The coordinate functions and concrete profile use only standard axioms; existing protocol proof gaps remain. All dependency revisions other than CompPoly are retained.

Part of the B2 consumer migration in #893. Protocol orientation and stronger profile contracts in #615 remain separate.

Issue #893, opened 2026-09-08
roadmap: complete Binius interfaces, rejection semantics and proofs


alexanderlhicks:
Related trackers: [CompPoly #322 — binary fields/towers](https://github.com/Verified-zkEVM/CompPoly/issues/322) and [leanerVM #3 — Flock](https://github.com/Verified-zkEVM/leanerVM/issues/3).

## Current independent review — 2026-09-09

A new independent review of CompPoly #325–#331 and ArkLib #894–#895 found no introduced arithmetic or API correctness regression. It did find a pre-existing ArkLib packing defect preserved by the migrations: row and column coordinates were reversed relative to DP24 while the verifier used the paper's orientation. Over GF(4)/GF(2), actual honest packing of `t(X₀,X₁)=X₀` at zero rejected the correct claim. [The repair](https://github.com/Verified-zkEVM/ArkLib/pull/896) corrects both coordinate definitions and reconstruction laws, with regressions through actual packing, original-claim acceptance/rejection, batching and the final multiplier. Independent compilable wrong-selector mutations detect each boundary. Full source, native and axiom gates retain the same 291 existing `sorryAx`-tainted declarations, with no new taint; local API docs and independently rendered blueprint/declaration checks also pass. The repair is draft #896 at `6a524b70f5bd0b740a283ca7d848fac13c9fd29f`; the exact-head core run [34348993115](https://github.com/Verified-zkEVM/ArkLib/actions/runs/34348993115), including native, axiom and documentation checks, and import CI now pass on its stacked branch. This corrects paper fidelity beyond the earlier migration-only review.

[CompPoly #332](https://github.com/Verified-zkEVM/CompPoly/pull/332), originally reviewed at `65b9dd36b05665cc713c8df8e8d332fbb1327c8f`, separately corrects the extension guide's false divisibility/binomial-irreducibility claims and describes the actual cardinality/irreducibility certificates. Independent review and all five applicable CI workflows passed on that revision. These are documentation corrections; executable field operations do not change.

[...]

PR #895, opened 2026-09-08
fix(probability): support nominal sextic extension elements

alexanderlhicks:
The nominal extension carrier in [CompPoly #326](https://github.com/Verified-zkEVM/CompPoly/pull/326) no longer reduces to a vector, so the sextic sampler’s Vector-based `LawfulBEq` instance is invalid. Pin the reviewed nominal-carrier commit (`10cf263`) and use CompPoly’s generic lawful equality instance instead.

The explicit coefficient equivalence, six-limb sampler, and uniformity theorem are unchanged. Only the CompPoly package record changes in the dependency manifest.

Validation: full `./scripts/validate.sh --axioms`, independent source/polish/dependency review, and focused checks of coefficient round trips, lawful equality, and the sampling theorem’s standard-axiom closure.

Stacked on #894. This is the sextic-sampling compatibility portion of B3-P in #893; field encoding and Flock admission remain separate obligations.

The explicitly dispatched [core CI run](https://github.com/Verified-zkEVM/ArkLib/actions/runs/34287683414) passed at the exact head `20729d19fe4ae427955f8f2e46418b5ab2a215a4`, including clean/warm/native builds, axiom checks and the full validation wrapper. The separately dispatched [import check](https://github.com/Verified-zkEVM/ArkLib/actions/runs/34287684779) also passed. These dispatches cover the workflows whose automatic PR triggers exclude this stacked base branch.


quangvdao (dismissed):
Reviewed the repaired head on #894's refreshed base. No remaining blocking finding.

The extension carrier is now a nominal structure retaining its presentation, with an executable equivalence to six coefficients. Transporting six independent uniform base-field samples through that equivalence preserves the intended uniform distribution without enumerating all extension elements. The sampler, equivalence, and uniformity theorem are unchanged.

The removed local LawfulBEq instance was valid at the previous vector-alias pin. It becomes incompatible and redundant after the nominal change; CompPoly now owns the generic coefficient-based instance. This is not evidence that the old equality theorem was unsound. The final unique diff is only three files, +3/-7, with matching exact CompPoly pins and no unrelated dependency changes.

Full `./scripts/validate.sh --axioms` passed on the final source tree, including compiled execution and the public sampling/code-generation consumers. The sweep checked 12,421 declarations in 491 modules: 291 inherited sorry-tainted declarations, zero nonstandard taint, no new taint. #896 completes the coordinated stack's protocol-orientation repair.

quangvdao (approved):
Reconfirmed after #894 squash landing: this revision preserves the independently reviewed and fully validated tree `efbc7372cf44c8e5702bd7a91665a747aafff2fb` exactly. Only ancestry was updated and the PR retargeted to main. The review and validation conclusions above remain unchanged.

PR #913, opened 2026-09-21
refactor(probability)!: retire ArkLib PMF probability surface

quangvdao:
## Summary

Retire ArkLib's remaining PMF-based probability surface after the Lean 4.34 upgrade in #903. Coding-theory and protocol events now use VCVio's native measures, `ProbComp`, and `Pr{let x ← $ᵗ S}[...]` throughout.

Closes #904. The closure gates are listed under *Closure gates* below. VCVio #532 is a separate, broader upstream campaign and remains open.

## Motivation and changes

#903 converted deprecated VCVio-facing calls so Lean 4.34 could reach main, while leaving ArkLib's independent Mathlib-PMF toolkit for this follow-up. Merely replacing notation would leave old constants in elaborated proofs. This PR removes that surface and makes an empty semantic retirement scan mandatory.

| Area | Before | After |
|---|---|---|
| Uniform coding-theory events | PMF notation and finite/nonempty carriers | Native uniform samplers with explicit `SampleableType` assumptions at probability boundaries |
| Probability toolkit | Local notation and generic probability wrappers | Specialized Schwartz–Zippel, dot-product, and collision-image mathematics; generic laws imported from VCVio |
| Compatibility imports | `ArkLib/ToVCVio/` | Tree deleted; consumers import upstream owners directly |
| Serialization distance | Global `Dist (PMF α)` instance | Explicit finite-measure L1 distance, retaining twice-total-variation normalization |
| Enforcement | Warnings did not detect all PMF uses | Syntax rejection plus declaration type/body scanning, with no baseline exceptions |

The collision-image theorem remains universe-polymorphic and returns a positive-mass witness from a countable full-mass carrier. Its native computation corollary converts that witness to operational support; ToyProblem still extracts operationally reachable sampled parameters. Probability bounds, round independence, and protocol game terms are preserved. Existing admitted results remain admitted; this PR does not claim to complete their mathematics.

[VCVio #770](https://github.com/Verified-zkEVM/VCVio/pull/770), now on VCVio `main` through #771, supplies conditional uniform-product bounds, a `Fin.snoc` additive bound, and an explicit dependent-product sampler. These remove repeated reasoning from generator and folding proofs instead of adding ArkLib shims. The sampler constructor is not a global fallback instance.

## Completeness review

Measured with proof-body line counts from the final `:= by`; full tables are in `docs/design/native-measure-ledger.md`.

[...]

Issue #904, opened 2026-09-20
Convert ArkLib to VCVio v4.34 native measure semantics (campaign tracker)


dtumad:
## Goal

Convert ArkLib to VCVio's native Mathlib-measure probability semantics (`𝒟[…]` / `Pr{…}[…]`,
VCVio issue Verified-zkEVM/VCVio#532) on Lean v4.34.0, retiring both of ArkLib's probability
surfaces:

1. the VCVio-facing scalar API (`Pr[… | …]`, `probEvent`/`probOutput`/`probFailure`, `evalSPMF`/`𝒮[…]`,
   `NeverFail`, the PMF-based `IsUniformSpec`/`IsProbabilitySpec`): ~162 `Pr[`, 175 `probEvent`,
   38 `probFailure`, 22 `probOutput`, 14 `𝒮[` across `OracleReduction/`, `ProofSystem/`, `Commitments/`;
2. the pure-PMF notation `Pr_{ let x ←$ᵖ S }[…]` (`ArkLib/Data/Probability/Notation.lean`): 331 sites in
   41 files (coding theory, STIR, ToyProblem, BatchedFri) plus the 847-line `Instances.lean` and
   360-line `Combinatorial.lean` toolkit over `PMF`.

Constraints agreed for the campaign:

- **No shims.** Every modeling gap is fixed upstream in VCVio first. `ArkLib/ToVCVio/` shrinks and
  is deleted at the end; `Data/Probability/Notation.lean` is deleted.
- **Proofs must get shorter, not longer.** Any line-count regression is recorded per family in
  `docs/design/native-measure-ledger.md` with one of four admissible reasons (R1 subprobability
  honesty `Pr{¬p}` vs `1 - Pr[p]`; R2 explicit losslessness of an `OptionT` prefix for a lower bound;
  R3 support-vs-AE bridge on a non-`OracleComp` monad; R4 a proof-local `let : MeasurableSpace α := ⊤`).
  Anything else is a missing upstream lemma.
- Coding theory's carrier is `ProbComp` + `[SampleableType S]`: `Pr{let x ← $ᵗ S}[…]`, replacing
  `[Fintype S] [Nonempty S]` binders; subtypes/Finsets sample through `SampleableType.subtype`/`finsetCoe`.
- The legacy `OracleReduction/` security definitions are converted **in place** (same game terms:
[...]

PR #926, opened 2026-09-21
feat(lift-context): close the structural obligations in LiftContext/Reduction.lean

### LINKED ISSUE #676: Sequential-composition and context-lifting layer is stubbed, making composed security theorems vacuous
## Summary

`ArkLib/OracleReduction/Composition/Sequential/Append.lean`,
`ArkLib/OracleReduction/LiftContext/`, `ArkLib/OracleReduction/Cast.lean` and
`ArkLib/OracleReduction/Security/Implications.lean` are stubbed to the point where **every protocol
that composes rounds ends up with `sorryAx` in its top-level security theorems**, no matter how
complete the protocol's own proofs are.

This is currently the highest-leverage gap in the library: it is not one protocol's problem, it is
the shared substrate every protocol lands on.

### (a) Sorry-*defined* objects — the severe class

Here the theorem's *subject* does not exist, so a downstream statement is not a weaker claim but a
vacuous one, and no amount of docstring disclosure can make it honest.

| object | location | extent |
| --- | --- | --- |
| `OracleVerifier.append` | `Composition/Sequential/Append.lean:148,158` | `verify` field is `sorry` |
| `OracleVerifier.liftContext` | `LiftContext/OracleReduction.lean:53,57,58` | **all three** fields (`verify`, `embed`, `hEq`) |
| `OracleVerifier.cast` | `Cast.lean:128` | `verify` field is `sorry` |

### (b) Sorry-*proved* theorems

| theorem | location |
| --- | --- |
| `OracleVerifier.append_toVerifier` | `Append.lean:191` |
| `Verifier.append_rbrKnowledgeSoundness` | `Append.lean:506` (body `:515`) |
| `Reduction.append_completeness` | `Append.lean:439` |
| `OracleReduction.append_completeness` | `Append.lean:473` |
| `OracleReduction.append_perfectCompleteness` | `Append.lean:486` |
| `Verifier.rbrKnowledgeSoundness_implies_knowledgeSoundness` | `Security/Implications.lean:102` |

Supporting machinery also stubbed and on the same paths: `Extractor.RoundByRound.append`
(`Append.lean:255–260`), `Verifier.StateFunction.append.toFun_next/toFun_full` (`:299–300`),
`ProtocolSpec.Transcript.fst/snd` (`SeqCompose.lean:122,125,133`),
`seqComposeChallengeEquiv.left_inv` (`:500`), `LiftContext/Reduction.lean` (10 sorries).

## Measured impact

**PR #383 (FRI-Binius).** A census over its 37 modules (`Lean.collectAxioms` on every non-internal
constant) gives **2,013 axiom-clean / 89 `sorryAx`-tainted**, with **no `sorryAx` originating in the
PR**. All 89 trace to the stubs above, and all twelve of its advertised top-level theorems are
tainted. Its per-step mathematics — the DP24 soundness core, all four step reductions, the query
phase — is clean.

**PR #503 (LogUp).** Independently the same shape: `LogUp/**` sorry-free, but
`#print axioms logup_{soundness,completeness}` returns `sorryAx` inherited from `append_*`, the
`Implications.lean` implications, and `liftContext_*`.

Two unrelated protocol formalizations blocked at the identical point is the argument for treating
this as substrate work rather than as something each protocol PR works around.

## A possible reason `append_rbrKnowledgeSoundness` has stayed a stub

Offered as a hypothesis, not a finding.

`Verifier.rbrKnowledgeSoundness` is the **averaged** notion: the transcript prefix is prover-sampled
and the probability ranges over prefix and challenge together. PR #668 adds
`rbrKnowledgeSoundnessWorstCase` — worst-case-per-prefix, which its docstring calls "the standard
literature shape" — and proves worst-case ⟹ averaged, noting the converse fails.

Composition is exactly where that gap would bite: appending makes the second protocol's prefix
distribution depend on the first protocol's execution. A worst-case-per-prefix bound survives that
reindexing uniformly; an averaged bound is tied to the distribution it was proved against.

If that is right, #668 may already carry the missing ingredients — both the worst-case notion and
the mixture master bounds in `Security/RbrGame.lean`
(`probEvent_simulateQ_addLift_getChallenge_bind_le` and its `Option`-valued and prefix-extended
variants). Worth trying `append_rbrKnowledgeSoundness` against the worst-case notion before
attacking it against the averaged one.

Note #668 also removes all three `Security/RoundByRound.lean` sorries, but neither it nor #505
touches `Append.lean`, `Implications.lean`, `LiftContext/` or `Cast.lean`.

## The question to settle first

**Does #433 replace this layer wholesale?** PR #383's author explicitly paused waiting for #433 and
then resumed without it. If #433 supersedes `Append.lean`, filling these stubs is wasted effort and
the right move is to prioritise #433 and have protocol PRs land with disclosed taint in the interim.
If #433 does not cover the composition operators, this layer needs owning independently.

That decision is worth making explicitly, because right now protocol PRs are accumulating against a
substrate nobody has claimed.

## Suggested ordering if we do own it

1. `OracleVerifier.append`'s `verify` — everything else in (a) and (b) for the append path is
   downstream of the object existing.
2. `Verifier.append_rbrKnowledgeSoundness`, attempted against #668's worst-case notion first.
3. `OracleReduction.append_{completeness,perfectCompleteness}`.
4. `Verifier.rbrKnowledgeSoundness_implies_knowledgeSoundness` — independent of the append path, and
   on its own it unblocks the scalar-KS statements that protocol PRs most want to cite.
5. `OracleVerifier.liftContext` and `OracleVerifier.cast`.

## Interim convention worth adopting regardless

Make a `#print axioms` disclosure standing practice for advertised top-level security theorems, so
"the files are sorry-free" is never mistaken for "the results are proved". Both #383 and #503 would
have been accurately described from the outset under such a convention.

Abraxas1010:
Closes the four **structural** obligations in `OracleReduction/LiftContext/Reduction.lean`, plus the
completeness transport that sits on top of them.

Measured against `main`: that file carries **10 `sorry` occurrences across 8 declarations**. This PR
closes **4 of the 8** (10 → 5 occurrences). The 4 that remain are exactly the `Verifier.*` soundness
theorems, which are deliberately out of scope — see *Scope* below.

This is **not** "closes #676". Please read the scope section before merging.

The design question behind the largest change is raised separately in the #676 thread; if you'd
rather the interface took a different shape, that discussion should probably settle before this is
reviewed in detail.

---

## The finding this PR is really about

`Verifier.StateFunction.liftContext` was **not provable as typed**. The `sorry` was a symptom, not
unfinished work.

`StateFunction.toFun_empty` is a biconditional, and both directions are genuinely consumed:
`Composition/Sequential/Append/StateFunction.lean` uses `.mp` once (line 318) and `.mpr` four times
(324, 430, 521, 648), and `Security/RoundByRound.lean` uses `.mpr` once. But
`Statement.Lens.IsSound` supplies only `proj_sound : ∉ outerLangIn → ∉ innerLangIn`. The converse
doesn't exist anywhere in the API at the **language** level — `Context.Lens.IsComplete.proj_complete`
has the right shape but lives at the **relation** level (`Set (Stmt × Wit)`) and cannot discharge a
`Set Stmt` obligation.

The two conditions turn out not to be two hypotheses but **one equation**:

```lean
theorem Statement.Lens.eq_preimage_of_projSound_of_isComplete :
    outerLangIn = lens.proj ⁻¹' innerLangIn
```

`proj_sound` is one containment; the missing datum is the other; `toFun_empty` is that equation read
pointwise (`mem_iff_proj_mem`). Hence a new language-level class, mirroring the existing
relation-level one, passed as an instance argument.

**The alternative is deliberately not taken.** Weakening `toFun_empty` from `↔` to `→` removes the
`sorry` immediately, but breaks the already-landed append state function and weakens the
state-function notion for every protocol in the library — converting a faithful encoding of
Chiesa–Yogev Definition 31.1.1 into something weaker while leaving every downstream theorem name
unchanged.

---

## What changed, split by what needs your decision

**Interface changes (need a maintainer call) — `LiftContext/Lens.lean`**

| Item | What it is |
|---|---|
| `class Statement.Lens.IsComplete` | the missing language-level containment |
| `[lensComplete : …]` on `Verifier.StateFunction.liftContext` | new instance argument |
| 5 instances | incl. `instIsCompletePreimage`, the terminal one: every complete pair factors through the preimage language |
| `isSound_not_implies_isComplete`, `isComplete_not_implies_isSound`, `isSound_not_implies_isComplete_nonvacuously` | both separation directions, plus a witness where soundness is discharged against a genuine out-of-language statement rather than an empty quantifier |
| `preimage_subset_of_projSound` | isolates the single classical step (a contraposition out of a negated statement) so the layer's classical content stays auditable |

**Pure proof contributions (no interface change; statements unchanged)**

| Obligation | Note |
|---|---|
| `Extractor.RoundByRound.liftContext` | applies `E.extractMid`/`E.extractOut` and routes the statement through the lens; ships `liftContext_extractMid`/`_extractOut` pinning that it does |
| `StateFunction.toFun_full` | consumes `lensSound.lift_sound` |
| `Reduction.liftContext_runWithLog` | consumes `Prover.liftContext_runWithLog` |
| `Reduction.liftContext_completeness` | statement unchanged, no added hypothesis, no import change |

**One new general lemma, `Reduction.simulateQ_optionT_map`.** A pure post-map passes straight through
any simulation: `simulateQ impl (f <$> oa) = Option.map f <$> simulateQ impl oa`, for an arbitrary
target monad — naturality of the monad morphism `simulateQ impl` in the value. It's needed twice
(`liftContext_runWithLog` is the `WriterT` instance, `liftContext_completeness` the `StateT` one), so
the with-log lemma is now a one-line corollary. Net new API across both proofs is this single
declaration, and it is choice-free (`[propext, Quot.sound]`).

Two obstructions it resolves, neither visible from the statement:

1. **The lift starts trapped inside the simulation.** `liftContext_run` exhibits the lifted run as a
   pure post-map, but `completeness` evaluates its predicate on `(simulateQ pImpl _).run' s`, so the
   two probability events don't share a base computation until that post-map is commuted out.
2. **`compatContext` is stated about the wrong computation.** It's defined against
   `support (R.run …)` — the *unsimulated* reduction — while `probEvent_mono` supplies membership in
   the support of the simulated, state-threaded, `OptionT`-wrapped one. The bridge is that
   simulating can only shrink the support (`support_simulateQ_run'_subset`), since every `QueryImpl`
   answer already lies in the query's own range.

---

## Disclosed limitation of the instance set

All five instances are structural families (identity, input-only, output-only, the preimage/terminal
one) plus the separation witnesses. **None is a lens taken from `ArkLib/ProofSystem/`.**

I'd rather state why than leave it implicit: `Statement.Lens.IsSound` appears in no ArkLib source
file other than its own definition in `Lens.lean` — it has **zero instances**, is taken as an
instance argument at several sites, and is produced by no protocol. There is no protocol-level
precedent at the language level to mirror. At the *relation* level the picture differs
(`Sumcheck/Spec/SingleRound.lean` has a proved `proj_complete`; `Binius/FRIBinius` has one stubbed),
which is itself the point: the relation level is populated and the language level never was.

So the honest status is that the class is **inhabitable across structural families and demonstrably
non-redundant**, but has **not** been exercised on a real protocol. If you want that first, it's a
larger piece of work than this PR.

---

## Scope

- **In:** the non-probabilistic half of `LiftContext/Reduction.lean`, plus `liftContext_completeness`.
- **Out:** `liftContext_soundness`, `liftContext_knowledgeSoundness`, `liftContext_rbr_soundness`,
  `liftContext_rbr_knowledgeSoundness`, and all of `Security/Implications.lean`.
- **Why:** the excluded theorems are probability-statement-bearing, and #913 is actively retiring the
  PMF surface they're written over. Proving them now would invite rework.

The honest claim is: **this closes the structural half of #676; the soundness half remains open and
is gated behind the probability migration.**

**One excluded theorem does change, and it is a hypothesis change.** Three of the four
(`liftContext_soundness`, `liftContext_knowledgeSoundness`, `liftContext_rbr_knowledgeSoundness`) are
untouched. `liftContext_rbr_soundness` gains exactly one instance binder,
`[lensComplete : lens.IsComplete outerLangIn innerLangIn]`, because constructing the lifted state
function now requires it; its proof body and its `sorry` are unchanged. That's a strengthened
hypothesis on an out-of-scope theorem, and it's the one place this PR makes a downstream statement
weaker. If you'd prefer that theorem keep its current signature, the alternative is a separate
constructor taking the completeness datum explicitly — happy to do it that way.

**Interaction with #913.** I checked rather than assumed: #913 modifies this file only by deleting
six lines of dead scaffolding from `liftContext_rbr_knowledgeSoundness` (an out-of-scope theorem), and
it does **not** alter `structure StateFunction` or `toFun_empty` — so the argument above survives it.
The completeness proof here consumes VCVio's probability API directly, which is the target of that
migration rather than the surface being retired. Expect one trivial textual conflict in that
out-of-scope theorem depending on merge order; happy to rebase either way.

---

## Verification

| Check | Result |
|---|---|
| Declaration set preserved | **0 removed**, 42 added |
| `sorry` / `admit` / `stop` / `native_decide` in the four bodies | 0 |
| `Security/RoundByRound.lean` diff | **empty** — `structure StateFunction` untouched |
| `Security/Implications.lean` diff | **empty** |
| `#print axioms`, **every one of the 42 added declarations** (41 addressable; 1 is `private`) | **zero `sorryAx`** |
| of those 41: axiom-free / choice-free / using `Classical.choice` | **17 / 35 / 6** |
| `Extractor.RoundByRound.liftContext`, `simulateQ_optionT_map`, `run_simulateQ_writerT_optionT_map` | `[propext, Quot.sound]` — choice-free |
| `preimage_subset_of_projSound` | `[propext, Classical.choice, Quot.sound]` — the one deliberately classical step, isolated so it stays auditable |
| Substitution-body mutation fixtures (kept local, not submitted) | 4 harnesses, **21/21** mutants rejected, **7/7** controls still elaborating |
| Independent check of the pullback characterisation | brute-force enumeration over all 4 projections × 4 outer × 4 inner languages on `Bool`: **0 disagreements / 64** |

**What I could not check.** My development tree is pinned to **v4.31**, so the content above is
verified green there, not at v4.34. The module syntax (`module`, `public import`,
`@[expose] public section`) is taken from the current `main` file layout, and the `simp` → `simp only`
tightenings already on `main` are preserved rather than reverted. **CI here is the acceptance gate,
not my local build** — if it finds a v4.34-only breakage I'll fix it in this PR.

One note, since 4.34 makes unfolding export-controlled: every lemma these proofs rewrite with is a
*theorem*, so `simp only` consumes its statement and never its body, and the module system can't
change whether they fire. The one step discharging a defeq by computation rather than a named lemma
is the trailing `rfl` in `OptionT.probEvent_eq_of_run_map_eq _ _ _ _ rfl` — the first place to look
if 4.34 disagrees.


Abraxas1010:
Moving this to draft — CI is red at v4.34 and the reason is worth stating plainly, because I think it
is a finding about the issue rather than just a problem with my patch.

**What fails.** Four errors, in three of the four obligations:

| Declaration | Failure |
|---|---|
| `Verifier.StateFunction.liftContext` (`toFun_full`) | `introN failed: no additional binders` — the goal after unfolding has a different shape, and the leftover metavariable then trips the kernel check on the definition |
| `Reduction.liftContext_runWithLog` | `apply funext` can no longer unify; the two sides no longer present as a function equality at that point |
| `Reduction.liftContext_completeness` | `OptionT.probEvent_eq_of_run_map_eq … rfl` — the defeq that `rfl` was discharging no longer holds |

`Extractor.RoundByRound.liftContext` is unaffected. That is the one obligation whose proof is pure
data construction with no probability layer in it.

**Why I think this is structural rather than incidental.** All three failures are in tactic proofs
whose goals are probability expressions — `𝒟[…]`, `probEvent`, `simulateQ … .run`. Their *statements*
are not probability statements, which is why they read as the "structural" half of this issue. But
their *proofs* are, and the probability normal forms are exactly what #904/#913 are moving.

So the structural/probabilistic split that makes this issue look separable is a split of
**statements**, not of **proofs**. Three of the four supposedly-structural obligations sit on the
surface being migrated. I do not think that makes them unfixable — but it does mean fixing them
against a moving normal form is the same work the migration is doing, and doing it in parallel
invites the churn I was trying to avoid by scoping the soundness half out.

**What I am not going to do.** My development tree is pinned to v4.31, so I cannot reproduce these
goal states locally and would be iterating blind against CI on a public PR. I would rather not push a
sequence of guesses at your CI.

**Two questions, whichever is more useful to you:**

1. Is #913 close enough to landing that the sensible order is to wait for it and rebase these proofs
   onto the settled probability API? If so I am happy to park this and return after it merges.
2. Independently of the proofs — does the interface direction in #676 look right to you? The
   `Statement.Lens.IsComplete` finding stands on its own and does not depend on any of the three
   proofs above. `Extractor.RoundByRound.liftContext` and the `Lens.lean` additions build clean at
   v4.34 today; I could split those into a much smaller PR and leave the probability-layer proofs for
   after the migration.

Option 2 is probably the better shape if the direction is acceptable, and I am glad to cut it that
way. Sorry for the red CI in the meantime.

